PR #172, opened 2021-12-15
feat(components/Icon): add-specific-case-for-viewbox-size

mjovicic:
# Description
The new svg Icon component forces the viewBox to be "0 0 32 32". For most icons this is fine. However some (like "info-badge-filled" and "error-legacy-badge-filled"), the viewPort in the svg is "0 0 14 14" . When this gets overwritten, the svg is now being drawn with the wrong viewPort, so it appears smaller than it should (14/32 of the size it should be for the two examples). This PR fixes that case.

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-281641

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -2,7 +2,14 @@ import React from 'react';
 import './Icon.style.scss';
 import { Props } from './Icon.types';
 import { useDynamicSVGImport } from '../../hooks/useDynamicSVGImport';
-import { COLOR_INHERIT, DEFAULTS, GLYPH_NOT_FOUND, STYLE } from './Icon.constants';
+import {
+  COLOR_INHERIT,
+  DEFAULTS,
+  EXCEPTION_ICONS_LIST,
+  GLYPH_NOT_FOUND,
+  STYLE,
+  VIEW_BOX_SPECS,
+} from './Icon.constants';
 import classnames from 'classnames';
 
 /**
@@ -62,16 +69,6 @@ const Icon: React.FC<Props> = (props: Props) => {
 
   const { inheritedColors, styleColors } = getColors();
 
-  const viewBoxDefault = '0, 0, 32, 32';
-  const viewBoxSmall = '0, 0, 14, 14';
-  //This exception list is added for Icons with scale 14, so that view box can setup properly
-  const exceptionIconsList = [
-    'check-circle-badge',
-    'error-legacy-badge',
-    'info-badge',
-    'priority-badge',
-  ];
-
   if (SvgIcon) {
     return (
       <div className={classnames(STYLE.wrapper, className)} id={id} style={style}>
@@ -81,7 +78,9 @@ const Icon: React.FC<Props> = (props: Props) => {
           className={classnames({ [STYLE.coloured]: isColoredIcon })}
           style={{ ...styleColors }}
           {...inheritedColors}
-          viewBox={exceptionIconsList.includes(name) ? viewBoxSmall : viewBoxDefault}
+          viewBox={
+            EXCEPTION_ICONS_LIST.includes(name) ? VIEW_BOX_SPECS.SMALL : VIEW_BOX_SPECS.NORMAL
+          }
           width="100%"
           height="100%"
           data-scale={!autoScale && (scale || DEFAULTS.SCALE)}
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -62,6 +62,16 @@ const Icon: React.FC<Props> = (props: Props) => {
 
   const { inheritedColors, styleColors } = getColors();
 
+  const viewBoxDefault = '0, 0, 32, 32';
+  const viewBoxSmall = '0, 0, 14, 14';
+  //This exception list is added for Icons with scale 14, so that view box can setup properly
+  const exceptionIconsList = [
+    'check-circle-badge',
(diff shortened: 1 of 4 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -62,6 +62,16 @@ const Icon: React.FC<Props> = (props: Props) => {
 
   const { inheritedColors, styleColors } = getColors();
 
+  const viewBoxDefault = '0, 0, 32, 32';
+  const viewBoxSmall = '0, 0, 14, 14';
+  //This exception list is added for Icons with scale 14, so that view box can setup properly
+  const exceptionIconsList = [
```
torgeadelin, line 68: test

src/components/Icon/Icon.tsx
```diff
@@ -62,6 +62,16 @@ const Icon: React.FC<Props> = (props: Props) => {
 
   const { inheritedColors, styleColors } = getColors();
 
+  const viewBoxDefault = '0, 0, 32, 32';
+  const viewBoxSmall = '0, 0, 14, 14';
```
InteractiveTimmy, line 66: These values should be stored in the `Icon.constants.ts` file as something similar to:

```js
const VIEW_BOX_SPECS = {
  NORMAL: '0, 0, 32, 32',
  SMALL: '0, 0, 14, 14',
};

const DEFAULTS = {
  /* ...other defaults... */
  VIEW_BOX_SPEC: VIEW_BOX_SPECS.NORMAL,
};
```

src/components/Icon/Icon.tsx
```diff
@@ -62,6 +62,16 @@ const Icon: React.FC<Props> = (props: Props) => {
...
+  //This exception list is added for Icons with scale 14, so that view box can setup properly
+  const exceptionIconsList = [
+    'check-circle-badge',
+    'error-legacy-badge',
+    'info-badge',
+    'priority-badge',
+  ];
```
InteractiveTimmy, line 73: This should also exist in the `Icon.constants.ts` file, much like the above described `VIEW_BOX_SPECS` constants.

src/components/Icon/Icon.tsx
```diff
@@ -71,7 +81,7 @@ const Icon: React.FC<Props> = (props: Props) => {
           className={classnames({ [STYLE.coloured]: isColoredIcon })}
           style={{ ...styleColors }}
           {...inheritedColors}
-          viewBox="0, 0, 32, 32"
+          viewBox={exceptionIconsList.includes(name) ? viewBoxSmall : viewBoxDefault}
```
InteractiveTimmy, line 84: Please add an attribute test and a snapshot test for this change within the local test file for this component.

InteractiveTimmy (changes requested):
Great bug find/fix. A few code-style items as well as a request for adding some testing.

src/components/Icon/Icon.unit.test.tsx
```diff
@@ -266,4 +266,23 @@ describe('<Icon />', () => {
       await mountAndWait(<Component />);
     });
   });
+
+  describe('exception case for small icons', () => {
+    const exceptionIconsList = [
```
torgeadelin, line 271: Once you fix @InteractiveTimmy 's suggestions you should probably import the exceptionList from the `.constants.ts` file here.

src/components/Icon/Icon.unit.test.tsx
```diff
@@ -266,4 +266,23 @@ describe('<Icon />', () => {
       await mountAndWait(<Component />);
     });
   });
+
+  describe('exception case for small icons', () => {
```
torgeadelin, line 270: Also, add one test for a non-exception icon, making sure it has the correct viewbox.

torgeadelin (changes requested):
Thanks for adding the test.
Small changes are still needed.

torgeadelin:
> Great bug find/fix. A few code-style items as well as a request for adding some testing.
> 
> This is potentially something that we should have corrected on the design level as well, as all icons should be provided in a standard size.

These icons are only designed for viewport 14 as far as I understood. And they are the only exceptions to the rule. It is due to accessibility issues.

All rest icons are designed at 32 viewport.

src/components/Icon/Icon.unit.test.tsx
mjovicic: I have made the adjustments.

src/components/Icon/Icon.unit.test.tsx
mjovicic: I have made the adjustments.

src/components/Icon/Icon.tsx
mjovicic: I have made the adjustments.

src/components/Icon/Icon.tsx
mjovicic: I have made the adjustments.

src/components/Icon/Icon.unit.test.tsx
torgeadelin: One more thing, please add these tests under `describe('attributes' () => { ...` instead of having them separate.

mjovicic: It makes sense, I fixed it.

torgeadelin (approved):
looks good :)

InteractiveTimmy (approved):
LGTM! Approved.

PR #173, opened 2021-12-15
feat(sace-list-item): fix copact color

torgeadelin:
# Description

Small css fixes for compact mode / size 32 of the SpaceListItem component.
It also includes font colour correction for default vs activity states.

I have also removed the colour on the Text component. This should inherit the default colour from the theme provider. It makes it hard to override if it has a specific class with colour assigned to it.

### DIFF
--- a/src/components/ListItemBase/ListItemBase.style.scss
+++ b/src/components/ListItemBase/ListItemBase.style.scss
@@ -5,7 +5,7 @@
   align-items: center;
   font-size: 1rem;
   line-height: 1.5rem;
-  color: var(--theme-text-primary-normal);
+  color: var(--listitem-textsecondary);
   transition: background-color 0.2s, color 0.2s, border-color 0.2s;
   box-sizing: border-box;
   outline: none;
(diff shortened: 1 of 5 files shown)


robstax (approved):

src/components/ListItemBase/ListItemBase.style.scss
```diff
@@ -60,8 +65,13 @@
   }
 
   &[data-size='32'] {
+    & > *,
+    .md-text-wrapper {
+      font-weight: 500;
+      font-size: 0.75rem;
+    }
```
InteractiveTimmy, line 72: The usage of the comma here is a bit strange. So, all immediate children as well as any child with the class `.md-text-wrapper`? Is there a better way to write this?

InteractiveTimmy (approved):
One non-blocking comment.

src/components/ListItemBase/ListItemBase.style.scss
torgeadelin: Yeah, you got it! I don't know.
I need the immediate children for cases where we have something like.
```js
<ListItemBase>
  <ListItemBaseSection position="fill">Text here</ListItemBaseSection>
</ListItemBase>

//or

<ListItemBase>
  Text here
</ListItemBase>
```

InteractiveTimmy (approved):
Thank you for adding that comment. Approved.

PR #177, opened 2021-12-20
refactor(automation): improve deployment pipeline

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to improve our deployment of both documentation and package references. This also enables deployment to a non-protected branch for documentation.

# Links

[SPARK-293872](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-293872)

### DIFF
--- a/.github/workflows/deploy-documentation.yml
+++ b/.github/workflows/deploy-documentation.yml
@@ -1,4 +1,4 @@
-name: Deploy CD
+name: Deploy Documentation CD
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-documentation.yml
```diff
@@ -55,82 +55,36 @@ jobs:
...
-          then
-            echo "Found $FILE_COUNT changed documentation files."
-            git commit -m "ci(docs): deploy documentation [skip ci] [skip release]"
-            git push
-          else
-            echo "Documentation files have not changed."
-            echo "Documentation will not be published."
-          fi
\ No newline at end of file
+          git add ./
```
torgeadelin, line 88: Is it always going to be in sync with `origin documentation`?
If yes, then it's fine, otherwise maybe we need to sync the current branch with any `documentation` branch changes. I can't think of a way `documentation` will be ahead of master (at least in terms of contents of the `/docs`).

torgeadelin (approved):
Looks good, see my only comment but I'm almost certain that wouldn't be a problem.

.github/workflows/deploy-documentation.yml
InteractiveTimmy: Well, documentation will contain only the docs folder. Basically, the script maps the cached `./docs` from `master` to `./` on the `documentation` branch. It should never be up to date with master and should diverge immediately.

The automation flow: `checkout master` => `install dependencies` => `build dist && docs` => `cache docs` => `clear documentation branch` => `uncache docs to ./ on documentation branch` => `push to documentation branch`

PR #184, opened 2021-12-30
test: added react-testing-library and added first demo test

cscheffauer:
# Description

This MR adds React Testing Library to Momentum and a first demo test. This test should showcase how testing with React Testing Library can look like and how easy it can be to write proper tests in comparison to Enzyme. 

Feel free to discuss underneath.

# Links

[React Testing Library](https://testing-library.com/docs/react-testing-library/intro)


jest.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   snapshotSerializers: ['enzyme-to-json/serializer'],
   snapshotResolver: '<rootDir>/config/jest/jest.snapshot-resolver.js',
   setupFiles: ['jest-canvas-mock'],
-  setupFilesAfterEnv: ['<rootDir>/src/helpers/enzyme_setup'],
+  setupFilesAfterEnv: ['<rootDir>/src/helpers/enzyme_setup', '<rootDir>/jest-setup.ts'],
```
InteractiveTimmy, line 13: The setup file should be stored within the same scope of the other setup files. e.g. `./src/helpers/jest-setup.ts`, or potentially stored within the `./config` folder.

tsconfig.json
```diff
@@ -25,6 +25,7 @@
     ]
   },
   "include": [
-    "./src/**/*.*"
+    "./src/**/*.*",
+    "./jest-setup.ts"
```
InteractiveTimmy, line 29: Moving this file into the `./src/helpers/**` folder will prevent this change from being necessary.

src/components/ButtonCircle/ButtonCircle.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  jest.clearAllMocks();
+});
+
+//TODO: get rid of ButtonCircle.unit.test.tsx and move over snapshot tests, after it has been discussed with the team
+
+it('should have provided title when title is provided', () => {
+  const buttonTitleMock = 'Example Title';
+  render(<ButtonCircle title={buttonTitleMock} />);
+
+  expect(screen.getByTitle(buttonTitleMock)).toBeVisible();
```
InteractiveTimmy, line 18: This non-mutable approach to testing components allows for a very clear test scope. Since we are not storing the render output of `render(<Component />)` in a mutable variable using `react-testing-library`, we can ultimately only test the final output, and not the internal mutations performed on the **ReactNode**.

This will further prevent unnecessary tests from being written, as we will only test the finalized DOM component.

src/components/ButtonCircle/ButtonCircle.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+});
+
+it('should disable button as expected', () => {
+  const buttonLabelMock = 'Test Button';
+  render(<ButtonCircle onPress={mockCallback} aria-label={buttonLabelMock} disabled />);
+
+  userEvent.click(screen.getByRole('button', { name: buttonLabelMock }));
+  expect(mockCallback).not.toHaveBeenCalled();
```
InteractiveTimmy, line 26: This approach to handling the rendered component without storing a ReactNode removes the need for utils wrapped for `ReactAria`, which further cleans up potentially unnecessary code in this project.

This also prevents tests from directly calling `component.onPress()` or similar event methods to simulate events incorrectly.

src/components/ButtonCircle/ButtonCircle.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+});
+
+it('should fire onClick callback as expected after clicking', () => {
+  const buttonLabelMock = 'Test Button';
+  render(<ButtonCircle onPress={mockCallback} aria-label={buttonLabelMock} />);
+
+  userEvent.click(screen.getByRole('button', { name: buttonLabelMock }));
```
InteractiveTimmy, line 33: It appears that we also have `hover` events as well that should be tested to confirm that they work along side `ReactAria`. Events available [here](https://testing-library.com/docs/ecosystem-user-event/#pointer-events-options).

src/components/ButtonCircle/ButtonCircle.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  expect(mockCallback).toHaveBeenCalledTimes(1);
+});
+
+it('should fire onClick callback as expected after keyboard interaction', () => {
+  const buttonLabelMock = 'Test Button';
+  render(<ButtonCircle onPress={mockCallback} aria-label={buttonLabelMock} />);
+
+  userEvent.tab();
+  userEvent.keyboard('{enter}');
```
InteractiveTimmy, line 42: This is a much more direct way of testing keyboard events, which was a testing gap we had. 👍👍👍

InteractiveTimmy (changes requested):
I really like these changes, mostly non-blocking discussion, but I would like to see the config moved into its appropriate folder, as well as more testing within `<ButtonCircle />` once we all agree on these conventions moving forwards.

src/components/ButtonCircle/ButtonCircle.test.tsx
cscheffauer: what exact hover events on the ButtonCircle are relevant to test? the onHoverStart, onHoverEnd callbacks for example?

tsconfig.json
cscheffauer: will do that whenever we decide using RTL as a way forward! :)

src/components/ButtonCircle/ButtonCircle.test.tsx
InteractiveTimmy: This is correct, `onHoverStart` and `onHoverEnd`, you can see examples of these events in the testing utils file: `./test/utils.ts`. These are used in `./src/components/ButtonSimple/ButtonSimple.unit.test.tsx`, which `<ButtonCircle />` extends.

cscheffauer: thanks - will try adding tests for them :)

PR #185, opened 2022-01-06
feat(avatar): export PresenceType enum

torgeadelin:
# Description

Export `PresenceType` enum as part of the default exports of the library.

### DIFF
--- a/src/components/Avatar/index.ts
+++ b/src/components/Avatar/index.ts
@@ -1,6 +1,9 @@
 import { default as Avatar } from './Avatar';
-import { Props } from './Avatar.types';
+import * as CONSTANTS from './Avatar.constants';
 
+import { Props, PresenceType } from './Avatar.types';
+
+export { CONSTANTS as AVATAR_CONSTANTS, PresenceType };
 export type AvatarProps = Props;
 
 export default Avatar;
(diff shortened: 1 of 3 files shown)


src/components/Avatar/index.ts
```diff
@@ -1,6 +1,9 @@
 import { default as Avatar } from './Avatar';
-import { Props } from './Avatar.types';
+import * as CONSTANTS from './Avatar.constants';
 
+import { Props, PresenceType } from './Avatar.types';
```
cscheffauer, line 4: just something I love doing - you can do `import type {...} from './Avatar.types'` to highlight that these are types :) feel free to do it as well if u like to :)

torgeadelin: Technically everything inside the `Avatar.types` files will be a type, I'm thinking the extension is already enough maybe?

cscheffauer: gotcha - I think tough if types are exported from somewhere else than a `*.types` file (what could theoretically happen, it will be useful to mark them with the type keyboard. Also thinking from it of a perspective of Cantina as well :)

cscheffauer (approved):

PR #186, opened 2022-01-06
feat: create popover component

cscheffauer:
# Description

This PR adds a new Popover component to the Momentum library, which should act as a replacement for the Overlay trigger, etc. It could potentially replace the legacy popover component as well as the tooltip component.

It uses Tippy.js under the hood to achieve the positioning of the popover and other certain features.

Tests have been done in Storybook, within Chrome & Firefox (more tests in other browsers will follow).
Unit tests - in progress.

# Links

[Tippy.js](https://atomiks.github.io/tippyjs/v6/getting-started/)


src/components/Popover/Popover.types.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Determines if the Popover has interactive content inside of it,
+   * so that it can be hovered over and clicked inside without hiding.
+   *
+   * Default: false
+   */
+  interactive?: boolean;
+  /**
+   * Color of the Container
+   */
+  color?: Color;
```
torgeadelin, line 78: This type should come from ModalContainer.

src/components/Popover/Popover.unit.test.tsx
```diff
@@ -0,0 +1,265 @@
...
+            </button>
+          </>
+        );
+      };
+
+      const container = mount(<ParentComponent />);
+
+      expect(container.find(ContentContainer).length).toEqual(0);
+      container.find('#show').simulate('click');
+      container.update();
```
torgeadelin, line 261: You're not testing the hide

torgeadelin (changes requested):
Looks great so far! 
I'm going to be doing some live testing as well and will leave comments after.

Biggest blocker is the fact that there has been introduced a lot of code duplication around ModalContainer / ContentContainer components. I recommend removing the ContentContainer and reuse ModalContainer.

torgeadelin:
If the trigger on the component is `mousein`, then we automatically need to also have a trigger for `focusin`. This is an accessibility requirement.

torgeadelin:
When having the popover aligned to bottom-start/end, the arrow should be aligned towards the end, not towards the middle of the trigger.

image

torgeadelin:
I think we also need a delay for displaying a tooltip, especially for `mousein` triggers.

src/components/Popover/Popover.unit.test.tsx
cscheffauer: right, cause there were issues with the test atm. I will be working on the tests and then also add the hide. Thanks for reminding

src/components/Popover/Popover.types.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Possible values: 0 | 25 | 50 | 75 | 100 | 125 | 150
+   */
+  round?: Round;
+  /**
+   * elevation of the Container - each value corresponds to a box shadow applied to the container
+   *
+   * Possible values: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8
+   */
+  elevation?: Elevation;
+};
```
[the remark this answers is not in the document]
cscheffauer: removed

src/components/ModalArrow/ModalArrow.constants.ts
```diff
@@ -9,19 +10,37 @@ const COLORS: Record<string, Color> = {
...
+
+  // choose the side with most space:
+  AUTO: 'auto',
+  AUTO_START: 'auto-start',
+  AUTO_END: 'auto-end',
 };
 
 const DEFAULTS = {
   COLOR: COLORS.PRIMARY,
+  PLACEMENT: PLACEMENTS.BOTTOM as string,
```
InteractiveTimmy, line 35: Should this be `auto`?

src/components/ModalArrow/index.ts
```diff
@@ -1,6 +1,6 @@
 import { default as ModalArrow } from './ModalArrow';
 import * as CONSTANTS from './ModalArrow.constants';
-import { Props } from './ModalArrow.types';
+import type { Props } from './ModalArrow.types';
```
InteractiveTimmy, line 3: Is there a reason we're adding the `type` keyword?

src/components/ModalContainer/ModalContainer.stories.args.ts
```diff
@@ -1,18 +1,41 @@
...
+import { DEFAULTS } from './ModalContainer.constants';
 
 const modalContainerArgTypes = {
-  arrow: {
-    description: 'Where to render the `<ModalArrow />` for this component.',
+  showArrow: {
+    description: `Whether the arrow should be shown
+    <br/><br/>
+    NOTE: showing the arrow and using the placement prop will not position it correctly - positioning of the arrow
+    correctly will only be through the Popover component. Whenever there's the use case
```
InteractiveTimmy, line 12: `pls` => `please`

```
<br/><br/>default \`true\``
```

Doesn't need to exist, `argTypes.{prop}.table.defaultValue.summary` does this.

src/components/ModalContainer/ModalContainer.types.ts
```diff
@@ -1,15 +1,28 @@
...
+  /**
+   * Placement of the Modal relative to the trigger component. The
+   * arrow will be placed accordingly.
+   *
+   * Possible values: `top`, `top-start`, `top-end`, `right`, `right-start`, `right-end`, `bottom`, `bottom-start`, `bottom-end`,
+   * `left`, `left-start`, `left-end`, `auto`, `auto-start`, `auto-end`
+   *
+   * @default `bottom`
+   */
+  placement?: PlacementType;
```
InteractiveTimmy, line 25: It may be beneficial to move `ModalArrow` to a compound of `ModalContainer`

src/components/ModalContainer/index.ts
```diff
@@ -1,6 +1,6 @@
 import { default as ModalContainer } from './ModalContainer';
 import * as CONSTANTS from './ModalContainer.constants';
-import { Props } from './ModalContainer.types';
+import type { Props } from './ModalContainer.types';
```
InteractiveTimmy, line 3: Same as above, unnecessary keyword?

src/components/Popover/Popover.stories.args.ts
```diff
@@ -0,0 +1,88 @@
...
+  },
+  showArrow: {
+    description: `Whether the arrow should be shown`,
+    control: { type: 'boolean' },
+    table: {
+      type: {
+        summary: 'boolean',
+      },
+      defaultValue: {
+        summary: 'true',
```
InteractiveTimmy, line 38: Should be stored as a constant, doesn't need to be a string.

src/components/Popover/Popover.stories.args.ts
```diff
@@ -0,0 +1,88 @@
...
+  variant: {
+    description: `Variant of the Popover - can be either small or medium`,
+    options: ['small', 'medium'],
+    control: { type: 'select' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: 'small',
```
InteractiveTimmy, line 51: Constant please.

src/components/Popover/Popover.stories.args.ts
```diff
@@ -0,0 +1,88 @@
...
+      },
+    },
+  },
+  interactive: {
+    description: `Determines if the Popover has interactive content inside of it,
+    so that it can be hovered over and clicked inside without hiding.
+    <br/><br/>default \`false\``,
```
InteractiveTimmy, line 58: Goes into table.defaultValue.

src/components/Popover/Popover.stories.args.ts
```diff
@@ -0,0 +1,88 @@
...
+      },
+    },
+  },
+  children: {
+    description: 'Provides the child nodes for this element.',
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'ReactNode',
+      },
```
InteractiveTimmy, line 72: defaultValue === unedefined.

src/components/Popover/Popover.stories.args.ts
```diff
@@ -0,0 +1,88 @@
...
+  color: {
+    description: 'What color to render this `<Popover />` as.',
+    control: { type: 'select' },
+    options: [...Object.values(COLORS)],
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: 'undefined',
```
InteractiveTimmy, line 84: Here is the undefined that the above should have.

InteractiveTimmy (changes requested):
Some changes to Storybook, a few additional comments.

src/components/ModalContainer/ModalContainer.types.ts
cscheffauer: I would like to leave it as it is for now, since we haven't used the Compound Component pattern yet and it isn't a convention yet. It would need some training for the team first I think.
We can bring that up in a future conventions meeting and introduce that concept to Momentum 💯

src/components/ModalArrow/index.ts
cscheffauer: the type keyword is just a visual indicator - it has no effect on Typescript. I have used it before, since it makes it easier to distinguish between a normal import and a type import. I can remove it if that is too confusing :)

src/components/ModalArrow/ModalArrow.constants.ts
```diff
@@ -9,19 +10,37 @@ const COLORS: Record<string, Color> = {
...
 };
 
 const STYLE = {
-  wrapper: `${CLASS_PREFIX}-wrapper`,
```
torgeadelin, line 24: I think we should always keep the wrapper class even tho we're not using it. It's a good default selector for the component in testing.

src/components/ModalArrow/index.ts
torgeadelin: Let's discuss this in the conventions, I think we talked about this in a PR that I had last week.

src/components/ModalArrow/ModalArrow.constants.ts
cscheffauer: the arrow doesn't have any wrapper, since it is just a SVG, which gets rendered. That's why I have renamed it to svg to have it named properly.

src/components/ModalContainer/ModalContainer.stories.args.ts
```diff
@@ -1,18 +1,40 @@
...
+    to show ModalContainer with a arrow, pls use the Popover component instead.`,
```
torgeadelin, line 13: Still shows pls. Please change it to `please` haha

src/components/ModalContainer/ModalContainer.stories.args.ts
```diff
@@ -1,18 +1,40 @@
...
+        summary: DEFAULTS.SHOW_ARROW,
+      },
+    },
+  },
+  placement: {
+    description: `Placement of the Modal relative to the trigger component. 
+      <br/><br/>
+    NOTE: showing the arrow and using the placement prop will not position it correctly - positioning of the arrow
+    correctly will only be through the Popover component. Whenever there's the use case
+    to show ModalContainer with a arrow, pls use the Popover component instead.`,
```
torgeadelin, line 29: pls - > please

src/components/ModalContainer/ModalContainer.tsx
```diff
@@ -3,50 +3,62 @@ import classnames from 'classnames';
...
+import { ARROW_ID, DEFAULTS, STYLE } from './ModalContainer.constants';
+import type { Props } from './ModalContainer.types';
 import './ModalContainer.style.scss';
+import { getArrowOrientation } from './ModalContainer.utils';
 
 /**
- * The ModalContainer component.
+ * The `<ModalContainer/>` component, which is used by the Popover for example to wrap
+ * the content of a Popover properly.
+ *
```
torgeadelin, line 14: This is the same text as in the docs. Not sure it adds a lot of value to duplicate it here too. Keeping it in both location will increase the maintaining effort.

src/components/ModalContainer/ModalContainer.utils.ts
```diff
@@ -0,0 +1,22 @@
+import { PLACEMENTS } from '../ModalArrow/ModalArrow.constants';
+import { PlacementType } from '../ModalArrow/ModalArrow.types';
+
+type Orientation = 'horizontal' | 'vertical' | undefined;
+
+/**
+ * This function detects based on the placement arg if an the arrow
+ * is horizontal (left or right) or vertical (top or bottom)
+ *
+ * `undefined` if placement is `undefined`
```
torgeadelin, line 10: `undefined` if placement is `undefined` should be on line 12?

src/components/Popover/Popover.constants.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const CLASS_PREFIX = 'md-popover';
+
+const DEFAULTS = {
+  VARIANT: 'small',
+  TRIGGER: 'click',
+  PLACEMENT: PLACEMENTS.AUTO as string,
+  SHOW_ARROW: true,
+  INTERACTIVE: false,
+  COLOR: 'primary',
```
torgeadelin, line 11: use the constant from ModalContainer.constants.ts

src/components/Popover/Popover.stories.args.ts
```diff
@@ -0,0 +1,95 @@
...
+import { PLACEMENTS } from '../ModalArrow/ModalArrow.constants';
+import { COLORS } from '../ModalContainer/ModalContainer.constants';
+import { DEFAULTS } from './Popover.constants';
+
+export default {
+  ...commonStyles,
+  trigger: {
+    description: `Determines the events that cause the Popover to show. <br/><br/>
+    Multiple event names should be separated by spaces. For example to allow both click and hover, use \`click mouseenter\` as the trigger.
+    <br/><br/>Possible event names: \`click\`, \`mouseenter\`, \`focusin\`, \`manual\` (to programmatically trigger the popover)`,
```
torgeadelin, line 11: I'm not sure how I feel about the `<br/>`. I think we should let storybook handle the formatting.

torgeadelin: Also, if you wrap the string with '' you don't need to escape `\\``

src/components/Popover/Popover.stories.docs.mdx
```diff
@@ -0,0 +1,3 @@
+The `<Popover />` component allows adding a Popover to whatever provided
```
torgeadelin, line 1: `provided as`

src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+  triggerComponent: <ButtonSimple>Click me!</ButtonSimple>,
+};
+
+const InteractiveContent = Template<PopoverProps>(Popover).bind({});
+
+InteractiveContent.argTypes = { ...argTypes };
+
+InteractiveContent.args = {
+  trigger: 'click',
+  placement: 'bottom',
```
torgeadelin, line 48: use constant

src/components/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+    },
+    {
+      children: <p>Non-interactive Content on PRIMARY color, variant small, without arrow</p>,
+      trigger: 'mouseenter',
+      triggerComponent: (
+        <ButtonSimple style={{ height: '50px', width: '100px', marginRight: '10px' }}>
+          Hover me!
+        </ButtonSimple>
+      ),
+      placement: 'bottom-start',
```
torgeadelin, line 95: Constant here too, use it everywhere, I won't leave a comment in all the places.

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <LazyTippy
+      ref={tippyRef}
+      render={(attrs) => (
+        <ModalContainer
+          id={id}
+          showArrow={showArrow}
+          placement={attrs['data-placement']}
+          isPadded
+          round={variant === 'medium' ? ROUNDS[75] : ROUNDS[50]}
```
torgeadelin, line 53: Why this `variant` prop? Does the design spec only allow 2 variants for Popovers in terms of the border radius?

src/components/Popover/Popover.types.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export type PlacementType = TippyProps['placement'];
+export type TriggerType = TippyProps['trigger'];
+
+export type PopoverInstance = {
+  show: () => void;
+  hide: () => void;
+};
+
+export interface Props {
```
torgeadelin, line 16: I think we have a rule to leave empty space between types here. Not sure if it's enforced by eslint.

src/components/Popover/Popover.unit.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+    it('should have provided attributes when attributes are provided', async () => {
+      expect.assertions(5);
+      const className = 'example-class';
+      const style = { color: 'pink' };
+      const styleString = 'color: pink;';
+      const id = 'example-id';
+
+      render(
+        <Popover
+          triggerComponent={<button>Click Me!</button>}
```
torgeadelin, line 89: I see why you did this, but the purpose of separating all the attribute tests in tiny individual tests, is to have a better report when tests are failing. If only one is failing, reading it from the console might not be as obvious, because the test is called `should have provided attributes when attributes are provided`, but if we have them separated, will be easy to spot which one is actually failing from the test name.

src/index.js
```diff
@@ -144,6 +144,7 @@ export {
   AvatarMeetingsListItem,
   SpaceListItem as SpaceListItemNext,
   List as ListNext,
+  Popover as PopoverNext,
```
torgeadelin, line 147: I think we should export the placements too maybe? Not sure.

torgeadelin (changes requested):

src/components/Popover/Popover.tsx
cscheffauer: jup - in Figma there are 2 variants of the popover, one small and one medium.

src/components/Popover/Popover.unit.test.tsx
cscheffauer: testing library does give u a really in depth error message with the line it is appearing and more. So I think with RTL we would need to re-evaluate if splitting these tests is valuable.

src/index.js
cscheffauer: i will review that once I'm reusing the Popover component in other places

src/components/Popover/Popover.stories.args.ts
cscheffauer: multiline doesn't work with ", that's why ` is used and the escape is needed.

cscheffauer: to the <br/> - without it the description would look like this (no line breaks or anything between possible event names and so on:
image

src/components/ModalContainer/ModalContainer.tsx
cscheffauer: will remove it here then, since the docs will be readable by the consumers as well.

src/components/ModalContainer/ModalContainer.stories.args.ts
cscheffauer: Alt Text

torgeadelin (approved):
Looks good to me.
Great work Christopher! 🍻

src/components/Popover/Popover.stories.args.ts
torgeadelin: I think that's fine tho.

src/components/ModalArrow/ModalArrow.constants.ts
InteractiveTimmy: Any comments/details on this?

InteractiveTimmy (approved):
Looks good to me! thank you for making the changes.

PR #188, opened 2022-01-10
Fix/theme provider/style attribute

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to add improvements to the theme provider that allow for a better viewing experience within storybook.

## Screenshots

Screen Shot 2022-01-10 at 2 32 44 PM
Screen Shot 2022-01-10 at 2 33 07 PM
Screen Shot 2022-01-10 at 2 33 12 PM

# Links

[SPARK-296628](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-296628)

### DIFF
--- a/config/storybook/cjs/preview.js
+++ b/config/storybook/cjs/preview.js
@@ -10,6 +10,7 @@ export const parameters = {
       date: /Date$/,
     },
   },
+  layout: 'fullscreen',
 };
 
 export const globalTypes = { ...globals };
--- a/src/components/ThemeProvider/ThemeProvider.stories.args.ts
+++ b/src/components/ThemeProvider/ThemeProvider.stories.args.ts
@@ -18,7 +18,6 @@ const themeProviderArgTypes = {
     },
   },
   children: {
-    defaultValue: 'Example Text',
     description:
       'Provides the child nodes for this element. The child nodes of this element are themed by the assigned theme.',
     control: { type: 'text' },
--- a/src/components/ThemeProvider/ThemeProvider.stories.args.ts
+++ b/src/components/ThemeProvider/ThemeProvider.stories.args.ts
@@ -1,0 +1,41 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+
+import { THEME_PROVIDER_CONSTANTS as CONSTANTS } from './';
+
+const themeProviderArgTypes = {
+  theme: {
+    description:
+      'The name of the target theme to apply to all child nodes of this `<ThemeProvider />`.',
+    options: [...Object.values(CONSTANTS.THEME_NAMES)],
+    control: { type: 'select' },
+    table: {
+      type: {
+        sumary: 'string',
+      },
+      defaultValue: {
+        summary: CONSTANTS.DEFAULTS.THEME,
+      },
+    },
+  },
+  children: {
+    defaultValue: 'Example Text',
+    description:
+      'Provides the child nodes for this element. The child nodes of this element are themed by the assigned theme.',
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'ReactNode',
+      },
+      defaultValue: {
+        summary: 'undefined',
+      },
+    },
+  },
+};
+
+export { themeProviderArgTypes };
+
+export default {
+  ...commonStyles,
+  ...themeProviderArgTypes,
+};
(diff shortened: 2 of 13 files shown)


config/storybook/cjs/preview.js
```diff
@@ -10,6 +10,7 @@ export const parameters = {
       date: /Date$/,
     },
   },
+  layout: 'fullscreen',
```
cscheffauer, line 13: will that set storybook to a fullscreen mode by default?

src/components/ThemeProvider/ThemeProvider.stories.args.ts
```diff
@@ -0,0 +1,41 @@
...
+      type: {
+        sumary: 'string',
+      },
+      defaultValue: {
+        summary: CONSTANTS.DEFAULTS.THEME,
+      },
+    },
+  },
+  children: {
+    defaultValue: 'Example Text',
```
cscheffauer, line 21: is `Example Text` really a good defaultValue for children? I have seen that before in other places. Was just wondering if that could be changed to something more meaningful :)

src/components/ThemeProvider/ThemeProvider.stories.docs.mdx
```diff
@@ -0,0 +1 @@
+The `<ThemeProvider />` component. This component is used to theme all child nodes within. This component primiarily utilizes **CSS Variables** for downstream components to consume and is completely independently stylable.
```
cscheffauer, line 1: like that a lot - short, but really descriptive! 💯

src/components/ThemeProvider/ThemeProvider.stories.tsx
```diff
@@ -1,79 +1,86 @@
...
+        Example Text
       </div>
-    </ThemeProvider>
-  );
+    </div>
+  ),
 };
 
-const Example = Template.bind({});
+// const Template: Story<Props> = (args) => {
```
cscheffauer, line 52: why is that commented code still necessary to keep? I would suggest removing it if there is no real good reason to keep it.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -9,10 +9,11 @@ import './ThemeProvider.style.scss';
 /**
  * Provides a collection of CSSVariables based on a ThemeToken to all child elements.
  */
-const ThemeProvider: FC<Props> = ({ children, id, theme }: Props) => (
+const ThemeProvider: FC<Props> = ({ children, id, style, theme }: Props) => (
   <div
```
cscheffauer, line 13: i would recommend adding a comment to the component to state that the ThemeProvider is wrapping the children in a div. There are solutions out there, where no bounding div will be wrapped around children, so I think it will be worth adding it to the docs.

Just as an example - if theme is passed to children through a React.Context, the children will not be wrapped in a div :)

src/components/ThemeProvider/ThemeProvider.unit.test.tsx
```diff
@@ -52,6 +62,19 @@ describe('<ThemeProvider />', () => {
       expect(element.classList.contains(`${THEME_CLASS_PREFIX}-${themeName}`)).toBe(true);
     });
 
+    it('should have provided style when style is provided', () => {
+      expect.assertions(1);
+
+      const style = { color: 'pink' };
+      const styleString = 'color: pink;';
```
cscheffauer, line 69: just out of curiosity - would it work if u do `const styleString = style.toString();` ?

config/storybook/cjs/preview.js
InteractiveTimmy: This removes padding within stories.

src/components/ThemeProvider/ThemeProvider.stories.args.ts
InteractiveTimmy: Removing this.

src/components/ThemeProvider/ThemeProvider.unit.test.tsx
InteractiveTimmy: styleString would render as `{ "color": "pink" }` with the `toString()` method.

src/components/ThemeProvider/ThemeProvider.tsx
InteractiveTimmy: Due to the layer we are applying the theme at [Browser/DOM CSS level], using a `React.Context` would require themed data to be handled differently in downstream components [not via css variables].

Adding the comment mentioned here and updating.

PR #192, opened 2022-01-17
Spark 287037 dot compact mode

shnaaz:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-287037

image

### DIFF
--- a/src/components/SpaceListItem/SpaceListItem.tsx
+++ b/src/components/SpaceListItem/SpaceListItem.tsx
@@ -8,6 +8,7 @@ import ListItemBase from '../ListItemBase';
 import ListItemBaseSection from '../ListItemBaseSection';
 import Text from '../Text';
 import Icon from '../Icon';
+import DividerDot from '../DividerDot';
 
 //TODO: support 2-line labels for right/position-end section.
 /**
@@ -46,8 +47,7 @@ const SpaceListItem: FC<Props> = forwardRef(
             <Text type="body-primary" data-test="list-item-first-line">
               {firstLine}
             </Text>
-            {/* //TODO: change with dot divider when available */}
-            {isCompact && <span className={STYLE.dotDivider}> - </span>}
+            {isCompact && <DividerDot data-test="compact-mode-divider-dot" />}
             <Text
               style={{ color: `var(--theme-text-team-${teamColor}-normal)` }}
               type="body-secondary"
--- a/src/components/SpaceListItem/SpaceListItem.unit.test.tsx
+++ b/src/components/SpaceListItem/SpaceListItem.unit.test.tsx
@@ -7,7 +7,6 @@ import Avatar from '../Avatar';
 import Icon from '../Icon';
 import { mountAndWait } from '../../../test/utils';
 import ListItemBase from '../ListItemBase';
-import DividerDot from '../DividerDot';
 import * as ListContext from '../List/List.utils';
 import { STYLE } from './SpaceListItem.constants';
 
--- a/src/components/SpaceListItem/SpaceListItem.unit.test.tsx
+++ b/src/components/SpaceListItem/SpaceListItem.unit.test.tsx
@@ -7,7 +7,7 @@ import Avatar from '../Avatar';
 import Icon from '../Icon';
 import { mountAndWait } from '../../../test/utils';
 import ListItemBase from '../ListItemBase';
-
+import DividerDot from '../DividerDot';
 import * as ListContext from '../List/List.utils';
 import { STYLE } from './SpaceListItem.constants';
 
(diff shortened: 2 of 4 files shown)


src/components/SpaceListItem/SpaceListItem.unit.test.tsx
```diff
@@ -255,6 +254,18 @@ describe('<SpaceListItem />', () => {
       expect(element.textContent).toBe('one - two');
     });
 
+    it('should have provided dot in compact mode', async () => {
+      expect.assertions(1);
```
marcin-bazyl, line 258: as far as I understand this, this line is not needed, because we're using async/await, right?

cscheffauer:
I would recommend adding a story for it to showcase it and test it.

src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -46,8 +47,7 @@ const SpaceListItem: FC<Props> = forwardRef(
             <Text type="body-primary" data-test="list-item-first-line">
               {firstLine}
             </Text>
-            {/* //TODO: change with dot divider when available */}
-            {isCompact && <span className={STYLE.dotDivider}> - </span>}
+            {isCompact && <DividerDot data-test="compact-mode-divider-dot" />}
```
cscheffauer, line 50: the divider should be between the first and second line?
I thought the divider should replace the `_secondLineArray.join ...` in line 57 (that's what the comment there is saying)

cscheffauer:
I would recommend refactoring the SpaceListItem component as well. For example the renderText function can be simplified in one JSX tag. Maybe it could be abstracted as a sub component (in the same folder). let me know if u have questions about it.

src/components/SpaceListItem/SpaceListItem.unit.test.tsx
cscheffauer: this line is a convention @InteractiveTimmy introduced. It helps to assert if the amount of expected assertions inside of a test is there. It is also helps identifying how many assertions are done in a test. We could discuss this in a conventions meeting if we wanna continue this or not :)

src/components/SpaceListItem/SpaceListItem.unit.test.tsx
```diff
@@ -255,6 +254,18 @@ describe('<SpaceListItem />', () => {
...
+
+      const secondLine = 'testteam';
+
+      const container = await mountAndWait(
+        <SpaceListItem firstLine="firstLine" secondLine={secondLine} isCompact={true} />
+      );
+
+      expect(container.find("[data-test='compact-mode-divider-dot']").exists()).toBe(true);
```
cscheffauer, line 266: would there be a way in Enzyme to test if the divider is placed between the `firstLine` and `secondLine`? that would be a more meaningful test imho. since that is what the code does at the moment.

src/components/SpaceListItem/SpaceListItem.tsx
shnaaz: That is to divide multiple text in second line which is not consumed by Cantina atm. Will create separate story in momentum to cover that.

shnaaz:
> I would recommend adding a story for it to showcase it and test it.

There is already a story, this screenshot is from that.

shnaaz:
> I would recommend refactoring the SpaceListItem component as well. For example the renderText function can be simplified in one JSX tag. Maybe it could be abstracted as a sub component (in the same folder). let me know if u have questions about it.

Refactoring could be part of a separate story. Feel free to create one.

src/components/SpaceListItem/SpaceListItem.unit.test.tsx
```diff
@@ -255,6 +255,32 @@ describe('<SpaceListItem />', () => {
...
+      const container = await mountAndWait(
+        <SpaceListItem firstLine="firstLine" secondLine={secondLine} isCompact={true} />
+      );
+
+      expect(
+        container.find(DividerDot).filter("[data-test='compact-mode-divider-dot']").props()
+      ).toEqual({ 'data-test': 'compact-mode-divider-dot' });
+    });
+
+    it('checks divider dot position', async () => {
```
torgeadelin, line 272: this test should be part of the `describe('snapshot')` suite.

src/components/SpaceListItem/SpaceListItem.unit.test.tsx
```diff
@@ -255,6 +255,32 @@ describe('<SpaceListItem />', () => {
       expect(element.textContent).toBe('one - two');
     });
 
+    it('should have provided dot in compact mode', async () => {
+      expect.hasAssertions();
```
torgeadelin, line 259: Please use `expect.assertions(1)` for consistency

PR #193, opened 2022-01-18
Feat/components/new overlay

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to migrate the `<Overlay />` component to be a focus-based overlay/fade as originally intended. This is inclusive of removing the redundant `<OverlayTrigger />` in favor of the new `<Popover />` component.

The new component is inclusive of only the `Overlays and Fades` section in the linked Figma.

At the time of this PRs creation, our projects are not actively consuming this component.

Additional changes to plop and package configs are included in this PR.

# Screenshots

Screen Shot 2022-01-18 at 5 19 30 PM
Screen Shot 2022-01-18 at 5 19 24 PM
Screen Shot 2022-01-18 at 5 19 20 PM

# Links

[SPARK-253323](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-253323).
[Figma](https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp/Components---Windows?node-id=9994%3A0).


src/components/Overlay/Overlay.stories.docs.mdx
```diff
@@ -1,3 +1,24 @@
...
 
-For additional information on how the WAI-ARIA in this component operates, view this [React-Aria documentation](https://react-spectrum.adobe.com/react-aria/useDialog.html).
\ No newline at end of file
+Note that the render state of this component must be managed manually from within its child components. This includes any action that is used to remove/deconstruct the `<Overlay />` component.
+
+Example:
+
+```jsx
+const App = () => {
+  const [overlayOpen, setOverlayOpen] = useState(false);
```
cscheffauer, line 9: in this example the `overlayOpen` state isn't used - i would recommend using it as well to showcase how the Overlay component can be shown conditionally.

src/components/Overlay/Overlay.tsx
```diff
@@ -1,100 +1,35 @@
...
-    isOpen = DEFAULTS.IS_OPEN,
-    onClose,
-    autoFocus = DEFAULTS.AUTO_FOCUS,
-    contain = DEFAULTS.CONTAIN,
-    positioning,
-    restoreFocus = DEFAULTS.RESTORE_FOCUS,
+    className,
+    color = DEFAULTS.COLOR,
+    fullscreen = DEFAULTS.FULLSCREEN,
+    height = DEFAULTS.HEIGHT,
```
cscheffauer, line 17: i would recommend using a different naming for this - cause `height` can be easily misunderstood as the pixel height on the screen. In this case we allow passing in the zIndex - therefore the prop should be named `zIndex` or similar imho

cscheffauer: as a note - it is not necessary to have that zIndex prop, since the className is applied on the same div, so consumers could also override the zIndex through a class

cscheffauer:
I really like the storybook example for the overlay - awesome stuff! 💯

src/components/Overlay/Overlay.tsx
InteractiveTimmy: This is fixed, you now must use the `style` prop to inject additional styles, such as z-index. `height` is now a removed prop.

src/components/Overlay/Overlay.stories.docs.mdx
InteractiveTimmy: Fixed and added to the example.

cscheffauer (approved):

PR #194, opened 2022-01-20
Feat/components/add disabled button control

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to add the disabled ui state to `<ButtonControl />`.

# Screenshots

Screen Shot 2022-01-20 at 5 33 11 PM
Screen Shot 2022-01-20 at 5 33 05 PM
Screen Shot 2022-01-20 at 5 32 58 PM
Screen Shot 2022-01-20 at 5 32 52 PM

# Links

[SPARK-299712](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-299712)

### DIFF
--- a/src/components/ButtonControl/ButtonControl.style.scss
+++ b/src/components/ButtonControl/ButtonControl.style.scss
@@ -1,14 +1,10 @@
 .md-button-control-wrapper {
   align-items: center;
-  background-color: rgba(0, 0, 0, 0); // IEFIX
   background-color: var(--button-ghost-background);
-  border: 0rem solid rgba(0, 0, 0, 0); // IEFIX
   border: var(--md-globals-border-clear);
-  color: rgba(0, 0, 0, 0.95); // IEFIX
   color: var(--button-ghost-text);
   cursor: pointer;
   display: flex;
-  font-family: CiscoSansTT; // IEFIX
   font-family: var(--md-globals-font-default);
   height: 2rem;
   justify-content: center;
@@ -19,17 +15,29 @@
   width: 2rem;
 
   &:hover {
-    background-color: rgba(0, 0, 0, 0.07); // IEFIX
     background-color: var(--button-ghost-background-hovered);
-    color: rgba(0, 0, 0, 0.95); // IEFIX
     color: var(--button-ghost-text-hovered);
   }
 
   &:active {
-    background-color: rgba(0, 0, 0, 0.2); // IEFIX
     background-color: var(--button-secondary-background-pressed);
-    color: rgba(0, 0, 0, 0.95); // IEFIX
     color: var(--button-secondary-text-pressed);
+  }
+
+  &[data-disabled='true'] {
+    background-color: var(--button-ghost-background);
+    color: var(--button-primary-text-disabled);
+    cursor: auto;
+
+    &:hover {
+      background-color: var(--button-ghost-background);
+      color: var(--button-primary-text-disabled);
+    }
+
+    &:active {
+      background-color: var(--button-ghost-background);
+      color: var(--button-primary-text-disabled);
+    }
   }
 
   &[data-circular='true'] {
--- a/src/components/ButtonControl/ButtonControl.tsx
+++ b/src/components/ButtonControl/ButtonControl.tsx
@@ -1,8 +1,8 @@
-import React, { cloneElement, FC } from 'react';
+import React, { FC } from 'react';
 import classnames from 'classnames';
 
 import ButtonSimple from '../ButtonSimple';
-import Icon from '../Icon';
+import Icon, { ICON_CONSTANTS, IconProps } from '../Icon';
 
 import { ICONS, STYLE } from './ButtonControl.constants';
 import { Props } from './ButtonControl.types';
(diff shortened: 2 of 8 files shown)


src/components/ButtonControl/ButtonControl.style.scss
```diff
@@ -19,19 +15,31 @@
...
-    background-color: rgba(0, 0, 0, 0.2); // IEFIX
     background-color: var(--button-secondary-background-pressed);
-    color: rgba(0, 0, 0, 0.95); // IEFIX
     color: var(--button-secondary-text-pressed);
   }
 
+  &[data-disabled='true'] {
+    background-color: var(--button-ghost-background);
+    color: var(--button-primary-text-disabled);
+    cursor: auto;
```
cscheffauer, line 30: so auto uses the browser default - would it make more sense to do a `not allowed` for the disabled state?

src/components/ButtonControl/ButtonControl.tsx
```diff
@@ -15,26 +15,30 @@ const ButtonControl: FC<Props> = (props: Props) => {
...
-
-  let controlElement = <Icon name={iconName} weight="bold" autoScale />;
+  const iconProps: IconProps = {
+    autoScale: true,
+    name: iconName,
+    weight: ICON_CONSTANTS.WEIGHTS.bold,
+  };
 
   // Unique icon configurations
   switch (iconName) {
```
cscheffauer, line 25: theres a switch with only one case - this should be refactored to a if statement

src/components/ButtonControl/ButtonControl.tsx
```diff
@@ -15,26 +15,30 @@ const ButtonControl: FC<Props> = (props: Props) => {
...
 
   return (
     <ButtonSimple
       className={classnames(STYLE.wrapper, className)}
       data-circular={isCircular}
+      data-disabled={props.isDisabled}
       {...otherProps}
     >
-      {controlElement}
+      <Icon {...iconProps} />
```
cscheffauer, line 41: instead of constructing the iconProps on the top and then modifying it once, it would be actually better to do it as a inline JSX here - like: 
```
{ iconName === ICONS.favorite ? 
< Icon with the props to apply />
:
< Icon with the other props to apply />
```

src/components/ButtonControl/ButtonControl.unit.test.tsx
```diff
@@ -52,6 +52,14 @@ describe('<ButtonControl />', () => {
 
       expect(container).toMatchSnapshot();
     });
+
+    it('should match snapshot with isDisabled', async () => {
```
cscheffauer, line 56: as of now with enzyme we do these attribute tests - i think that attribute test for disabled is missing here

src/components/ButtonControl/ButtonControl.tsx
```diff
@@ -15,26 +15,30 @@ const ButtonControl: FC<Props> = (props: Props) => {
...
+        ? 'var(--button-primary-text-disabled)'
+        : 'var(--theme-text-warning-normal)';
       break;
   }
 
   return (
     <ButtonSimple
       className={classnames(STYLE.wrapper, className)}
       data-circular={isCircular}
+      data-disabled={props.isDisabled}
```
cscheffauer, line 38: `isDisabled` should be destructured on the top as the other props to keep it consistent i think

src/components/ButtonControl/ButtonControl.style.scss
InteractiveTimmy: This is something we'll need to discuss, but other disabled states use the auto for button's being disabled.

src/components/ButtonControl/ButtonControl.tsx
InteractiveTimmy: It's set to switch in the case of a smaller refactor of adding additional buttons [i.e. if we add an announcement button, etc].

This can be switched to an if for now as described in this comment.

src/components/ButtonControl/ButtonControl.tsx
InteractiveTimmy: This does increase the amount of refactor work in the case that we need to add additional custom icon props for different icons.

I will make sure to make the change to the commented details though.

src/components/ButtonControl/ButtonControl.unit.test.tsx
InteractiveTimmy: Since these tests are explicitly UI, and the attribute is tested downstream [in `<ButtonSimple />`], there's no need to add a redundant attribute test.

src/components/ButtonControl/ButtonControl.tsx
InteractiveTimmy: Destructuring it at the top requires it to be passed into `<ButtonSimple />` again:

```jsx
return (
  <ButtonSimple
    /* ...other props... */
    data-disabled={isDisabled}
    isDisabled={isDisabled}
    /* ...otherProps... */
  />
```

cscheffauer: ah i see - that makes sense then! :)

cscheffauer (approved):

PR #196, opened 2022-01-24
Chore/meeting list item/improvements

torgeadelin:
# Description

Added a `GrayStatic` variation for the `MeetingListItem` component.
Only difference is that is has no hover color (compared to `Gray`)
# Screenshot

image


src/components/MeetingListItem/MeetingListItem.style.scss
```diff
@@ -22,6 +22,17 @@
     }
   }
 
+  &[data-color='GrayStatic'] {
```
InteractiveTimmy, line 25: Interesting color name choice.

InteractiveTimmy (approved):
Looks good! I don't think we need tests for this change either [unless someone thinks otherwise]. Awesome job.

PR #198, opened 2022-01-24
feat(components/popover): add manual triggers

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to amend the abstracted functional logic from `Tippy.js` for externally handling popovers from outside of the provided trigger component.

# Links

[SPARK-300653](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-300653)

### DIFF
--- a/src/components/Popover/Popover.stories.args.ts
+++ b/src/components/Popover/Popover.stories.args.ts
@@ -8,7 +8,7 @@ export default {
   trigger: {
     description: `Determines the events that cause the Popover to show.
     Multiple event names should be separated by spaces. For example to allow both click and hover, use \`click mouseenter\` as the trigger.
-    Possible event names: \`click\`, \`mouseenter\`, \`focusin\`, \`manual\` (to programmatically trigger the popover)`,
+    Possible event names: \`click\`, \`mouseenter\`, \`focusin\`, \`manual\` (to programmatically trigger the popover, this disables the popover-close events on DOM focus change)`,
     control: { type: 'text' },
     table: {
       type: {
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -46,6 +46,8 @@ const Popover: FC<Props> = (props: Props) => {
   return (
     <LazyTippy
       ref={tippyRef}
+      /* needed to prevent the popover from closing when the focus is changed via click events */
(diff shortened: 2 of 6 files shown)


torgeadelin (approved):
lgtm!

src/components/Popover/Popover.stories.args.ts
```diff
@@ -8,7 +8,7 @@ export default {
   trigger: {
     description: `Determines the events that cause the Popover to show.
     Multiple event names should be separated by spaces. For example to allow both click and hover, use \`click mouseenter\` as the trigger.
-    Possible event names: \`click\`, \`mouseenter\`, \`focusin\`, \`manual\` (to programmatically trigger the popover)`,
+    Possible event names: \`click\`, \`mouseenter\`, \`focusin\`, \`manual\` (to programmatically trigger the popover, this disabled on-focus-change popover-close events)`,
```
cscheffauer, line 11: what does this line mean? i don't understand the last part :D

src/components/Popover/Popover.tsx
```diff
@@ -46,6 +46,7 @@ const Popover: FC<Props> = (props: Props) => {
   return (
     <LazyTippy
       ref={tippyRef}
+      hideOnClick={!trigger.includes('manual')}
```
cscheffauer, line 49: would recommend adding a short description above it why this is necessary

src/components/Popover/Popover.types.ts
```diff
@@ -90,6 +91,11 @@ export interface Props {
    */
   setInstance?: React.Dispatch<React.SetStateAction<PopoverInstance>>;
 
+  /**
+   * Whether or not this component is visible on render.
+   */
+  visible?: boolean;
```
cscheffauer, line 97: this prop should be removed, since its not reliable - we should use the popoverInstance.show() or .hide() for it instead (as discussed)

src/components/Popover/Popover.tsx
InteractiveTimmy: added.

src/components/Popover/Popover.types.ts
InteractiveTimmy: removed, forgot to remove this from my local changes before pushing.

cscheffauer (approved):

PR #199, opened 2022-01-26
fix(components/button-simple): Not bubble up onClick

cscheffauer:
# Description

Consuming the ButtonSimple (and its variants) within the Popover led to an issue where the click event was catched correctly as a trigger by the Popover, but tabbing and pressing enter on a ButtonSimple didn't cause a onClick event.

The problem is related to React-aria, where the onKeyDown event on a button results in a onPress and doesn't bubble up the onClick event as you would expect from a normal HTML button element. 

To fix this, we are going to override the onKeyDown coming from react-aria, if we need to (for example when using it on a Popover).

# Links

[React-aria issue](https://github.com/adobe/react-spectrum/issues/2100)

### DIFF
--- a/src/components/ButtonControl/ButtonControl.types.ts
+++ b/src/components/ButtonControl/ButtonControl.types.ts
@@ -1,31 +1,15 @@
-import { CSSProperties } from 'react';
-import { AriaButtonProps } from '@react-types/button';
+import type { ButtonSimpleProps } from '../ButtonSimple';
 
 export type ButtonControlControl = 'close' | 'favorite' | 'maximize' | 'minimize' | 'mute';
 
-export interface Props extends AriaButtonProps {
-  /**
-   * Custom class for overriding this component's CSS.
-   */
-  className?: string;
-
+export interface Props extends ButtonSimpleProps {
   /**
    * The control type.
    */
   control: ButtonControlControl;
 
   /**
-   * Custom id for overriding this component's CSS.
-   */
-  id?: string;
-
-  /**
    * Whether to render this button as a circle.
    */
   isCircular?: boolean;
-
-  /**
-   * Custom style for overriding this component's CSS.
-   */
-  style?: CSSProperties;
 }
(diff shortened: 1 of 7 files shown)


src/components/ButtonControl/ButtonControl.types.ts
```diff
@@ -1,31 +1,15 @@
...
 
 export type ButtonControlControl = 'close' | 'favorite' | 'maximize' | 'minimize' | 'mute';
 
-export interface Props extends AriaButtonProps {
-  /**
-   * Custom class for overriding this component's CSS.
-   */
-  className?: string;
-
+export interface Props extends ButtonSimpleProps {
```
InteractiveTimmy, line 5: good migration.

src/components/ButtonSimple/ButtonSimple.types.ts
```diff
@@ -27,4 +27,11 @@ export interface Props extends AriaButtonProps, HoverProps {
    * title to use for this component.
    */
   title?: string;
+
+  /**
+   * Use the native onKeyDown event handler to allow `enter` & `space` keypress events fire a onClick (like a native HTML button does)
+   *
+   * This is necessary since `react-aria` supress that behaviour by design.
+   */
+  useNativeKeyDown?: boolean;
```
InteractiveTimmy, line 36: This needs to be added to `argTypes` for the story.

InteractiveTimmy (changes requested):
Looks good! just needs argTypes added for the new prop.

src/components/ButtonSimple/ButtonSimple.stories.args.ts
```diff
@@ -31,6 +31,19 @@ const buttonSimpleArgTypes = {
       },
     },
   },
+  useNativeKeyDown: {
```
InteractiveTimmy, line 34: Thank you, looks good!

InteractiveTimmy (approved):
LGTM! thanks for adding the argtypes!

PR #208, opened 2022-02-24
feat(component-variant/solid-button-2): adding styles for circle and pill solid outline button

nataliadelmar:
# Description

PR to add the variant of outline circle and pill buttons with non-ghost or solid background, as specified here:
image

# Links
[*Figma link*](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=6361%3A0)

# Important notes
* New components tokens were created for this with the prefix --videoSupport. The reason why the name is like that is because this component is located in page Realtime > Video support in the Figma library, for now. We have requested Momentum team to move these solid buttons to the buttons sections to keep consistency. They have it on their backlog. To keep consistency with Momentum as it is right now, the component token was named with the actual component location in mind (--videoSupport). When the Momentum team updates this, we will update the token and MRV2 as well.
* Button Pill already had an outline solid variant, but **theme** tokens were used for it. Theme tokens were replaced with the new components tokens --videoSupport as part of the PR.
* Button Circle did not have an outline non-ghost variant, so that was added to it. It is relevant to say that the previous outline button circle assumed that the button was ghost, which is not always the case. As consequence of this PR (if it gets merged), I will update all Circle buttons outline in Cantina for them to have ghost=false (as this is the first time we would be using outline=true && ghost=true variant.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "@momentum-ui/animations": "^1.3.0",
-    "@momentum-ui/design-tokens": "^0.0.28",
+    "@momentum-ui/design-tokens": "^0.0.34",
     "@momentum-ui/icons": "8.27.0",
     "@momentum-ui/icons-rebrand": "^1.22.0",
     "@momentum-ui/tokens": "^1.7.1",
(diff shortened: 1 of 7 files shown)


InteractiveTimmy (approved):
LGTM! Since this is strictly a style change, we wont need to amend tests.

PR #212, opened 2022-03-02
fix(components/list): focus swap on child change

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to fix the list focus issue in which children are updated within the list.

# Links

[SPARK-305639](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-305639)

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useRef, useState, useCallback } from 'react';
+import React, { FC, ReactNode, useEffect, useRef, useState, useCallback } from 'react';
 import classnames from 'classnames';
 
 import { STYLE } from './List.constants';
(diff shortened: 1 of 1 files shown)


src/components/List/List.tsx
```diff
@@ -9,10 +9,19 @@ import { useKeyboard } from '@react-aria/interactions';
 
 const List: FC<Props> = (props: Props) => {
   const [currentFocus, setCurrentFocus] = useState<number>(0);
+  const [currentFocusChild, setCurrentFocusChild] = useState<ReactNode>(
```
torgeadelin, line 12: Maybe add a comment to why this is needed. After reading the code is not that obvious?

torgeadelin (approved):
Looking good!

src/components/List/List.tsx
InteractiveTimmy: Added comments.

PR #215, opened 2022-03-10
Feat/menu trigger/refactor

torgeadelin:
# Description
Refactor of `MenuTrigger` component to take advantage of the new `Popover ` component. There are no behaviour  nor accessibility changes. (Some improvements for accessibility).

# Links
[SPARK-300539](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-300539)


src/components/MenuTrigger/MenuTrigger.constants.ts
```diff
@@ -1,13 +1,28 @@
 const CLASS_PREFIX = 'md-menu-trigger';
+import { PLACEMENTS } from '../ModalArrow/ModalArrow.constants';
+import { COLORS } from '../ModalContainer/ModalContainer.constants';
+import type { PlacementType } from '../ModalArrow/ModalArrow.types';
+
+const MENU_TRIGGER_PLACEMENTS: Record<string, PlacementType> = PLACEMENTS;
+delete MENU_TRIGGER_PLACEMENTS.LEFT;
```
cscheffauer, line 7: i think instead of deleting it you could iterate over the PLACEMENTS and filter them like `placement.startsWith('top')` etc.

src/components/MenuTrigger/MenuTrigger.constants.ts
```diff
@@ -1,13 +1,28 @@
...
+delete MENU_TRIGGER_PLACEMENTS.RIGHT_END;
+delete MENU_TRIGGER_PLACEMENTS.AUTO;
+delete MENU_TRIGGER_PLACEMENTS.AUTO_START;
+delete MENU_TRIGGER_PLACEMENTS.AUTO_END;
 
 const DEFAULTS = {
-  BACKGROUND: 'primary' as const,
-  OVERLAY_RADIUS: 75 as const,
+  VARIANT: 'medium',
+  PLACEMENT: MENU_TRIGGER_PLACEMENTS.BOTTOM_START as string,
```
cscheffauer, line 19: is casting as string necessary here?

src/components/MenuTrigger/MenuTrigger.stories.args.ts
```diff
@@ -23,8 +24,38 @@ export default {
         summary: 'boolean',
       },
       defaultValue: {
-        summary: 'false',
+        summary: 'true',
       },
     },
   },
+  isOpen: {
+    description: 'Whether the overlay is open by default (controlled).',
```
cscheffauer, line 32: the description might not 100% correct here. It is not only controlling if default open, but you can also change the isOpen later on to open and close it from the outside correct?

src/components/MenuTrigger/MenuTrigger.stories.tsx
```diff
@@ -31,12 +32,15 @@ const Example = Template<MenuTriggerProps>(MenuTrigger).bind({});
 Example.argTypes = { ...argTypes };
 
 Example.args = {
-  closeOnSelect: false,
   'aria-label': 'Menu trigger',
-  children: [
+  closeOnSelect: true,
+  onOpenChange: action('onOpenChange'),
+  triggerComponent: (
     <ButtonPill key="1">
```
cscheffauer, line 39: why are those key's necessary?

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -1,18 +1,21 @@
...
 import { useMenuTrigger } from '@react-aria/menu';
 import Menu, { MenuContext } from '../Menu';
-import { DismissButton, useOverlay } from '@react-aria/overlays';
+import { DismissButton } from '@react-aria/overlays';
 import { verifyTypes } from '../../helpers/verifyTypes';
 import { FocusScope } from '@react-aria/focus';
-import ModalContainer from '../ModalContainer';
-import ContentSeparator from '../ContentSeparator';
 import { useKeyboard } from '@react-aria/interactions';
+import { FocusStrategy } from '@react-types/shared';
```
cscheffauer, line 14: nitpick: this could be a `type` only import i think

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -1,18 +1,21 @@
...
+import ContentSeparator from '../ContentSeparator';
+import Popover from '../Popover';
+import { PopoverInstance, VariantType } from '../Popover/Popover.types';
```
cscheffauer, line 17: nitpick: this could be a `type` only import i think

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -21,98 +24,132 @@ const MenuTrigger: FC<Props> = (props: Props) => {
     style,
     closeOnSelect,
     children,
-    overlayRadius = DEFAULTS.OVERLAY_RADIUS,
+    isOpen,
```
cscheffauer, line 27: i think this default could also be a constant like DEFAULTS.ISOPEN

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -21,98 +24,132 @@ const MenuTrigger: FC<Props> = (props: Props) => {
...
-      isDismissable: state.isOpen,
-      isKeyboardDismissDisabled: false,
-    },
-    overlayRef
-  );
+  /**
+   * For some reason restoreFocus prop on <FocusScope> doesn't
+   * work. We focus back to trigger manually.
+   */
+  const handleFocusBackOnTrigger = () => {
```
cscheffauer, line 56: i would recommend using `useCallback` to improve performance - see here:
https://stackoverflow.com/a/36677798

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -21,98 +24,132 @@ const MenuTrigger: FC<Props> = (props: Props) => {
...
     onClose: state.close,
     closeOnSelect,
     ref: menuRef,
-    autoFocus: state.focusStrategy,
   };
 
-  // This should work out of the box, but for some reason it doesn't
-  // Added a temporary fix until I get the time to look deeper
+  /**
+   * Handle closeOnSelect from @react-aria manually
```
cscheffauer, line 68: for someone who doesn't know about what the closeOnSelect does, this comment isn't really useful. It would be better to actually explain what the code does imho

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -21,98 +24,132 @@ const MenuTrigger: FC<Props> = (props: Props) => {
...
+   */
+  useEffect(() => {
+    if (!state.isOpen && popoverInstance?.state.isVisible) {
+      popoverInstance.hide();
+      handleFocusBackOnTrigger();
+    }
+  }, [state.isOpen, popoverInstance]);
+
+  /**
+   * Handle isOpen from @react-aria manually
```
cscheffauer, line 78: same here - i would recommend mentioning that this is necessary to do cause we use our own custom popover :)

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -21,98 +24,132 @@ const MenuTrigger: FC<Props> = (props: Props) => {
...
+   * Add manual keyboard accessibility because our Popover component
+   * doesn't work well with @react-aria
+   */
   const { keyboardProps } = useKeyboard({
     onKeyDown: (event) => {
       if (event.key === 'Escape') {
         state.close();
       }
-      if (state.isOpen && event.key === 'Tab') {
+      if (state.isOpen && event.key === 'Tab' && menus.length === 1) {
```
cscheffauer, line 99: maybe a short comment about why the tab outside in multiple menus should not close the popover might be useful here

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -21,98 +24,132 @@ const MenuTrigger: FC<Props> = (props: Props) => {
...
+        {menus.map((menu: ReactElement, index) => {
+          return (
+            <MenuContext.Provider
+              value={
+                // when we have multiple menus inside the menu trigger, we want only the first menu
+                // to autoFocus on open. If we add autoFocus for all Menus, the last menu in the menu
+                // trigger will have the first focus, which is wrong
+                index === 0 ? { ...menuContext, autoFocus: 'first' as FocusStrategy } : menuContext
+              }
+              key={`{fragment-${index}}`}
```
cscheffauer, line 142: why fragment as a prefix? also why does the MenuContext need the key?

src/components/MenuTrigger/MenuTrigger.constants.ts
torgeadelin: Yes, you did that in the Popover as well.
image

src/components/MenuTrigger/MenuTrigger.stories.args.ts
torgeadelin: Changed.

src/components/MenuTrigger/MenuTrigger.stories.tsx
torgeadelin: No longer needed here.

The button used to be part of `children` in the previous implementation, and that was needed because I was providing children as an array in the stories/test.

src/components/MenuTrigger/MenuTrigger.tsx
torgeadelin: Good catch!

src/components/MenuTrigger/MenuTrigger.tsx
torgeadelin: Good catch!

src/components/MenuTrigger/MenuTrigger.tsx
torgeadelin: I don't think that should be a constant. DEFAULTS.IS_OPEN will never be true for this component. DEFAULTS are more for configuration of the component (mostly style). 

`isOpen` is controlling the state, so it's more a functional prop than customisation/style.

src/components/MenuTrigger/MenuTrigger.tsx
torgeadelin: This is explained in the stories.

src/components/MenuTrigger/MenuTrigger.tsx
torgeadelin: Explained in the stories.

src/components/MenuTrigger/MenuTrigger.tsx
torgeadelin: fixed.

src/components/MenuTrigger/MenuTrigger.tsx
cscheffauer: we should keep these docs here also up to date, cause if consumers are using the component just with Typescript and don't look at the storybook, they still want to have a good explanation about that prop imho :)

src/components/MenuTrigger/MenuTrigger.tsx
cscheffauer: same comment as above ;)

src/components/MenuTrigger/MenuTrigger.constants.ts
```diff
@@ -1,13 +1,24 @@
 const CLASS_PREFIX = 'md-menu-trigger';
+import { PLACEMENTS } from '../ModalArrow/ModalArrow.constants';
+import { COLORS } from '../ModalContainer/ModalContainer.constants';
+import type { PlacementType } from '../ModalArrow/ModalArrow.types';
+
+const MENU_TRIGGER_PLACEMENTS: Record<string, PlacementType> = Object.entries(PLACEMENTS)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  .filter(([_, value]) => value.startsWith('top') || value.startsWith('bottom'))
```
cscheffauer, line 8: this whole function looks a bit complicated - can u try to use the `pickBy` utility from lodash instead?
https://lodash.com/docs/4.17.15#pickBy

I think that would make it cleaner

cscheffauer (approved):

PR #229, opened 2022-04-05
fix(components/button-group): fix right border if MenuTrigger is a child

chburket:

Fix the left and right borders of ButtonGroup when a non-button is included in the children (e.g. MenuTrigger).

# Description

Added a MenuTrigger to the Audio Video Controls story as a test. Updated the .scss to select button last/first-of-type when adding begin/end border and border-radius.


torgeadelin (approved):
lgtm!

PR #235, opened 2022-04-13
feat(menu-trigger): bug fixes

torgeadelin:
# Description
- the internal state of the menuTrigger doesn't reflect correctly when popover was closed via click outside
- warnings about type of children in MenuTrigger, removed
- children type for MenuTrigger is `ReactElement[]` which is wrong, it's supposed to be `ReactNode`. 
- updated docs for MenuTrigger

### DIFF
--- a/src/components/MenuTrigger/MenuTrigger.types.ts
+++ b/src/components/MenuTrigger/MenuTrigger.types.ts
@@ -1,4 +1,4 @@
-import { CSSProperties, ReactElement } from 'react';
+import { CSSProperties, ReactElement, ReactNode } from 'react';
 import { MenuTriggerProps } from '@react-types/menu';
 import type { PopoverCommonStyleProps, PlacementType } from '../Popover/Popover.types';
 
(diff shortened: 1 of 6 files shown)


src/components/MenuTrigger/MenuTrigger.types.ts
```diff
@@ -1,4 +1,4 @@
-import { CSSProperties, ReactElement } from 'react';
+import { CSSProperties, ReactElement, ReactNode } from 'react';
```
InteractiveTimmy, line 1: Aren't these all types? Could we use `type` keyword?

src/components/Popover/Popover.tsx
```diff
@@ -70,6 +71,7 @@ const Popover: FC<Props> = (props: Props) => {
       trigger={trigger === 'mouseenter' ? 'mouseenter focusin' : trigger}
       interactive={interactive}
       appendTo="parent"
+      onClickOutside={onClickOutside}
```
InteractiveTimmy, line 74: Strange workaround for popover, i thought tippy supported this already?

InteractiveTimmy (changes requested):
Just a small item around type imports, otherwise, looks good!

src/components/Popover/Popover.tsx
torgeadelin: It's not a workaround, this is from tippy, we just didn't use it.

InteractiveTimmy (approved):
LGTM!

PR #238, opened 2022-04-22
feat(btn-hyper-and-input): bug fixes

torgeadelin:
# Description
- fixed input message bugs around text input
- fixed HTML element for button hyperlink

### DIFF
--- a/src/components/ButtonHyperlink/ButtonHyperlink.tsx
+++ b/src/components/ButtonHyperlink/ButtonHyperlink.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useRef, FC, RefObject } from 'react';
+import React, { forwardRef, useRef, RefObject } from 'react';
 import { useButton } from '@react-aria/button';
 import FocusRing from '../FocusRing';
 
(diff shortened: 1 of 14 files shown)


src/components/ButtonHyperlink/ButtonHyperlink.tsx
```diff
@@ -8,7 +8,7 @@ import './ButtonHyperlink.style.scss';
 import classnames from 'classnames';
 
 const ButtonHyperlink: FC<Props> = forwardRef(
-  (props: Props, providedRef: RefObject<HTMLButtonElement>) => {
+  (props: Props, providedRef: RefObject<HTMLAnchorElement>) => {
```
InteractiveTimmy, line 11: Was this changed to `HTMLAnchorElement` per react-aria? or was this overlooked during the original implementation?

src/components/ButtonHyperlink/ButtonHyperlink.tsx
```diff
@@ -21,19 +21,19 @@ const ButtonHyperlink: FC<Props> = forwardRef(
     delete mutatedProps.disabled;
     delete mutatedProps.className;
 
-    const { buttonProps } = useButton(mutatedProps, ref);
+    const { buttonProps } = useButton({ ...mutatedProps, elementType: 'a' }, ref);
 
     return (
       <FocusRing disabled={props.disabled}>
-        <button
+        <a
```
InteractiveTimmy, line 28: Glad we're changing this to meet ARIA for links!

src/components/InputMessage/InputMessage.unit.test.tsx.snap
```diff
@@ -29,13 +28,12 @@ exports[`InputMessage snapshot should match snapshot 1`] = `
             <svg
               className=""
               data-autoscale={false}
-              data-scale={20}
+              data-scale={16}
```
InteractiveTimmy, line 31: We should review "scale", as scale = 1 [normalized] should be used when it's at the default size. 

Not relevant to the scope of this PR though. See [Scalars](https://www.britannica.com/science/scalar) [derived from scale].

InteractiveTimmy (approved):
LGTM! thoughts/questions/comments attached.

src/components/ButtonHyperlink/ButtonHyperlink.tsx
torgeadelin: since the new base element is <a>, the ref needs the appropriate type.

src/components/InputMessage/InputMessage.unit.test.tsx.snap
torgeadelin: yeah, scale was definitely a bad name..

PR #244, opened 2022-05-16
feat(button-simple): text overflow

torgeadelin:
# Description

Added text overflow default for button simple.

### DIFF
--- a/src/components/ButtonSimple/ButtonSimple.style.scss
+++ b/src/components/ButtonSimple/ButtonSimple.style.scss
@@ -1,0 +1,8 @@
+.md-button-simple-wrapper {
+  > span,
+  .md-text-wrapper {
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+  }
+}
--- a/src/components/ButtonSimple/ButtonSimple.tsx
+++ b/src/components/ButtonSimple/ButtonSimple.tsx
@@ -6,6 +6,7 @@ import FocusRing from '../FocusRing';
 
 import { STYLE } from './ButtonSimple.constants';
 import type { Props } from './ButtonSimple.types';
+import './ButtonSimple.style.scss';
 
 /**
  * A simple button component without overhead styling. This is used as an injectable button component for other sibling components.
@@ -28,10 +29,10 @@ const ButtonSimple = forwardRef((props: Props, providedRef: RefObject<HTMLButton
         title={title}
         {...buttonProps}
         {...hoverProps}
-        // override of onKeyDown to ensure the standard html button behaviour (on enter => onClick) happens
+        // override of onKeyDown to ensure the standard html button behavior (on enter => onClick) happens
         onKeyDown={useNativeKeyDown ? props.onKeyDown : buttonProps.onKeyDown}
       >
-        {children}
+        {typeof children === 'string' ? <span>{children}</span> : children}
       </button>
     </FocusRing>
   );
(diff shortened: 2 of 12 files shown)


src/components/ButtonSimple/ButtonSimple.tsx
```diff
@@ -28,10 +29,10 @@ const ButtonSimple = forwardRef((props: Props, providedRef: RefObject<HTMLButton
         title={title}
         {...buttonProps}
         {...hoverProps}
-        // override of onKeyDown to ensure the standard html button behaviour (on enter => onClick) happens
+        // override of onKeyDown to ensure the standard html button behavior (on enter => onClick) happens
         onKeyDown={useNativeKeyDown ? props.onKeyDown : buttonProps.onKeyDown}
       >
-        {children}
+        {typeof children === 'string' ? <span>{children}</span> : children}
```
cscheffauer, line 35: i think putting a span in a button is semantically incorrect - see here:
https://stackoverflow.com/questions/12982269/is-it-semantically-incorrect-to-put-a-div-or-span-inside-of-a-button

torgeadelin: we need something tho, we can't overflow the text directly on the container because it has flex.

src/components/ButtonSimple/ButtonSimple.style.scss
```diff
@@ -0,0 +1,8 @@
+.md-button-simple-wrapper {
+  > span,
+  .md-text-wrapper {
+    text-overflow: ellipsis;
```
cscheffauer, line 4: i think this can be simply put on the button html element to work - without the need of a span.

I have checked in Storybook to set the width of the button element to 100px - it didn't put on the elipsis:
image

src/components/ButtonSimple/ButtonSimple.tsx
cscheffauer: which container do you mean?

cscheffauer:
i think it would be good to create a story for the overflow scenario - such that we see the ellipsis and can play around with the text and width of a button.

src/components/ButtonSimple/ButtonSimple.style.scss
torgeadelin: we can't add the overflow to the container / button HTML element because it has flex on it :(
it needs to be a child flex

src/components/ButtonSimple/ButtonSimple.tsx
Coread: That SO article says span is fine but div is not

src/components/ButtonSimple/ButtonSimple.style.scss
cscheffauer: for me the button has display: inline-block in the browser:
image

src/components/ButtonSimple/ButtonSimple.tsx
cscheffauer: ah good catch - was reading it to quickly :)

src/components/ButtonSimple/ButtonSimple.style.scss
torgeadelin: I know it doesn't work on the ButtonSimple, but I feel like that's fine because it's used internally only.

cscheffauer: thanks - that makes sense now for me!

PR #249, opened 2022-06-01
Fix svgr dynamic loading

Coread:
# Description

Update the module for both cjs and esm to es2020. This is necessary for SVGs to be dynamically loaded.
The reaction animations work the same way so I also fixed those as well.

A way to test this is working on the storybook is to pop out the Icon story page as a separate window. Open the network tab and refresh the page. There should be a separate request for the icon chunk (it will be a js bundle)

This will break downstream projects that require less than es2020 support. They should push momentum through babel if that support is required.

# Links

### DIFF
--- a/src/hooks/useDynamicJSONImport.ts
+++ b/src/hooks/useDynamicJSONImport.ts
@@ -38,8 +38,9 @@ function useDynamicJSONImport(
       try {
         setError(undefined);
         setError(undefined);
-        ImportedJSONRef.current =
-          await require(`@momentum-ui/animations/lottie/reactions/${name}.json`);
+        ImportedJSONRef.current = (
+          await import(`@momentum-ui/animations/lottie/reactions/${name}.json?lottie`)
+        ).default;
         if (isMounted()) {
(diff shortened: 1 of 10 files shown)


torgeadelin:
Since this is a breaking change, we should use the appropriate commits to make sure the automatic versioning will create the correct version.

src/hooks/useDynamicJSONImport.ts
```diff
@@ -38,8 +38,9 @@ function useDynamicJSONImport(
       try {
         setError(undefined);
         setError(undefined);
-        ImportedJSONRef.current =
-          await require(`@momentum-ui/animations/lottie/reactions/${name}.json`);
+        ImportedJSONRef.current = (
+          await import(`@momentum-ui/animations/lottie/reactions/${name}.json?lottie`)
```
torgeadelin, line 42: Why is the lottie necessary here? 
Lottie are for animations afaik.

torgeadelin: Oh, I seee -- this is for reactions... the name of the hook is a bit misleading.

PR #252, opened 2022-06-02
fix(components/card): status color positioning

InteractiveTimmy:
# Description

The scope of the changes in this pull request is to fix the missing status-color visual elements from the `<Card />` component for `<MeetingContainerCard />` consumption.

## Screenshots

Screen Shot 2022-06-02 at 9 42 48 AM
Screen Shot 2022-06-02 at 9 42 55 AM
Screen Shot 2022-06-02 at 9 43 04 AM
Screen Shot 2022-06-02 at 9 43 22 AM
Screen Shot 2022-06-02 at 9 43 31 AM

# Links

[SPARK-329844](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-329844)


src/components/Card/Card.style.scss
```diff
@@ -12,7 +12,8 @@
   cursor: pointer;
   display: flex;
   margin: 0.25rem;
-  overflow: hidden;
+  overflow: hidden !important;
```
cscheffauer, line 15: why is the !important necessary here?

chburket (approved):
LGTM.  Make sure to check the MeetingContainer > Common story with a narrow browser to see if the long title is still handled properly.

src/components/Card/CardStatus/CardStatus.style.scss
```diff
@@ -1,6 +1,7 @@
 .md-card-status-wrapper {
   --local-background-color: transparent;
 
+  align-self: stretch;
```
cscheffauer, line 4: instead of `align-self: stretch` i think `height: 100%` might fix the issue here better (both work, but I believe setting the height might make more sense here).

src/components/Card/Card.style.scss
InteractiveTimmy: Yeah, default behavior from the original momentum `scss` files is overwriting this style.

src/components/Card/CardStatus/CardStatus.style.scss
InteractiveTimmy: this does not work, height:100% doesn't detect the dynamic height values of the `<Card />` parent. Feel free to try this if you have this branch pulled.

InteractiveTimmy:
> LGTM. Make sure to check the MeetingContainer > Common story with a narrow browser to see if the long title is still handled properly.

Screen Shot 2022-06-02 at 10 19 11 AM

Looks like it is, lmk if you see any issues.

PR #255, opened 2022-06-08
feat(checkbox): add checkbox component

nataliadelmar:
# Description

Add the Checkbox component as defined in the Component Library.
image

# Links

[*Links to relevent resources.*](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=40300%3A80462)

### DIFF
--- a/src/components/Checkbox/Checkbox.stories.args.ts
+++ b/src/components/Checkbox/Checkbox.stories.args.ts
@@ -3,18 +3,14 @@ import { commonStyles } from '../../storybook/helper.stories.argtypes';
 import { CHECKBOX_CONSTANTS as CONSTANTS } from '.';
 
 const checkboxArgTypes = {
-  // NOTE: Below is an example. See [Storybook argTypes documentation]{@link https://storybook.js.org/docs/react/api/argtypes}.
   children: {
-    description:
-      'Provides the child nodes for this element. If a label exists, it should appear here.', // NOTE: Description of this prop.
-    control: { type: 'text' }, // NOTE: Control type for this prop.
+    description: 'The label that the checkbox should have.',
+    control: { type: 'text' },
     table: {
       type: {
-        summary: 'ReactNode', // NOTE: Explicit type of this prop.
+        summary: 'string',
       },
-      defaultValue: {
-        summary: 'undefined', // NOTE: Default value for this prop.
-      },
+      defaultValue: CONSTANTS.DEFAULTS.LABEL,
     },
   },
   isSelected: {
@@ -45,16 +41,6 @@ const checkboxArgTypes = {
         summary: 'boolean',
       },
       defaultValue: CONSTANTS.DEFAULTS.IS_INDETERMINATE,
-    },
-  },
-  label: {
-    description: 'The label that the checkbox should have.',
-    control: { type: 'text' },
-    table: {
-      type: {
-        summary: 'string',
-      },
-      defaultValue: CONSTANTS.DEFAULTS.LABEL,
     },
   },
 };
--- a/src/components/Checkbox/Checkbox.stories.args.ts
+++ b/src/components/Checkbox/Checkbox.stories.args.ts
@@ -3,7 +3,7 @@ import { commonStyles } from '../../storybook/helper.stories.argtypes';
 import { CHECKBOX_CONSTANTS as CONSTANTS } from '.';
 
 const checkboxArgTypes = {
-  children: {
+  label: {
     description: 'The label that the checkbox should have.',
     control: { type: 'text' },
     table: {
--- a/src/components/Checkbox/Checkbox.stories.args.ts
+++ b/src/components/Checkbox/Checkbox.stories.args.ts
@@ -1,0 +1,67 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+
+import { CHECKBOX_CONSTANTS as CONSTANTS } from '.';
+
+const checkboxArgTypes = {
+  // NOTE: Below is an example. See [Storybook argTypes documentation]{@link https://storybook.js.org/docs/react/api/argtypes}.
+  children: {
+    description:
+      'Provides the child nodes for this element. If a label exists, it should appear here.', // NOTE: Description of this prop.
+    control: { type: 'text' }, // NOTE: Control type for this prop.
+    table: {
+      type: {
+        summary: 'ReactNode', // NOTE: Explicit type of this prop.
+      },
+      defaultValue: {
+        summary: 'undefined', // NOTE: Default value for this prop.
+      },
+    },
+  },
+  isSelected: {
+    description: 'Whether the checkbox is selected or not',
+    control: { type: 'boolean' },
+    table: {
+      type: {
+        summary: 'boolean',
+      },
+      defaultValue: CONSTANTS.DEFAULTS.IS_SELECTED,
+    },
+  },
+  disabled: {
+    description: 'Whether the checkbox is disabled or not',
+    control: { type: 'boolean' },
+    table: {
+      type: {
+        summary: 'boolean',
+      },
+      defaultValue: CONSTANTS.DEFAULTS.DISABLED,
+    },
+  },
+  isIndeterminate: {
+    description: 'Whether the checkbox should appear in an indeterminate state or not',
+    control: { type: 'boolean' },
+    table: {
+      type: {
+        summary: 'boolean',
+      },
+      defaultValue: CONSTANTS.DEFAULTS.IS_INDETERMINATE,
+    },
+  },
+  label: {
+    description: 'The label that the checkbox should have.',
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: CONSTANTS.DEFAULTS.LABEL,
+    },
+  },
+};
+
+export { checkboxArgTypes };
+
+export default {
+  ...commonStyles,
+  ...checkboxArgTypes,
+};
--- a/src/components/Checkbox/Checkbox.stories.args.ts
+++ b/src/components/Checkbox/Checkbox.stories.args.ts
@@ -23,14 +23,14 @@ const checkboxArgTypes = {
       defaultValue: CONSTANTS.DEFAULTS.IS_SELECTED,
     },
   },
-  disabled: {
+  isDisabled: {
     description: 'Whether the checkbox is disabled or not',
     control: { type: 'boolean' },
     table: {
       type: {
         summary: 'boolean',
       },
-      defaultValue: CONSTANTS.DEFAULTS.DISABLED,
+      defaultValue: CONSTANTS.DEFAULTS.IS_DISABLED,
     },
   },
   isIndeterminate: {
--- a/src/components/Checkbox/Checkbox.stories.tsx
+++ b/src/components/Checkbox/Checkbox.stories.tsx
@@ -32,17 +32,14 @@ States.parameters = {
   variants: [
     {
       label: 'Checked',
-      children: 'Checked',
       isSelected: true,
     },
     {
       label: 'Unchecked',
-      children: 'Checked',
       isSelected: false,
     },
     {
       label: 'Indeterminate',
-      children: 'Checked',
       isIndeterminate: true,
     },
   ],
--- a/src/components/Checkbox/Checkbox.stories.tsx
+++ b/src/components/Checkbox/Checkbox.stories.tsx
@@ -1,0 +1,71 @@
+import { MultiTemplate, Template } from '../../storybook/helper.stories.templates';
+import { DocumentationPage } from '../../storybook/helper.stories.docs';
+import StyleDocs from '../../storybook/docs.stories.style.mdx';
+
+import Checkbox, { CheckboxProps } from './';
+import argTypes from './Checkbox.stories.args';
+import Documentation from './Checkbox.stories.docs.mdx';
+import { DEFAULTS } from './Checkbox.constants';
+
+export default {
+  title: 'Momentum UI/Checkbox',
+  component: Checkbox,
+  parameters: {
+    expanded: true,
+    docs: {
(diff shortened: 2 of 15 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,59 @@
...
+        name={indeterminate ? 'minus' : 'check'}
+        weight="bold"
+        scale={12}
+      />
+    </div>
+  );
+
+  return (
+    <FocusRing>
+      <label className={classnames(STYLE.wrapper, className)} data-disabled={isDisabled}>
```
maxinteger, line 46: I think you should add `data-selected` and `data-indeterminate` attributes as well

src/components/Checkbox/Checkbox.unit.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+      const style = { color: 'pink' };
+      const styleString = 'color: pink;';
+
+      const element = mount(<Checkbox style={style} />)
+        .find(Checkbox)
+        .getDOMNode();
+
+      expect(element.getAttribute('style')).toBe(styleString);
+    });
```
maxinteger, line 95: Do we need these tests? They are checking the same thing as the snapshot tests

src/components/Checkbox/Checkbox.unit.test.tsx.snap
```diff
@@ -0,0 +1,140 @@
...
+  </label>
+</Checkbox>
+`;
+
+exports[`<Checkbox /> snapshot should match snapshot with className 1`] = `
+<Checkbox
+  className="example-class"
+>
+  <label
+    style={
```
maxinteger, line 40: something wrong with the snapshots 

for example here the `{display: "block"}` should not be added to the label and the className form the props is missing

it is might be because of the icon loading asynchronously the svg file
try to use `mountAndWait` instead of `mount` in the snapshot tests

maxinteger (changes requested):
Added few comments for the tests and a suggestion for additional data- attributes

src/components/Checkbox/Checkbox.stories.args.ts
```diff
@@ -0,0 +1,67 @@
...
+  children: {
+    description:
+      'Provides the child nodes for this element. If a label exists, it should appear here.', // NOTE: Description of this prop.
+    control: { type: 'text' }, // NOTE: Control type for this prop.
+    table: {
+      type: {
+        summary: 'ReactNode', // NOTE: Explicit type of this prop.
+      },
+      defaultValue: {
+        summary: 'undefined', // NOTE: Default value for this prop.
```
torgeadelin, line 16: Remove comments

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <VisuallyHidden>
+          <input {...inputProps} {...focusProps} ref={ref} />
+        </VisuallyHidden>
+        {(selected || indeterminate) && <div className={STYLE.selected}>{icon}</div>}
+        {!selected && !indeterminate && <div className={STYLE.notSelected} />}
+        {label && <Text className={STYLE.label}>{label}</Text>}
```
torgeadelin, line 52: Remove the `<Text>` component, the `label` tag only allows inline blocks inside. It should be sufficient to add the label styling on the label itself (i think you already have the font family, just add the correct size and text color in there)

torgeadelin: Not sure if they need to be tokenised?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  return (
+    <FocusRing>
+      <label className={classnames(STYLE.wrapper, className)} data-disabled={isDisabled}>
+        <VisuallyHidden>
+          <input {...inputProps} {...focusProps} ref={ref} />
+        </VisuallyHidden>
+        {(selected || indeterminate) && <div className={STYLE.selected}>{icon}</div>}
+        {!selected && !indeterminate && <div className={STYLE.notSelected} />}
```
torgeadelin, line 51: Could be simplified in

```js
{(selected || indeterminate) 
  ? <div className={STYLE.selected}>{icon}</div>
  : <div className={STYLE.notSelected} />
}
```
}

```

File: src/components/Checkbox/Checkbox.style.scss [hunk 7]
```diff
@@ -0,0 +1,83 @@
+.md-checkbox-wrapper {
```
torgeadelin, line 1: We also need `width: fit-content`, otherwise, the label will grow on the whole line and you'd be able to toggle the checkbox when clicking far away from the label. See pic: 
image

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,59 @@
...
+        }
+        name={indeterminate ? 'minus' : 'check'}
+        weight="bold"
+        scale={12}
+      />
+    </div>
+  );
+
+  return (
+    <FocusRing>
```
torgeadelin, line 45: FocusRing doesn't really work here, because the focus spec is a bit complicated, and it is only around the rectangle, and since we are not using the native html element for checkbox (which does that), we need to manually implement the focus using ` useFocusRing()` from Aria. There's an example on how to do it [here](https://react-spectrum.adobe.com/react-aria/useCheckbox.html)

src/components/Checkbox/Checkbox.types.ts
```diff
@@ -0,0 +1,43 @@
+import { CSSProperties, ReactNode } from 'react';
+
+export interface Props {
+  /**
+   * Child components of this Checkbox.
+   */
+  children?: ReactNode;
```
torgeadelin, line 7: Remove

torgeadelin (changes requested):
Looking good, a few bugs and things that need changing tho.
Also, needs testing :P

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+      disabled: false,
+      isIndeterminate: true,
+      label: 'Not selected + Not disabled + Indeterminate',
+    },
+    {
+      isSelected: false,
+      disabled: true,
+      isIndeterminate: true,
+      label: 'Not selected + Disabled + Indeterminate',
+    },
```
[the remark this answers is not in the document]
torgeadelin: storybook is now broken
image

src/components/Checkbox/Checkbox.stories.args.ts
```diff
@@ -0,0 +1,67 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+
+import { CHECKBOX_CONSTANTS as CONSTANTS } from '.';
+
+const checkboxArgTypes = {
+  // NOTE: Below is an example. See [Storybook argTypes documentation]{@link https://storybook.js.org/docs/react/api/argtypes}.
+  children: {
```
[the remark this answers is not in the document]
torgeadelin: i still see it?

src/components/Checkbox/Checkbox.style.scss
```diff
@@ -0,0 +1,87 @@
...
+  display: flex;
+  align-items: center;
+  width: fit-content;
+  font-size: 1rem;
```
torgeadelin, line 5: no need if you're using Text

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,76 @@
...
+import Icon from '../Icon';
+
+/**
+ * The Checkbox component.
+ */
+const Checkbox: FC<Props> = (props: Props) => {
+  const { children, className, isSelected, disabled, isIndeterminate, ...rest } = props;
+
+  const state = useToggleState(props);
+  const ref = React.useRef();
```
torgeadelin, line 23: Might be worth adding forward ref, see `ListItemBase` on how we do it.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const selected = isSelected || state.isSelected || DEFAULTS.IS_SELECTED;
+  const indeterminate = isIndeterminate || DEFAULTS.IS_INDETERMINATE;
+  const isDisabled = disabled || DEFAULTS.DISABLED;
+
+  const { pressProps } = usePress({
+    preventFocusOnPress: true, // we handle it ourselves
+    isDisabled,
+    ...rest,
+  });
```
torgeadelin, line 35: Why did you add this?

torgeadelin: `inputProps` returned by `useCheckbox` should already handle this

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const icon = (
+    <div>
```
torgeadelin, line 38: no need for the extra div

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useCheckbox } from 'react-aria';
+import { useToggleState } from '@react-stately/toggle';
+
+import Icon from '../Icon';
+
+/**
+ * The Checkbox component.
+ */
+const Checkbox: FC<Props> = (props: Props) => {
+  const { children, className, isSelected, disabled, isIndeterminate, ...rest } = props;
```
torgeadelin, line 20: you should grab `label` prop instead of children for the label. Toggle component does that, so we should follow the same pattern maybe. (#259)

src/components/Checkbox/Checkbox.types.ts
```diff
@@ -0,0 +1,43 @@
+import { ToggleButtonProps } from '@react-types/button';
+import { CSSProperties, ReactNode } from 'react';
+
+export interface Props extends ToggleButtonProps {
+  /**
+   * Child components of this Checkbox. If free text, it becomes the label of the Checkbox.
+   */
+  children?: ReactNode;
+  /**
```
torgeadelin, line 9: add newline

src/components/Checkbox/Checkbox.unit.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      expect.assertions(1);
+
+      const style = { color: 'pink' };
+
+      const container = await mountAndWait(<Checkbox style={style} />);
+
+      expect(container).toMatchSnapshot();
+    });
+
+    it('should match snapshot with not selected checkbox', async () => {
```
torgeadelin, line 46: Add one test with selected as well

src/components/Checkbox/Checkbox.unit.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      const disabled = true;
+
+      const wrapper = await mountAndWait(<Checkbox disabled={disabled} />);
+      const element = wrapper.find(Checkbox).getDOMNode();
+
+      expect(element.getAttribute('data-disabled')).toBe(disabled.toString());
+    });
+  });
+
+  describe('actions', () => {
```
torgeadelin, line 127: Add test for space key interaction, see example here #259

src/components/ThemeProvider/ThemeProvider.style.scss
```diff
@@ -133,6 +133,9 @@
   --md-globals-border-clear: 0rem solid rgba(0, 0, 0, 0);
   --md-globals-focus-ring-box-shadow: 0 0 0 0.25rem rgba(100, 180, 250, 0.3),
     0 0 0 0.125rem var(--theme-outline-theme-normal);
+  --md-globals-focus-ring-box-shadow-accent: 0 0 0 0.0625rem
```
torgeadelin, line 136: Why did you add this? is it because it's different in figma than the one we currently have?
Has this changed everywhere maybe? Can you check? If yes, we just need to update the existing one on line 134

torgeadelin (changes requested):
Few more small changes, but looking great!

torgeadelin (approved):
Looks really great!

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,75 @@
...
+const Checkbox = (props: Props, providedRef: RefObject<HTMLInputElement>) => {
+  const { className, isDisabled, label, isIndeterminate, isSelected, id, style } = props;
+
+  const state = useToggleState(props);
+  const internalRef = useRef<HTMLInputElement>();
+  const ref = providedRef || internalRef;
+  const { inputProps } = useCheckbox(props, state, ref);
+  const { isFocusVisible, focusProps } = useFocusRing();
+
+  const selected = isSelected || state.isSelected || DEFAULTS.IS_SELECTED;
```
maxinteger, line 27: I do not think we need any default value here

you can remove these lines

You can replace everywhere the `isSelected` with `state.isSelected` because `useToggleState` gets the all the props and it is using the `isSelected`

maxinteger (changes requested):
I added few minor comments

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { RefObject, useRef, forwardRef } from 'react';
+import classnames from 'classnames';
+
+import { VisuallyHidden } from '@react-aria/visually-hidden';
+import { useFocusRing } from '@react-aria/focus';
+
+import { Props } from './Checkbox.types';
+import { STYLE, DEFAULTS } from './Checkbox.constants';
+import './Checkbox.style.scss';
+import { useCheckbox } from 'react-aria';
```
maxinteger, line 10: The build is failing because you import the whole `react-aria` instead of `@react-aria/checkbox`

Not sure what is the root cause of this error tho. We might have incompatible versions of the react-aria in the project. But we can investigate that in another story

PR #256, opened 2022-06-09
fix: tsc-alias build & TextToast

cscheffauer:
# Description

- Fixed build issues by using tsc-alias package, which resolves the absolute paths as intended
- Small fix in TextToast to only pass className to parent div

# Links

https://www.npmjs.com/package/tsc-alias


torgeadelin (approved):
Lgtm! Thanks for working this out!!!

I'll make a separate pr to remove the webpack plugin i added for storybook.

PR #259, opened 2022-06-10
feat(toggle): add toggle component

maxinteger:
# Description

Add the Toggle component as defined in the Component Library.

image

# Links

- [Web design](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=40297%3A97282)
- [Windows design](https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp/Components---Windows?node-id=46518%3A104139#214985832)


src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  variants: [
+    { isSelected: true, isDisabled: false, label: 'Selected' },
+    { isSelected: true, isDisabled: true, label: 'Selected + Disabled' },
+    { isSelected: false, isDisabled: false, label: 'Not selected' },
+    { isSelected: false, isDisabled: true, label: 'Not selected + Disabled' },
+  ],
+};
+
+// NOTE: Export stories here. The first export should be `Example`, and the last export should be `Common`.
+export { Example, Common };
```
maxinteger, line 56: Anybody knows why Storybook shows the "Example" story under both exported stories?

If I change the order then I can see the "Common" in both case but not the Examples

cscheffauer: will double check

cscheffauer:
@maxinteger the build is failing - pls check why :)

src/components/Toggle/Toggle.style.scss
```diff
@@ -0,0 +1,119 @@
+@use "sass:math";
+
+.md-toggle-wrapper {
+  $switch-width: 2.75rem;
+  $switch-height: math.div($switch-width, 2);
```
cscheffauer, line 5: is this a requirement? that the height is half of the width?

src/components/Toggle/Toggle.style.scss
```diff
@@ -0,0 +1,119 @@
...
+  $thumb-size: 0.635rem;
+  $thumb-margin: math.div($switch-height - $thumb-size, 2);
+  $transition-time: 350ms;
```
cscheffauer, line 8: where is this coming from? was this defined in figma somewhere? :) i haven't been working with css animations in Momentum yet, so I'm curious how that has been defined.

src/components/Toggle/Toggle.unit.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    it('should have provided selected when selected is provided', () => {
+      expect.assertions(2);
+
+      const component = mount(<Toggle isSelected={true} aria-label={'toggle'} />);
+      const wrapper = component.find(Toggle).getDOMNode();
+      const input = component.find('input').getDOMNode();
+
+      expect(wrapper.getAttribute('data-selected')).toBe('true');
+      expect(input.hasAttribute('checked')).toBe(true);
```
cscheffauer, line 140: since we have jest-dom from testing-library extended, you could also simply do:
```typescript
expect(input).toBeChecked();
```

src/components/Toggle/Toggle.unit.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+      component.simulate('change', { target: { checked: true } });
+      expect(mockCallback).toBeCalledTimes(1);
+      expect(mockCallback).toBeCalledWith(true);
+
+      component.simulate('change', { target: { checked: false } });
+      expect(mockCallback).toBeCalledTimes(2);
+      expect(mockCallback).toBeCalledWith(false);
+    });
+  });
```
cscheffauer, line 162: in terms of confidence I think keyboard events would be also nice to be tested.

yarn.lock
```diff
@@ -2578,6 +2578,17 @@
     "@react-types/shared" "^3.11.2"
     clsx "^1.1.1"
 
+"@react-aria/focus@^3.6.0":
```
cscheffauer, line 2581: I believe these changes here were not meant to be made (I guess that happened when running yarn in the repository). Could be probably removed (but I don't know exactly how we deal with this normally).

src/components/Toggle/Toggle.stories.tsx
cscheffauer: i'm not experiencing that - there might be a storybook caching issue, which could be resolved by running:
`yarn storybook:start --no-manager-cache`

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { STYLE } from './Toggle.constants';
+import { Props } from './Toggle.types';
+import './Toggle.style.scss';
+
+/**
+ * The Toggle component. Also known as Switch.
+ */
+const Toggle: FC<Props> = (props: Props) => {
+  const { id, className, style, isDisabled, label } = props;
```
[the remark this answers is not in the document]
cscheffauer: ah sorry - found it - there is a onChange type

src/components/Toggle/Toggle.style.scss
maxinteger: based on the figma design, yes
the width of the toggle is double of the height I can not see any reason why it should change in any circumstances

src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  parameters: {
+    expanded: true,
+    docs: {
+      page: DocumentationPage(Documentation, StyleDocs),
+    },
+  },
+};
+
+const InteractiveExample: FC<ToggleProps> = (props: ToggleProps) => {
+  const [isSelected, setIsSelected] = useState(props.isSelected);
```
cscheffauer, line 22: this is awesome, but it's not really visible in the story. It would be great to fire a "onChange" action on the actions tab in storybook. To make that work, pls check other stories and how they do it:
image

src/components/Toggle/Toggle.style.scss
maxinteger: I got it from the Legacy toggle button, but probably I should check with the UX team

yarn.lock
maxinteger: Yes, I used wronger version of react-switch

src/components/Toggle/Toggle.tsx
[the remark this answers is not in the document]
maxinteger: the `onChange` is inherited from the `AriaToggleProps` and it is managed by the `useToggleState` hook
You can see an example in the last unit test

src/components/Toggle/Toggle.style.scss
cscheffauer: I would check that with @nataliadelmar if thats fine or not :)
I don't mind having good looking animations, but it depends on the spec really.

torgeadelin: it's ok to leave it, better than nothing i guess?

src/components/Toggle/Toggle.constants.ts
```diff
@@ -0,0 +1,19 @@
...
+  IS_SELECTED: false,
+  LABEL: undefined,
+};
+
+const STYLE = {
+  wrapper: `${CLASS_PREFIX}-wrapper`,
+  toggle: `${CLASS_PREFIX}-switch`,
+  label: `${CLASS_PREFIX}-label`,
+  disabled: `${CLASS_PREFIX}--disabled`,
+  focused: `${CLASS_PREFIX}--focused`,
```
torgeadelin, line 14: i don't think we should use double dashes. you're probably borrowing the concept from [BEM](https://en.bem.info/methodology/) but we don't really have that as a convention in this repo.

src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  component: Toggle,
+  parameters: {
+    expanded: true,
+    docs: {
+      page: DocumentationPage(Documentation, StyleDocs),
+    },
+  },
+};
+
+// NOTE: Primary story. This renders a single component with all external props.
```
torgeadelin, line 20: Remove coomment

src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+Example.argTypes = { ...argTypes };
+
+Example.args = {
+  defaultSelected: TOGGLE_CONSTANTS.DEFAULTS.IS_SELECTED,
+  isDisabled: TOGGLE_CONSTANTS.DEFAULTS.IS_DISABLED,
+  label: 'Example text',
+};
+
+// NOTE: Common variants story. This renders multiple variants of a single component.
```
torgeadelin, line 31: Remove

src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+Common.parameters = {
+  variants: [
+    { defaultSelected: true, isDisabled: false, label: 'Selected' },
+    { defaultSelected: true, isDisabled: true, label: 'Selected + Disabled' },
+    { defaultSelected: false, isDisabled: false, label: 'Not selected' },
+    { defaultSelected: false, isDisabled: true, label: 'Not selected + Disabled' },
+  ],
+};
+
+// NOTE: Export stories here. The first export should be `Example`, and the last export should be `Common`.
```
torgeadelin, line 50: Remove

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const ref = React.useRef();
```
torgeadelin, line 18: Might be worth adding ForwardRef as well. Look at `ListItemBase` component as an example on how we handle this.

src/components/Toggle/Toggle.types.ts
```diff
@@ -0,0 +1,29 @@
...
+   */
+  id?: string;
+
+  /**
+   * Custom style for overriding this component's CSS.
+   */
+  style?: CSSProperties;
+
+  /**
+   * Boolean to describe when toggle is on/off
```
torgeadelin, line 21: Might be worth mentioning that this is a prop that will control the component externally. Ie, if it's set to true, and you click on it, state wouldn't change.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const children = label && <span className={STYLE.label}>{label}</span>;
```
torgeadelin, line 20: In terms of the label, the spec doesn't really show any so not sure we need this.
I've asked a question, please continue the conversation with design team on this here: https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp?node-id=46518:104139#214980864

torgeadelin (changes requested):
A few comments, but looking really great!

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const internalRef = React.useRef<HTMLInputElement>();
+  const ref = providedRef || internalRef;
+
+  const state = useToggleState(props);
+  const { inputProps } = useSwitch(props, state, ref);
+  const { isFocusVisible, focusProps } = useFocusRing();
+
+  return (
+    <label
```
maxinteger, line 26: Note: this component does not have `label` prop (see [Figma discussion](https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp/Components---Windows?node-id=46518%3A104139#214985832)), but we need the `<label>` tag to handle accessibility. The only issue is the user of the component might wrap Toggle into another `<label>`. 
As far as I know label can not contain another label, but I tried in Chrome and it did not cause any issue.

If we want to remove the label then we have to remove the input as well, and implement a completely custom input component.

src/components/Toggle/Toggle.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+Example.argTypes = { ...argTypes };
+
+Example.args = {
+  defaultSelected: TOGGLE_CONSTANTS.DEFAULTS.DEFAULT_SELECTION,
+  isDisabled: TOGGLE_CONSTANTS.DEFAULTS.IS_DISABLED,
+  'aria-label': 'Example text',
+};
+
+const componentStateToProps = (state) => {
```
cscheffauer, line 30: why is this necessary?

src/components/Toggle/Toggle.unit.test.tsx
cscheffauer: I think what would be nice here as well is a focus. So that u do `userEvent.focus` and then you do a `expect(input).toBeFocused()`.

Also I would recommend a test with disabled state, where u set the Toggle to disabled and then u try to do the focus again and it shouldn't work. Also pressing space key or clicking shouldn't work on disabled.
With that we have more confidence that it works just as the user would use it :)

torgeadelin (approved):
Looking good!

src/components/Toggle/Toggle.style.scss
```diff
@@ -0,0 +1,100 @@
...
+
+    &:before {
+      background-color: var(--togglemomentumstyle-untoggled-thumb-tint-disabled);
+    }
+
+    &:checked {
+      background-color: var(--togglemomentumstyle-toggled-background-disabled);
+
+      &:before {
+        background-color: var(--togglemomentumstyle-toggled-thumb-tint-disabled);
```
cscheffauer, line 95: is that the name from the Tokens library? togglemomentumstyle ? that is sounds odd to me
@torgeadelin @nataliadelmar

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * The Toggle component. Also known as Switch.
+ */
+const Toggle = (props: Props, providedRef: RefObject<HTMLInputElement>) => {
+  const { id, className, style, isDisabled } = props;
+
+  const internalRef = React.useRef<HTMLInputElement>();
+  const ref = providedRef || internalRef;
+
+  const state = useToggleState(props);
```
cscheffauer, line 19: `isSelected` could be destructured here already

cscheffauer (approved):

src/components/Toggle/Toggle.style.scss
maxinteger: I created a PR to fix the variable names
https://github.com/momentum-design/tokens/pull/230

PR #260, opened 2022-06-13
feat(chips/modifiers): Add modifiers to chips 

breardon2011:
# Description
Add modifiers to the chips to support the search and multiline modifier as described in the figma and their corresponding supported sizes. 

# Links

[figma](https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp/Components---Windows?node-id=44750%3A103435)


src/components/Chip/Chip.style.scss
```diff
@@ -13,20 +14,67 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
     color: var(--theme-text-error-normal);
 
     &[data-outline='true'] {
-      border: 1px solid;
-      border-radius: 1rem;
+      border: 0.063rem solid;
       border-color: var(--theme-outline-cancel-normal);
```
torgeadelin, line 18: this should come from the component tokens, not theme

src/components/Chip/Chip.style.scss
```diff
@@ -13,20 +14,67 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
...
     }
   }
 
   &[data-error='false'] {
     &[data-outline='true'] {
-      border: 1px solid;
+      border: 0.063rem solid;
       border-radius: 1rem;
       border-color: var(--theme-outline-primary-normal);
```
torgeadelin, line 26: this should come from the component tokens, not theme

src/components/Chip/Chip.style.scss
```diff
@@ -51,6 +99,27 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
 
   > .md-chip-center-section:first-child:last-child {
     padding: 0em 0.625rem;
+
+    &[data-search='true'] {
+      font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+      padding: 0rem 0.438rem;
```
torgeadelin, line 105: suspicious value

src/components/Chip/Chip.style.scss
```diff
@@ -51,6 +99,27 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
 
   > .md-chip-center-section:first-child:last-child {
     padding: 0em 0.625rem;
+
+    &[data-search='true'] {
+      font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
```
torgeadelin, line 104: wrong font, we're not using that in the web library. we already set the font in the ThemeProvider so no need to write it here

src/components/Chip/Chip.style.scss
```diff
@@ -51,6 +99,27 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
...
+      padding: 0rem 0.438rem;
+
+      &[data-size='24'] {
+        padding: 0.063rem 0.5rem;
```
torgeadelin, line 108: suspicious value

src/components/Chip/Chip.style.scss
```diff
@@ -51,6 +99,27 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
...
+      }
+
+      &[data-size='28'] {
+        padding: 0.125rem 0.5rem;
+      }
+    }
+
+    &[data-multiline='true'] {
+      padding: 0rem 0.406rem;
```
torgeadelin, line 117: suspicious value

src/components/Chip/Chip.style.scss
```diff
@@ -51,6 +99,27 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
...
+      font-size: 0.875rem;
+      font-weight: 600;
+      font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
```
torgeadelin, line 120: remove

src/components/Chip/Chip.tsx
```diff
@@ -21,27 +21,43 @@ const Chip: FC<Props> = (props: Props) => {
...
+    size = DEFAULTS.SIZE,
+    chipColor = DEFAULTS.COLOR,
     error = DEFAULTS.ERROR,
+    multiline = DEFAULTS.MULTILINE,
+    search = DEFAULTS.SEARCH,
   } = props;
 
   const leftContent = avatar ? avatar : leftIcon;
   const changedText = text ? text : children;
+  const textStyle = search && size === 24 ? { fontSize: '.875rem', fontWeight: 600 } : undefined;
```
torgeadelin, line 33: instead of this, you can use the already defined `types` we have on the `<Text/>` component.
https://momentum-design.github.io/momentum-react-v2/?path=/story/momentum-ui-text--types

if you hover on the typography in Figma, you'll see what type you need
image

torgeadelin: so you can do something like `<Text type={search ? 'subheader-secondary': 'another-type'}/>`. You'd have more logic here tho, because it also depends on the size.

After doing this you can remove all type related css properties from the css file. The `<Text/>` component will have this based on the `type` prop.

src/components/Chip/Chip.tsx
```diff
@@ -21,27 +21,43 @@ const Chip: FC<Props> = (props: Props) => {
...
       {leftContent && (
         <div data-avatar={!!avatar} className={STYLE.leftSection}>
           {leftContent}
         </div>
       )}
-      <Text className={STYLE.centerSection}>{changedText}</Text>
+      <Text
+        style={textStyle}
+        data-search={search}
+        data-multiline={multiline}
```
torgeadelin, line 56: this doesn't really work, tried to use it in storybook and i see this.
image

src/components/Chip/Chip.style.scss
```diff
@@ -51,6 +99,27 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
 
   > .md-chip-center-section:first-child:last-child {
     padding: 0em 0.625rem;
```
torgeadelin, line 101: rem

src/components/Chip/Chip.style.scss
```diff
@@ -13,20 +14,67 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
...
-      border-radius: 1rem;
+      border: 0.063rem solid;
       border-color: var(--theme-outline-cancel-normal);
     }
   }
 
   &[data-error='false'] {
     &[data-outline='true'] {
-      border: 1px solid;
+      border: 0.063rem solid;
```
torgeadelin, line 24: suspicious value

src/components/Chip/Chip.tsx
breardon2011: Yeah I struggled with what was the cleanest, most easy to understand way to manage the variations in play, I dont mind doing it this way

src/components/Chip/Chip.style.scss
breardon2011: Hmm I guess I'm curious why they specify it in the figma, but will remove for sure

src/components/Chip/Chip.style.scss
breardon2011: ah ok, I was sort of unclear on how the tokens are used here

src/components/Chip/Chip.style.scss
```diff
@@ -13,20 +14,67 @@ button.md-button-pill-wrapper.md-chip-wrapper.md-button-simple-wrapper {
     color: var(--theme-text-error-normal);
 
     &[data-outline='true'] {
-      border: 1px solid;
-      border-radius: 1rem;
+      border: 0.063rem solid;
```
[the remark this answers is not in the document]
breardon2011: A note on the variations, the multicolor chips is put on hold because the UI colors did not exist, so only the modifiers will be submitted, the only color tokens I created were for multiline etc. There is a separate blocked jira for the multicolor chips

src/components/Chip/Chip.style.scss
breardon2011: Made a tokens pr to add https://github.com/momentum-design/tokens/pull/220

src/components/Chip/Chip.stories.tsx
```diff
@@ -84,6 +84,22 @@ Common.parameters = {
...
+      children: 'With:',
+      size: 24,
+      search: true,
+    },
+    {
+      children: 'L2',
+      multiline: true,
+      size: 16,
+      chipColor: 'cyan',
+    },
```
InteractiveTimmy, line 102: As Catalin mentioned, this should be done as its own sub-story: `ChipColors`. This can easily be done by iterating over the `CHIP_CONSTANTS.MULTILINE_COLORS` [or similar const] in the variants with a map:

```js
ChipColors.parameters = {
  variants: CHIP_CONSTANTS.MULTILINE_COLORS.map((color) => ({children: `chipColor = ${color}`, chipColor: color})),
}
```

src/components/Chip/Chip.unit.test.tsx
```diff
@@ -218,5 +236,26 @@ describe('<Chip />', () => {
...
+
+      expect(element.getAttribute('data-multiline')).toBe(String(multiline));
+    });
+    it('should have color when provided', () => {
+      const element = mount(<Chip chipColor={color} />)
+        .find(Chip)
+        .getDOMNode();
+
+      expect(element.getAttribute('data-modifier-color')).toBe(color);
+    });
```
InteractiveTimmy, line 259: It appears a `size` test may be missing.

InteractiveTimmy (changes requested):
Seems to be a missing test for `size`, needs the variants to be inclusive of the full list of new colors.

src/components/Chip/Chip.style.scss
torgeadelin: it's in figma because you're most likely looking at the [Windows Library](https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp/Components---Windows?node-id=0%3A1) and not the [Web](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=0%3A1) BUT, the windows does tend to be more up to date. I usually check both 🤷🏻‍♂️

But as a note, you should never have to write any typography (99.9% of the time) because these come from the Text component or are there by default (from the ThemeProvider). 

In terms of size and weight, yes, you might need to adjust those, bug again, if you use `Text` component the font, weight and size will automatically be set to the correct values and you'll only need to figure out the `type` property, which usually is in Figma if you hover over the `Ag` property in the Typography section on the right hand side. 
image

*to see the whole name hover over the value of `Ag` and it'll show. In the example the correct type is `body-primary`.

src/components/Chip/Chip.style.scss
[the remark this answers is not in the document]
torgeadelin: thanks for explaining this!

torgeadelin (approved):
Looking great! thanks for addressing all the comments :D

InteractiveTimmy (approved):
LGTM! Thanks for addressing the comments!

PR #264, opened 2022-06-19
fix/tsconfig/Update tsconfig.cjs.json

jonniespratley:


Coread:
Can you explain why this change is required?

jonniespratley:
Because testing on down stream project with jest will not run

PR #265, opened 2022-06-21
feat(notification-toast): add notification toast component

nataliadelmar:
# Description

Adding Notification Toast component to the library as seen in Figma.

image

https://www.figma.com/file/NaNrfXjygZtRgMfHAFHjsp/Components---Windows?node-id=36552%3A55990

# Links

### DIFF
--- a/src/components/ToastNotification/ToastNotification.stories.tsx
+++ b/src/components/ToastNotification/ToastNotification.stories.tsx
@@ -24,60 +24,39 @@ export default {
 
 const Example = Template<ToastNotificationProps>(ToastNotification).bind({});
 
-Example.argTypes = { ...argTypes };
+Example.argTypes = { ...argTypes, onClose: { action: 'closed' } };
 
 Example.args = {
   text: 'Example text',
-  onClose: () => {
-    alert('hello');
-  },
 };
 
 const LeadingVisual = MultiTemplate<ToastNotificationProps>(ToastNotification).bind({});
 
-LeadingVisual.argTypes = { ...argTypes };
+LeadingVisual.argTypes = { ...argTypes, onClose: { action: 'closed' } };
 
 LeadingVisual.parameters = {
   variants: [
     {
-      onClose: () => {
-        alert('hello');
-      },
       text: 'Example text',
       leadingVisual: <Avatar initials="CW" size={32} />,
     },
     {
-      onClose: () => {
-        alert('hello');
-      },
       text: 'Very long text that can occupy two, three or even more lines. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation.',
       leadingVisual: <Avatar initials="CW" size={32} />,
     },
     {
-      onClose: () => {
-        alert('hello');
-      },
       text: 'Example text',
       leadingVisual: <Icon name="help-circle" scale={24} weight="bold" />,
     },
     {
-      onClose: () => {
-        alert('hello');
-      },
       text: 'Very long text that can occupy two, three or even more lines. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation.',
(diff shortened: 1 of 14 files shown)


src/components/ToastNotification/ToastNotification.style.scss
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  .md-toast-notification-button-group {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    gap: 0.5rem;
+    padding-top: 0.25rem;
+    padding-bottom: 0.25rem;
+    padding-right: 0.1rem;
```
Coread, line 41: question: I always ask when I see non-pixel sized rem values. Is the right padding really 0.1rem here? That is 1.6px

Coread (approved):
Looks good, just check on the padding value and that unused import before merge

src/components/ToastNotification/ToastNotification.unit.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+      expect.assertions(1);
+
+      const style = { color: 'pink' };
+      const styleString = 'color: pink;';
+
+      const wrapper = await mountAndWait(<ToastNotification style={style} />);
+      const element = wrapper.find(ToastNotification).getDOMNode();
+
+      expect(element.getAttribute('style')).toBe(styleString);
+    });
```
torgeadelin, line 137: you forgot to add tests for leadingVisual, buttons and onClose

torgeadelin (approved):
Approving so in case I'm not here, it doesn't block you but i'd like for you to add the additional tests.

src/components/ToastNotification/ToastNotification.style.scss
```diff
@@ -0,0 +1,42 @@
...
+      padding-left: 0.75rem;
+    }
+  }
+
+  .md-toast-notification-button-group {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    gap: 0.5rem;
+    padding-top: 0.25rem;
```
maxinteger, line 39: based on the figma it should be 0.75rem between the text and the button group but the gap is only 0.5rem at the moment

src/components/ToastNotification/ToastNotification.style.scss
```diff
@@ -0,0 +1,42 @@
+.md-toast-notification-wrapper {
```
maxinteger, line 1: it should be 1rem padding on all sides, but it is only 0.5rem. Even if I add the leading visual, text or the button group paddings it is still just 0.75rem

src/components/ToastNotification/ToastNotification.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+  },
+};
+
+const Example = Template<ToastNotificationProps>(ToastNotification).bind({});
+
+Example.argTypes = { ...argTypes };
+
+Example.args = {
+  text: 'Example text',
+  onClose: () => {
```
maxinteger, line 31: instead of this you can add the `onClose` to the `argTypes` like this:
```ts
export default {
   ...commonStyles,
   ...toastNotificationArgTypes, 
   onClose: { action: 'closed' },
}
```

this way you can utilise the "Actions" tab in Sotrybook and you can remove all the `onClose` handlers from this file

src/components/ToastNotification/ToastNotification.types.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ButtonCircleProps } from '../ButtonCircle';
+
+export type SupportedLeadingVisuals = AvatarProps | IconProps | BadgeProps;
+export type SupportedButtonGroup = ButtonPillProps | ButtonCircleProps;
+
+export interface Props {
+  /**
+   * The text that appears inside the Notification toast.
+   */
+  text?: string;
```
maxinteger, line 17: text prop should be required

maxinteger (changes requested):
Few minor spacing problems and some chages for handling component actions in the story

src/components/ToastNotification/ToastNotification.style.scss
maxinteger: The left and bottom padding of the wrapper is good now, but the top and right padding are still small

maxinteger (changes requested):
minor style problems

PR #266, opened 2022-06-29
fix: added hideOnEsc on Popover

cscheffauer:
# Description

- Added a plugin to Popover to fix the hideOnEsc (hideOnEsc is turned on by default)
- Added unit tests for it
- Set hideOnEsc to false in MenuTrigger component (Escape is handled separately there).

Side Note: noticed Bug in MenuTrigger, which could be fixed with this new hideOnEsc Popover functionality:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-342208

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-342004


maxinteger (changes requested):
Added minor comment about typing, otherwise looks great

maxinteger (approved):
Nice job.

PR #273, opened 2022-07-13
feat(popover): add close btn and focusBackOnTrigger

torgeadelin:
# Description
- Added `closeButtonPlacement` to configure the popover to have a close button (`top-left' | 'top-right' | 'none'`)
- Added `focusBackOnTrigger` to configure the popover to focus back on trigger on close
- Added `closeButtonProps` to support the `closeButtonPlacement` and give flexibility to add accessibility props on the close button (aria label, title, translation etc)

# Screenshots
image
image

### DIFF
--- a/src/components/Popover/Popover.constants.ts
+++ b/src/components/Popover/Popover.constants.ts
@@ -9,6 +9,12 @@ const BOUNDARIES = {
   PARENT: 'scrollParent',
 };
 
+const CLOSE_BUTTON_PLACEMENTS: Record<string, string> = {
+  TOP_LEFT: 'top-left',
+  TOP_RIGHT: 'top-right',
+  NONE: 'none',
+};
+
 const DEFAULTS = {
   VARIANT: 'small',
   TRIGGER: 'click',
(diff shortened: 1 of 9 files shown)


src/components/Popover/Popover.constants.ts
```diff
@@ -9,6 +9,12 @@ const BOUNDARIES = {
   PARENT: 'scrollParent',
 };
 
+const CLOSE_BUTTON_PLACEMENTS: Record<string, string> = {
```
maxinteger, line 12: Please remove `Record<string, string>` because it define a much wider type than what we need.
TS can infer the best type for `CLOSE_BUTTON_PLACEMENTS`

src/components/Popover/Popover.stories.args.ts
```diff
@@ -104,6 +104,31 @@ const popoverArgTypes = {
       },
     },
   },
+  closeButtonPlacement: {
+    description: `Placement of the close button relative to the container of the popover`,
+    options: [...Object.values(CLOSE_BUTTON_PLACEMENTS as Record<string, string>)],
```
maxinteger, line 109: `as Record<string, string>` is not necessary here

src/components/Popover/Popover.style.scss
```diff
@@ -5,3 +5,17 @@
 .tippy-content {
   padding: 0;
 }
+
+.md-popover-close-button {
+  position: absolute;
+  z-index: 1000;
```
maxinteger, line 11: why it is 1000? Do we have any system to define which z-index can be used in different situation?

maxinteger (changes requested):

src/components/Popover/Popover.stories.args.ts
torgeadelin: removed

src/components/Popover/Popover.constants.ts
torgeadelin: removed

maxinteger (approved):

PR #276, opened 2022-07-19
feat(icons): add warning-badge as exception

torgeadelin:
# Description
Add `warning-badge` as small icon exception.
There's a small issue with the icon itself, it was designed at 16x16 instead of 14x14 like the rest of the exception icons.
[I've left a comment in the figma](https://www.figma.com/file/SXK8Gb5tMlN9xiG2cC4OBu?node-id=9091:12274#230785156) -- this will probabbly automatically get fixed when they fix it on their end.


chburket (commented):
lgtm!

PR #277, opened 2022-07-20
feat/add-radio-button

Jex-y:
# Description

Added RadioGroup component

# Links

[figma spec](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=40968%3A67045)


src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <span {...labelProps}>{label || DEFAULTS.GROUP_LABEL}</span>
+      {description && <div className={STYLE.description}>{description}</div>}
+      <RadioContext.Provider value={state}>{childElement}</RadioContext.Provider>
+    </div>
+  );
+};
+
+// Where should passed style, id, and className go?
+
+const Radio: FC<Props> = (props: Props) => {
```
cscheffauer, line 70: I would recommend moving this out into a separate file - that makes it more clear and abstracted.

src/components/RadioGroup/RadioGroup.stories.args.ts
```diff
@@ -0,0 +1,94 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+import { DEFAULTS } from './RadioGroup.constants';
+
+const radioGroupArgTypes = {
+  // NOTE: Below is an example. See [Storybook argTypes documentation]{@link https://storybook.js.org/docs/react/api/argtypes}.
```
cscheffauer, line 5: remove this line

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const selected = state.selectedValue === props.value;
+  const disabled = isDisabled || DEFAULTS.OPTION_DISABLED;
+
+  const icon = <Icon className={STYLE.icon} name="shape-circle" weight="filled" scale={8} />;
+
+  const radio = (
+    <div
+      className={classnames(
+        {
+          [STYLE.selected]: selected,
```
cscheffauer, line 85: replace with data-attribute

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+const Radio: FC<Props> = (props: Props) => {
+  const { className, children, isDisabled, id, style } = props;
+  const state = useContext(RadioContext);
+  const ref = useRef(null);
+  const { inputProps } = useRadio(props, state, ref);
+  const { isFocusVisible, focusProps } = useFocusRing();
+  const selected = state.selectedValue === props.value;
+  const disabled = isDisabled || DEFAULTS.OPTION_DISABLED;
+
+  const icon = <Icon className={STYLE.icon} name="shape-circle" weight="filled" scale={8} />;
```
cscheffauer, line 79: since its only used once, i would recommend using inline

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const radio = (
```
cscheffauer, line 81: since its only used once, i would recommend using inline

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <div
+      className={classnames(
+        {
+          [STYLE.selected]: selected,
+          [STYLE.notSelected]: !selected,
+          [STYLE.focus]: isFocusVisible,
```
cscheffauer, line 87: can u pls check if the FocusRing component also works here instead?

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+// Where should passed style, id, and className go?
+
+const Radio: FC<Props> = (props: Props) => {
+  const { className, children, isDisabled, id, style } = props;
+  const state = useContext(RadioContext);
+  const ref = useRef(null);
+  const { inputProps } = useRadio(props, state, ref);
+  const { isFocusVisible, focusProps } = useFocusRing();
+  const selected = state.selectedValue === props.value;
```
cscheffauer, line 76: instead of doing it here i would recommend passing the selected as a prop to the radio component

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+    childElement = (
+      <>
+        {options.map((option: string | Props) => {
+          if (typeof option === 'string') {
+            return (
+              <Radio key={option} value={option} isDisabled={disabled}>
+                {option}
+              </Radio>
+            );
+          } else if (React.isValidElement(option)) {
```
cscheffauer, line 42: i think this is not necessary

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,113 @@
...
+            return option;
+          } else {
+            const value = option.value;
```
cscheffauer, line 45: this can be done inline

src/components/RadioGroup/RadioGroup.tsx
Jex-y: If I do this, I would still need to take in the state to pass to `useRadio` so then there could be a case where the isSelected value does not match. Should I try and mock the state for `useRadio`, reimplement the features that it provides or keep the context and this bit the same?

src/components/RadioGroup/RadioGroup.tsx
Jex-y: I don't think the `FocusRing` component will work as the element being focused (the hidden input) is different from the one being styled.

src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,28 @@
...
+   * Custom class for overriding this component's CSS.
+   */
+  className?: string;
+
+  /**
+   * Custom style for overriding this component's CSS.
+   */
+  style?: CSSProperties;
+
+  options?: Array<Props | string | ReactNode>;
```
cscheffauer, line 15: i would recommend adding docs here for all props

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const RadioContext = React.createContext(null);
+
+const RadioGroup: FC<GroupProps> = (props: GroupProps) => {
+  const { className, label, id, style, options, isDisabled, setValue, orientation, description } =
+    props;
+
+  let onChange: (value: string) => void = props.onChange;
+
+  if (setValue) {
+    onChange = props.onChange
```
cscheffauer, line 24: that is really weird - i think we shouldn't allow onChange and setValue to be passed in.

src/components/RadioGroup/Radio.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const selected = state.selectedValue === props.value;
+
+  return (
+    <label
+      data-disabled={disabled}
+      className={classnames(STYLE.wrapper, className)}
+      style={style}
+      id={id}
+    >
+      <VisuallyHidden>
```
cscheffauer, line 32: i think we don't need visuallyhidden - we can simply use `<input type="radio" />` and style it how we want.
The icon in line 38 is also not necessary, cause the styling could be done through css.

Jex-y: It appears to be difficult to style radio buttons and they are generally done by hiding the original component. I misread the react-aria docs and it is done there as well: https://react-spectrum.adobe.com/react-aria/useRadioGroup.html#styling

src/components/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+    docs: {
+      page: DocumentationPage(Documentation, StyleDocs),
+    },
+  },
+  args: {
+    setValue: action('setValue'),
+  },
+};
+
+// NOTE: Primary story. This renders a single component with all external props.
```
torgeadelin, line 24: remove all comments

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,115 @@
...
+          border-style: none;
+        }
+      }
+    }
+  }
+}
+
+.md-radio-group-description {
+  font-size: 0.9rem;
+  line-height: 1.2rem;
```
torgeadelin, line 112: You should use `<Text/>` component with the corresponding `type` instead and remove the css related to typography.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const RadioContext = React.createContext(null);
+
+const RadioGroup: FC<GroupProps> = (props: GroupProps) => {
+  const { className, label, id, style, options, isDisabled, setValue, orientation, description } =
+    props;
+
+  const onChange = (value: string) => {
+    setValue?.(value);
+  };
```
torgeadelin, line 23: Any reason you had to do this?

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const state = useRadioGroupState({ ...props, onChange });
+  const { radioGroupProps, labelProps } = useRadioGroup(props, state);
+
+  const direction = orientation || DEFAULTS.GROUP_ORIENTATION;
+  const disabled = isDisabled || DEFAULTS.GROUP_DISABLED;
+
+  if (direction === 'horizontal' && description) {
+    console.warn('Using orientation=horizontal and description at the same time is not supported');
+  }
```
torgeadelin, line 33: Nice!

torgeadelin: But where did you see spec for horizontal layout for radio?

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <div
+      {...radioGroupProps}
+      className={classnames(className, STYLE.group)}
+      id={id}
+      style={style}
+      data-direction={direction}
+    >
+      <span {...labelProps}>{label || DEFAULTS.GROUP_LABEL}</span>
```
torgeadelin, line 43: Please assign defaults when you deconstruct the props and not in the return.
Applies for all defaults.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,63 @@
...
+      {description && <div className={STYLE.description}>{description}</div>}
```
torgeadelin, line 44: We don't seem to support description in Figma? Where did you get this from?

src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,48 @@
+import { CSSProperties } from 'react';
+import { AriaRadioGroupProps, AriaRadioProps } from '@react-types/radio';
+
+export interface GroupProps extends Omit<AriaRadioGroupProps, 'children' | 'onChange'> {
```
torgeadelin, line 4: Why omitting `onChange`?

src/components/RadioGroup/RadioGroup.unit.test.tsx
```diff
@@ -0,0 +1,781 @@
...
+            },
+          ]}
+          isDisabled={true}
+        />
+      );
+
+      expect(asFragment()).toMatchSnapshot();
+    });
+
+    // Start
```
torgeadelin, line 277: what's the comment for?

src/components/RadioGroup/RadioGroup.unit.test.tsx
```diff
@@ -0,0 +1,781 @@
...
+          ]}
+          defaultValue="option1"
+        />
+      );
+
+      const option1 = getByText('Option 1').querySelector('.md-radio-button');
+      const option2 = getByText('Option 2').querySelector('.md-radio-button');
+
+      expect(option1).toHaveAttribute('data-selected', 'true');
+      expect(option2).toHaveAttribute('data-selected', 'false');
```
torgeadelin, line 540: We should probably check the functional property, not the `data-selected`.
I think there's ` expect(input).toHaveValue(...etc);` assertion available. You can keep the data-selected too actually, it gives more coverage.

this applies to all tests in the `actions` suite.

torgeadelin (changes requested):
Small changes.
Huge praise for the excellent testing coverage! Probably the most well tested component in the library!! Great job!

src/components/RadioGroup/RadioGroup.tsx
Jex-y: It is not in the figma but the current component is being used in Cantia with a description. I could contact the UX team and see if we could get a spec but for now I have made it similar to the current radioGroups usage: 
image

src/components/RadioGroup/RadioGroup.tsx
Jex-y: In the case that setValue is not set this would do nothing rather than error and talking to Christoph he said that it would be preferred to something like
```js
if (setValue) {
    setValue(value)
}
```

src/components/RadioGroup/RadioGroup.tsx
Jex-y: Its not in the figma spec but is supported by react-aria radio group and it was a few lines. It can be removed if it would be better to keep the component closer to the spec

src/components/RadioGroup/RadioGroup.types.ts
Jex-y: Having spoken to @cscheffauer it would be better to only pass a setter for the value to make it more obviously a controlled component.

src/components/RadioGroup/RadioGroup.tsx
Jex-y: Currently reworking that menu with the new components like this: 
image

src/components/RadioGroup/RadioGroup.unit.test.tsx
Jex-y: @torgeadelin Would it be ok to just check the data attribute since the styles associated with it are tested in the snapshots? Im having some trouble getting the applied css in the test.

cscheffauer: you could actually do `expect(option1).toBeChecked();`

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,70 @@
...
+          options.map((option: string | Props) => {
+            if (typeof option === 'string') {
+              return (
+                <Radio key={option} value={option} isDisabled={isDisabled}>
+                  {option}
+                </Radio>
+              );
+            } else {
+              return <Radio key={option.value} isDisabled={isDisabled} {...option} />;
+            }
```
maxinteger, line 63: the `children` was excluded from the Radio's Props
This issue is visible in the test snapshots as well, in the second test case the labels are empty

```suggestion
           if (typeof option === 'string') {
              return <Radio key={option} value={option} isDisabled={isDisabled} label={option} />;
            } else {
              return <Radio key={option.value} isDisabled={isDisabled} {...option} />;
            }
```

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { STYLE, DEFAULTS } from './RadioGroup.constants';
+import { GroupProps, Props } from './RadioGroup.types';
+import './RadioGroup.style.scss';
+import Radio from './Radio';
+import Text, { TEXT_CONSTANTS } from '../Text';
+
+/**
+ * The RadioGroup component.
+ */
+
```
maxinteger, line 15: minor: please do not add extra empty line after the doc comment

src/components/RadioGroup/RadioGroup.stories.docs.mdx
```diff
@@ -0,0 +1,29 @@
...
+| onKeyUp          | `(e: KeyboardEvent) => void`   | Handler that is called when a key is released.                                                                         |
+| id               | `string`                       | The element's unique identifier. See MDN.                                                                              |
+| aria-label       | `string`                       | Defines a string value that labels the current element.                                                                |
+| aria-labelledby  | `string`                       | Identifies the element (or elements) that labels the current element.                                                  |
+| aria-describedby | `string`                       | Identifies the element (or elements) that describes the object.                                                        |
+| aria-details     | `string`                       | Identifies the element (or elements) that provide a detailed, extended description for the object                      |
```
maxinteger, line 27: this table is unnecessary, sotrybook autogenerate it for you, check the Docs tab and scroll to the bottom

maxinteger (changes requested):

src/components/RadioGroup/RadioGroup.stories.docs.mdx
Jex-y: This table is for the the props that are passed for each option in an array rather than the whole group and I think storybook only does it for the top level component.

Jex-y: Actually I may be wrong, just found this in the documentation https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components, will have a deeper look tomorrow.

src/components/RadioGroup/Radio.tsx
Jex-y: I have now been able to do this by using `appearance: none`

src/components/RadioGroup/RadioGroup.types.ts
Jex-y: Removed in [#518f267](https://github.com/momentum-design/momentum-react-v2/pull/277/commits/518f26794676d8f20c1eb6d6cf5d3ae7ddd23118)

src/components/RadioGroup/RadioGroup.tsx
Jex-y: Removed in [#518f26](https://github.com/momentum-design/momentum-react-v2/pull/277/commits/518f26794676d8f20c1eb6d6cf5d3ae7ddd23118)

src/components/RadioGroup/RadioGroup.tsx
torgeadelin: I mean more about the whole wrapping of the callback. I don't understand why we need `setValue` we should probably be using onChange

src/components/RadioGroup/RadioGroup.tsx
torgeadelin: nice!

src/components/RadioGroup/RadioGroup.types.ts
torgeadelin: we shouldn't decide this. this has to act exactly like the native HTML radio button and should also have the same API.
We're just styling it here, nothing more.

src/components/RadioGroup/RadioGroup.stories.docs.mdx
torgeadelin: We don't always include all the props that come from react-aria.. I'm not sure how I feel about having the table or not. I guess it helps for now but if they make any changes it will be hell to keep this updated.

We had an attempt to generate the storybook props table based on typescript and doc comments, not sure why we didn't follow with that.

I guess for now I'm not bothered if this stays here.

src/components/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -3,13 +3,15 @@ import { DocumentationPage } from '../../storybook/helper.stories.docs';
...
 import RadioGroup, { RadioGroupProps } from './';
+import Radio from './Radio';
 import argTypes from './RadioGroup.stories.args';
 import Documentation from './RadioGroup.stories.docs.mdx';
 import { action } from '@storybook/addon-actions';
 
 export default {
   title: 'Momentum UI/RadioGroup',
   component: RadioGroup,
+  subcomponents: { Radio },
```
torgeadelin, line 14: nice!

src/components/RadioGroup/RadioGroup.types.ts
torgeadelin: Because the `onChange` from the `useRadioGroup` has this type `(value: T) => void` we can always just pass the setter function of a useState hook directly without having to deal with `e.target.value`.

Also changing this to `setvalue` breaks the pattern of HTMLInputs -- they all have an onChange.

So please use onChange and remove the setValue prop.

torgeadelin (approved):
Looking great!

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+      width: 1rem;
+      border-radius: 0.5rem;
+      outline: none !important;
+      display: flex;
+      justify-content: center;
+      align-items: center;
+    }
+
+    // Could maybe use :has() when there is more support
+    // See https://caniuse.com/css-has
```
maxinteger, line 29: If we want to use `:has` then it should be part of a bigger refactoring
please remove this comment

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+    margin-left: 0.5rem;
+    margin-right: 0.5rem;
+
+    > .md-radio-button {
+      appearance: none;
+      -webkit-appearance: none;
+      height: 1rem;
+      width: 1rem;
+      border-radius: 0.5rem;
+      outline: none !important;
```
maxinteger, line 22: `!important` is not needed here

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+      display: flex;
+      justify-content: center;
+      align-items: center;
```
maxinteger, line 25: the `input` does not have any child component, so the these properties are not needed

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+
+    &[data-disabled='false'] {
+      > .md-radio-label {
+        color: var(--radio-unchecked-text);
+      }
+
+      > .md-radio-button {
+        &:checked {
+          border: 5px solid var(--radio-checked-background);
+          border-radius: 50%;
```
maxinteger, line 39: because of the consistency, please use `border-radius: 0.5rem`
BTW it is unnecessary because you already set the `border-radius` in line 21

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      &:hover,
+      &.hover {
+        color: var(--radio-checked-text-hovered);
+
+        > .md-radio-button {
+          &:checked {
+            border: 5px solid var(--radio-checked-background-hovered);
+            border-radius: 50%;
```
maxinteger, line 61: remove it please

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      &:active,
+      &.active {
+        color: var(--radio-unchecked-text-pressed);
+
+        > .md-radio-button {
+          &:checked {
+            border: 5px solid var(--radio-checked-background-pressed);
+            border-radius: 50%;
```
maxinteger, line 77: remove it please

src/components/RadioGroup/RadioGroup.style.scss
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    &[data-disabled='true'],
+    &.disable {
+      color: var(--radio-unchecked-text-disabled);
+
+      > .md-radio-button {
+        &:checked {
+          border: 5px solid var(--radio-checked-background-disabled);
+          border-radius: 50%;
```
maxinteger, line 94: remove it please

src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,42 @@
...
+  options?: Array<Props | string>;
+
+  /**
+   * The description for the RadioGroup.
+   */
+
+  description?: string;
+}
+
+export interface Props extends Omit<AriaRadioProps, 'children'> {
```
maxinteger, line 27: minor: for better readability please rename this interface to `RadioProps`

src/components/RadioGroup/RadioGroup.types.ts
```diff
@@ -0,0 +1,42 @@
+import { CSSProperties } from 'react';
+import { AriaRadioGroupProps, AriaRadioProps } from '@react-types/radio';
+
+export interface GroupProps extends Omit<AriaRadioGroupProps, 'children'> {
```
maxinteger, line 4: I see you removed the orientation prop and the radios rendered horizontally
In forms radios rendered in vertically so you have more space for labels and more readable.
Is UX team happy with this default rendering?

maxinteger (changes requested):
Only minor issues, otherwise looks great

src/components/RadioGroup/RadioGroup.types.ts
Jex-y: That was a mistake, thanks for spotting.

src/components/RadioGroup/index.ts
```diff
@@ -0,0 +1,9 @@
+import RadioGroup from './RadioGroup';
+import * as CONSTANTS from './RadioGroup.constants';
+import { RadioProps as _RadioProps, GroupProps } from './RadioGroup.types';
+
+export { CONSTANTS as RADIO_GROUP_CONSTANTS };
+
+export type RadioProps = _RadioProps;
+export type RadioGroupProps = GroupProps;
```
maxinteger, line 8: You can simplify the export here:
```suggestion
export * as RADIO_GROUP_CONSTANTS from './RadioGroup.constants';
export * from './RadioGroup.types';
```

maxinteger (changes requested):
Just one more request :)

maxinteger (approved):
Nice job

PR #288, opened 2022-08-11
Revert "Remove legacy checkbox"

robstax:
Reverts momentum-design/momentum-react-v2#281

PR 281 removed InputHelper, but Input was actually still using it


chburket (approved):
Looks good

PR #289, opened 2022-08-12
feat(menu-item): fix toggle accessibility issue

torgeadelin:
# Description
Legacy Menu Items were preventDefaulting events when a toggle is provided as children causing keyboard accessibility on the toggle to break. This PR makes sure the events for menu-item are only handled when the event target is the menu-item itself.


shnaaz:
Missing unittest but I am not sure how keen you are to write unittests for legacy component.
Tested this with Cantina and toggle works but after toggle the Menu loses focus and next tab or arrow down closes the menu.

shnaaz (approved):
Tested with Cantina and works fine, focus comes back after Shift + Tab

PR #290, opened 2022-08-19
feat(components): added notification-system component

cscheffauer:
# Description

- Added NotificationSystem component
- Added utils and exposed them as part of the compound component pattern
- Added unit tests using RTL for better confidence
- Added sample stories for the usage of the NotificationSystem
- NOTE: NotificationTemplate sub-component is not exposed and only used for storybook. Every consumer can create their notification templates on their own and inject them.

# Links

[Jira](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-350110)
[Tech breakdown](https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/SPARK-350110+-+Notification+System)

### DIFF
--- a/src/components/NotificationSystem/NotificationSystem.types.ts
+++ b/src/components/NotificationSystem/NotificationSystem.types.ts
@@ -77,6 +77,13 @@ export interface Props {
    * Default: `top-right`
    */
   position?: PositionType;
+
+  /**
+   * Allows overriding the zIndex of the NotificationSystem
+   *
+   * Default: 10000
+   */
+  zIndex?: number;
   /**
    * Child components of this NotificationSystem.
    */
--- a/src/components/NotificationSystem/NotificationSystem.types.ts
+++ b/src/components/NotificationSystem/NotificationSystem.types.ts
@@ -1,7 +1,6 @@
 import { CSSProperties, ReactNode } from 'react';
 import { ATTENTION, POSITION } from './NotificationSystem.constants';
-import { dismiss, notify, update } from './NotificationSystem.utils';
-import NotificationTemplate from './NotificationTemplate';
+import { dismiss, isActive, notify, update } from './NotificationSystem.utils';
 import type { ToastOptions, UpdateOptions } from 'react-toastify';
 
 export type AttentionType = typeof ATTENTION[keyof typeof ATTENTION];
@@ -43,10 +42,15 @@ export interface CompoundProps {
    */
   update: typeof update;
   dismiss: typeof dismiss;
+  isActive: typeof isActive;
   /**
    * The ATTENTION enum to give consumers access to the different attention types
    */
   ATTENTION: typeof ATTENTION;
+  /**
+   * The POSITION enum to give consumers access to the different positioning options
+   */
+  POSITION: typeof POSITION;
 }
 
 export interface Props {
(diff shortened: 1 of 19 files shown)


src/components/NotificationSystem/NotificationSystem.types.ts
```diff
@@ -0,0 +1,99 @@
+import { CSSProperties, ReactNode } from 'react';
+import { ATTENTION, POSITION } from './NotificationSystem.constants';
+import { dismiss, isActive, notify, update } from './NotificationSystem.utils';
+import type { ToastOptions, UpdateOptions } from 'react-toastify';
+
+export type AttentionType = typeof ATTENTION[keyof typeof ATTENTION];
+export type PositionType = typeof POSITION[keyof typeof POSITION];
```
torgeadelin, line 7: This seems a bit backwards, we usually make the type and instance the constants based on it.
I'm not too bothered.

src/components/NotificationSystem/NotificationSystem.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  const { notificationSystemId, attention, ...updateOptions } = options;
+  toast.update(toastId, {
+    ...updateOptions,
+    containerId: getContainerID(notificationSystemId, attention || ATTENTION.LOW),
+  });
+};
+
+export const dismiss = (toastId: Id): void => {
+  toast.dismiss(toastId);
+};
```
torgeadelin, line 36: can we add docs to these methods?

src/components/NotificationSystem/NotificationSystem.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  const defaultAutoClose =
+    options?.attention === ATTENTION.MEDIUM ? false : DEFAULTS.AUTOCLOSE_TIMEOUT;
+  return options?.autoClose > 0 ? options.autoClose : defaultAutoClose;
+};
+
+export const notify = (content: ToastContent, options?: NotifyOptionsType): Id =>
+  toast(content, {
+    toastId: options?.toastId,
+    autoClose: calculateAutoClose(options),
+    containerId: getContainerID(options?.notificationSystemId, options?.attention || ATTENTION.LOW),
```
torgeadelin, line 22: i'm assuming this will default to whatever the `container_id_suffix` if no containerId is provided

torgeadelin (changes requested):
Looks really good!
Just a few small questions, changes

src/components/NotificationSystem/NotificationSystem.types.ts
cscheffauer: i think you can't create JS constants of typescript types ;) 
so you would need to write it twice - with this approach the Typescript type is always in sync with the constant, which makes it more robust against changes :)

src/components/NotificationSystem/NotificationSystem.utils.ts
cscheffauer: the getContainerID function does that - see line 10. `DEFAULTS.CONTAINER_ID_SUFFIX` will be the default, if no id and attention is passed in.
just added a unit test for this in the utils.test.tsx file

src/components/NotificationSystem/NotificationSystem.types.ts
torgeadelin: you can type it with Record<"string", Type>, i saw we used this in some places.
but again, i'm not too bothered

torgeadelin:
Could you try to make this warning go away:
image

torgeadelin (changes requested):
Some issue I found while used in Cantina:
1. Overlay issue:
image

Probably we need to set a very high z-index for the notification wrapper.

2. There is a weird background if i provide `AlertBanner` notification to `NotificationSystem.notify` without specifying any wrapper. that should't be there.
image

torgeadelin (approved):
LGTM!

PR #292, opened 2022-08-26
SPARK-356970 and SPARK-357657: add selected prop to button circle

nataliadelmar:
# Description
Add isSelected property to ButtonCircle outline ghost and ghost

image

# Links

[*Links to relevent resources.*](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=48160%3A80022)


maxinteger (changes requested):
Sorry, but not clear how this component supposed to work.

1 ) the button always behaves  as a toggle button because and no way to use it as a normal button
2) the buttons has `isSelected` but no `onChange` which means the component state stored outside of the button, but it also has the `useToggleState` and `useToggleButton` hooks which are used for keeping the toggle state inside the component

I think we should leave the ButtonCircle component as it is and create a new `ToggleButtonSimple` button which is using the `ButtonSimple` and the toggle hooks. This way clear for the user (developer) which button is capable of switching between 2 states or not.

Another option would be to add a new prop like `isToggle` and do some logic inside the component, but because you can not put hook inside `if` you might end up with a complicated solution.

Regarding how the toggle state, it should be stored inside the component and exposed all the related props `isSelected`, `onChange` etc. You can see an example in the `src/components/Toggle/Toggle.tsx` component which is work _almost_ the same way as the toggleButton.

nataliadelmar:
Resolved in https://github.com/momentum-design/momentum-react-v2/pull/302

PR #295, opened 2022-09-02
fix(use-divider-dot): use dividor-dot for second-line in space-list-item

jor-row:
# Description

Replaced the '-' between words in secondLine in SpaceListitem with a dividerDot

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-299144

Before: 
image
After: 
image


src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -55,8 +55,19 @@ const SpaceListItem: FC<Props> = forwardRef(
               type="body-secondary"
               data-test="list-item-second-line"
             >
-              {/* //TODO: change with dot divider when available */}
-              {_secondLineArray.join(' - ')}
+              {_secondLineArray[0]}
+              {_secondLineArray.length > 1 && (
+                <>
```
maxinteger, line 60: the Fragment tag is not needed here because react can handle automatically when when you try tor render a variable which is a list of components.

You need fragment when you have have manually (not by `map` or `reduce` etc.) created component but you do not need a container, like in line 63

jor-row: I've removed the unnecessary tag now

src/components/SpaceListItem/SpaceListItem.constants.ts
```diff
@@ -4,6 +4,7 @@ const CLASS_PREFIX = 'md-space-list-item';
 
 const DEFAULTS = {
   TEAM_COLOR: TEAM_COLORS.default,
+  DIVIDER_DOT: false,
```
torgeadelin, line 7: rename this to SHOW_DIVIDER_DOT (same as the prop, but all caps)

src/components/SpaceListItem/SecondLineElement.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { FC } from 'react';
```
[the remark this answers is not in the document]
torgeadelin: Also this component needs a test.

src/components/SpaceListItem/SpaceListItem.unit.test.tsx
```diff
@@ -294,6 +318,105 @@ describe('<SpaceListItem />', () => {
       ).toEqual({ 'data-test': 'compact-mode-divider-dot' });
     });
 
+    // Begins here
```
torgeadelin, line 321: remove comment

torgeadelin (changes requested):
A few small changes, but great work!
Praise for the extremely well test coverage! 🎉

src/components/SpaceListItem/SecondLineElement/SecondLineElement.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * render the secondLine array with DividerDots inbetween.
+ * @internal use only. Should not be consumed from any external parties.
+ */
+const SecondLineElement: FC<Props> = (props: Props) => {
+  const { children, showDividerDot = DEFAULTS.SHOW_DIVIDER_DOT } = props;
+
+  return (
+    <>
+      {showDividerDot && <DividerDot data-test="multiple-string-second-line-divider-dot" />}
+      <span data-test="second-line-element-child">{children}</span>
```
jor-row, line 18: I added the span here to allow for testing of the children prop for this component. Is there a better way to go about this? This seemed like the simplest way to go about it.

maxinteger: You are right, you should not add extra DOM element just for testing.
It is possible to check the text content of the root element with enzyme the which should be sufficient
see my comment in the test

src/components/SpaceListItem/SecondLineElement/SecondLineElement.unit.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  describe('attributes', () => {
+    it('should have provided children', async () => {
+      expect.assertions(1);
+
+      const container = mount(<SecondLineElement>{exampleChildProp}</SecondLineElement>)
+        .find('[data-test="second-line-element-child"]')
+        .getDOMNode();
+
+      expect(container.textContent).toBe(exampleChildProp);
```
maxinteger, line 39: You can test text children without additional node like this:
```suggestion
      const container = mount(<SecondLineElement>{exampleChildProp}</SecondLineElement>);

      expect(container.text()).toEqual(exampleChildProp);
```

src/components/SpaceListItem/SecondLineElement/SecondLineElement.unit.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    it('should have provided dot when dot is provided', () => {
+      expect.assertions(1);
+
+      const container = mount(
+        <SecondLineElement showDividerDot={showDividerDot}>{exampleChildProp}</SecondLineElement>
+      ).find(DividerDot);
+
+      expect(
+        container.filter("[data-test='multiple-string-second-line-divider-dot']").length
+      ).toEqual(1);
```
maxinteger, line 51: Same as above
```suggestion
      const container = mount(
        <SecondLineElement showDividerDot={showDividerDot}>{exampleChildProp}</SecondLineElement>
      );
      expect(container.text()).toEqual(exampleChildProp);
```

src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -43,7 +44,16 @@ const SpaceListItem: FC<Props> = forwardRef(
     const renderText = () => {
       const _secondLineArray: string[] = typeof secondLine === 'string' ? [secondLine] : secondLine;
 
-      if (secondLine) {
+      const _secondLineArrayClean = _secondLineArray?.reduce((filteredArray, nextElement) => {
+        const nextElementClean = nextElement.trim();
+        if (nextElementClean) {
+          filteredArray.push(nextElementClean);
+          return filteredArray;
```
maxinteger, line 51: this `return` is unnecessary

maxinteger (changes requested):

src/components/SpaceListItem/SecondLineElement/SecondLineElement.unit.test.tsx
jor-row: That's much simpler, thank you! I've changed those unit tests and removed the unnecessary dom element from SecondLineElement.

src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -43,7 +44,15 @@ const SpaceListItem: FC<Props> = forwardRef(
     const renderText = () => {
       const _secondLineArray: string[] = typeof secondLine === 'string' ? [secondLine] : secondLine;
 
-      if (secondLine) {
+      const _secondLineArrayClean = _secondLineArray?.reduce((filteredArray, nextElement) => {
+        const nextElementClean = nextElement.trim();
```
maxinteger, line 48: do we need to trim the value for rendering as well? Is it specified somewhere?
If you need trim just to filter out empty items, then better if to use the `filter` method

src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -43,7 +44,15 @@ const SpaceListItem: FC<Props> = forwardRef(
...
+        if (nextElementClean) {
+          filteredArray.push(nextElementClean);
+        }
+        return filteredArray;
+      }, []);
+
+      if (_secondLineArrayClean?.length) {
```
maxinteger, line 55: `_secondLineArrayClean` is always an array so the `?` is unnecessary

src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -54,9 +63,16 @@ const SpaceListItem: FC<Props> = forwardRef(
...
-              {/* //TODO: change with dot divider when available */}
-              {_secondLineArray.join(' - ')}
+              {_secondLineArrayClean.map((secondLineContent, i) => {
+                const showDividerDot = i > 0;
+                return (
+                  <SecondLineElement key={`second-line-item-${i}`} showDividerDot={showDividerDot}>
+                    {secondLineContent}
+                  </SecondLineElement>
+                );
+              })}
```
maxinteger, line 75: It is a minor thing but you can simplify this code a bit
```suggestion
              {_secondLineArrayClean.map((secondLineContent, i) => (
                  <SecondLineElement key={`second-line-item-${i}`} showDividerDot={i > 0}>
                    {secondLineContent}
                  </SecondLineElement>
                )
             )}
```

src/components/SpaceListItem/SpaceListItem.tsx
```diff
@@ -43,7 +44,15 @@ const SpaceListItem: FC<Props> = forwardRef(
     const renderText = () => {
       const _secondLineArray: string[] = typeof secondLine === 'string' ? [secondLine] : secondLine;
 
-      if (secondLine) {
+      const _secondLineArrayClean = _secondLineArray?.reduce((filteredArray, nextElement) => {
```
maxinteger, line 47: minor but please do not start the variable name with `_`
actually you can remove it from `_secondLineArray` as well.

maxinteger (changes requested):
I found some mostly cosmetic problems :)

src/components/SpaceListItem/SpaceListItem.tsx
jor-row: This is to filter out empty items but also to ensure that if a string with leading empty space is passed into the secondLineArray that extra space is not rendered between the dividerDot and the text and to keep those spaces even. See below:
Screenshot 2022-09-20 at 11 34 09
The first example is with trimming the value and the second is without but leading empty space in each string.

src/components/SpaceListItem/SpaceListItem.tsx
jor-row: I thought that if secondLine was not provided we would need to check but line 45 handles that. I've removed the ```?```s

src/components/SpaceListItem/SpaceListItem.tsx
jor-row: Noted. I've removed the ```_``` from any variables in this component.

src/components/SpaceListItem/SpaceListItem.tsx
jor-row: Removing the ```?``` causes all of the unit tests to fail...

maxinteger (approved):
Great work

src/components/SpaceListItem/SpaceListItem.constants.ts
torgeadelin: this was not renamed

src/components/SpaceListItem/SecondLineElement.tsx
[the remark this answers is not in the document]
torgeadelin: any reason this was ommited?

torgeadelin (approved):
Great work!

PR #301, opened 2022-09-06
Fix: MenuTrigger separator width 

lucas0:
# Description

defines the width of the separator as being 100% of its parent element (it was 0 before)

# Links

[SPARK-332388](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-332388)
Separator for MenuTrigger with multiple menus doesn't get displayed


torgeadelin (approved):
Looking good. Please rebase off master, the changes for the tab are here, but shouldn't be.
You've merged those as part of another PR.

maxinteger (approved):

PR #302, opened 2022-09-08
feat(toggle-circle-button): implement button circle toggle

nataliadelmar:
# Description

PR to implement the buttonCircleToggle component. This is a component that is based on ButtonCircle and adds toggle hooks to it to make it toggable/selectable. 

As for now, only the outline and ghost buttons have a toggable version designed in the component library, which only cover these button colour types in the scope.

image
image

# Links

[*Links to relevent resources.*](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=4445%3A5884)

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "@momentum-ui/animations": "^1.3.0",
-    "@momentum-ui/design-tokens": "^0.0.50",
+    "@momentum-ui/design-tokens": "^0.0.53",
     "@momentum-ui/icons": "8.28.4",
     "@momentum-ui/icons-rebrand": "^1.39.0",
     "@momentum-ui/tokens": "^1.7.1",
@@ -116,6 +116,7 @@
     "@react-stately/collections": "^3.3.4",
     "@react-stately/data": "^3.4.1",
     "@react-stately/select": "^3.1.3",
+    "@react-stately/toggle": "^3.4.1",
     "@react-stately/tree": "^3.2.0",
     "@react-types/button": "^3.4.0",
     "@react-types/select": "^3.3.1",
--- a/src/components/ButtonCircleToggle/ButtonCircleToggle.stories.args.ts
+++ b/src/components/ButtonCircleToggle/ButtonCircleToggle.stories.args.ts
@@ -1,0 +1,21 @@
+import { commonStyles, commonAriaButtonToggle } from '../../storybook/helper.stories.argtypes';
+import { BUTTON_CIRCLE_TOGGLE_CONSTANTS as CONSTANTS } from './';
+import { buttonCircleArgTypes } from '../ButtonCircle/ButtonCircle.stories.args';
+
+const buttonCircleToggleArgTypes = {
+  ...buttonCircleArgTypes,
+  ...commonAriaButtonToggle,
+};
+
+delete buttonCircleToggleArgTypes.color;
+
+buttonCircleToggleArgTypes.disabled.table.defaultValue.summary = CONSTANTS.DEFAULTS.DISABLED;
+buttonCircleToggleArgTypes.ghost.table.defaultValue.summary = CONSTANTS.DEFAULTS.GHOST;
+buttonCircleToggleArgTypes.outline.table.defaultValue.summary = CONSTANTS.DEFAULTS.OUTLINE;
+
+export { buttonCircleToggleArgTypes };
+
+export default {
+  ...commonStyles,
+  ...buttonCircleToggleArgTypes,
+};
--- a/src/components/ButtonCircleToggle/ButtonCircleToggle.stories.tsx
+++ b/src/components/ButtonCircleToggle/ButtonCircleToggle.stories.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+
+import { MultiTemplateWithPseudoStates, Template } from '../../storybook/helper.stories.templates';
+import { DocumentationPage } from '../../storybook/helper.stories.docs';
+import StyleDocs from '../../storybook/docs.stories.style.mdx';
+
+import Icon from '../Icon';
+import ButtonCircleToggle, { ButtonCircleToggleProps } from './';
+import argTypes from './ButtonCircleToggle.stories.args';
+import Documentation from './ButtonCircleToggle.stories.docs.mdx';
+
+export default {
+  title: 'Momentum UI/ButtonCircleToggle',
+  component: ButtonCircleToggle,
+  parameters: {
+    expanded: true,
+    docs: {
+      page: DocumentationPage(Documentation, StyleDocs),
+    },
+    args: {
+      style: { margin: '0.5rem' },
+    },
+  },
+};
+
+const Example = Template<ButtonCircleToggleProps>(ButtonCircleToggle).bind({});
+
+Example.argTypes = { ...argTypes };
+
+Example.args = {
+  children: <Icon name="cancel" autoScale={150} />,
+};
+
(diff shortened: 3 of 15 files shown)


src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -0,0 +1,48 @@
...
+    outline = DEFAULTS.OUTLINE,
+    disabled = DEFAULTS.DISABLED,
+    ...otherProps
+  } = props;
+
+  const ref = useRef();
+  const state = useToggleState(props);
+  const { buttonProps } = useToggleButton(props, state, ref);
+
+  if (ghost === false && outline === false) {
```
maxinteger, line 25: Is this a toggle button specific problem or should we have this warning for the `ButtonCircle` as well?

I think we should not show this error in production build so please add the `process.env.NODE_ENV !== "production"` condition as well
```suggestion
if (process.env.NODE_ENV !== "production")
   if (ghost === false && outline === false) {
```

if the build fails because of this extra condition then you can remove it for now

src/components/ButtonCircleToggle/ButtonCircleToggle.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+// delete Common.argTypes.children;
+
+// // TODO: Provide default arguments for this story here. These populate into the argument table for this component for all variants.
+// Common.args = {
+//   children: 'Example',
+// };
+
+// Common.parameters = {
+//   variants: [{ children: 'Example A' }, { children: 'Example B' }, { children: 'Example C' }],
+// };
```
maxinteger, line 44: please remove the commented code

src/components/ButtonCircleToggle/ButtonCircleToggle.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import { MultiTemplate, Template } from '../../storybook/helper.stories.templates';
```
maxinteger, line 3: please remove the unnecessary import

maxinteger (changes requested):
Only few minor thing, otherwise looks great

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
nataliadelmar: we should not have ths warning for buttonCircle; only for this toggle

maxinteger (approved):

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
torgeadelin: I think it's fine to add this warning in production. It's there for a reason to prevent the consumers to use this component in the wrong way.

src/components/ButtonCircleToggle/ButtonCircleToggle.style.scss
```diff
@@ -0,0 +1,40 @@
+.md-button-circle-toggle-wrapper {
+  &[data-selected='true'] {
+    &[data-outline='true'],
+    &[data-outline='true'][data-ghost='true'] {
```
torgeadelin, line 4: you shouldn't need to select both, one or the other is sufficient.

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const {
+    children,
+    className,
+    ghost = DEFAULTS.GHOST,
+    outline = DEFAULTS.OUTLINE,
+    disabled = DEFAULTS.DISABLED,
+    ...otherProps
+  } = props;
+
+  const ref = useRef();
```
torgeadelin, line 21: we need to use forward ref to provide more flexibility for this component.
Please look how we do that for ButtonSimple

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  if (process.env.NODE_ENV !== 'production') {
+    if (ghost === false && outline === false) {
+      console.warn(
+        'ButtonCircleToggle is only designed for outline or ghost circle buttons. Outline and ghost properties cannot be false at the same time.'
+      );
+    }
+  }
+
+  delete buttonProps.disabled;
```
torgeadelin, line 33: please add a comment to explain why this is needed

src/components/ButtonCircleToggle/ButtonCircleToggle.types.ts
```diff
@@ -0,0 +1,4 @@
+import { ButtonCircleProps } from '../ButtonCircle';
+import { AriaToggleButtonProps } from '@react-types/button';
+
+export interface Props extends ButtonCircleProps, AriaToggleButtonProps {}
```
torgeadelin, line 4: We don't want all the props from buttonCircle at this point, please omit `color` for now

torgeadelin: `extends Omit<ButtonCircleProps, 'color'>, ...`

src/components/ButtonCircleToggle/ButtonCircleToggle.stories.args.ts
```diff
@@ -0,0 +1,33 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+import { BUTTON_CIRCLE_TOGGLE_CONSTANTS as CONSTANTS } from './';
+import { buttonCircleArgTypes } from '../ButtonCircle/ButtonCircle.stories.args';
+
+const buttonCircleToggleArgTypes = {
+  ...buttonCircleArgTypes,
```
torgeadelin, line 6: it would be nice to create another arg type for toggleButton and include those as well.
We should at least have these 
image

torgeadelin: I'd create a reusable util like `commonStyles`

src/components/ButtonCircleToggle/ButtonCircleToggle.unit.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+      const element = mount(<ButtonCircleToggle isSelected={isSelected} />)
+        .find(ButtonCircleToggle)
+        .getDOMNode();
+
+      expect(element.getAttribute('data-selected')).toBe(`${isSelected}`);
+    });
+  });
+
+  describe('actions', () => {
+    it('should handle mouse press events', () => {
```
torgeadelin, line 126: the name is technically wrong, you're not pressing the button in the component, you're only checking the callback gets used correctly if provided.

I would instead try to click the button and check the onChange gets called instead of calling onChange manually.
Also add a test for the same when it's disabled and check the onChange doesn't get called;

torgeadelin: Also click twice for the first test to check that it goes from false -> true -> false a.k.a it toggles

torgeadelin (changes requested):
A few comments but looking 🍇

torgeadelin (commented):
LGTM!

package.json
```diff
@@ -116,6 +116,7 @@
     "@react-stately/collections": "^3.3.4",
     "@react-stately/data": "^3.4.1",
     "@react-stately/select": "^3.1.3",
+    "@react-stately/toggle": "^3.4.1",
```
torgeadelin, line 119: remove

torgeadelin (changes requested):
one small change and should be ready

torgeadelin (approved):
LGTM

PR #305, opened 2022-09-15
feat(button-circle-toggle-props): fix onChange prop

nataliadelmar:
# Description

React Aria's useToggleButton hook was engineered to be used in an html button, meaning it is providing the callback action via the onClick property to the button. We are using this hook in our existent ButtonCircle, which expects onPress as the callback method, not onClick. 

Also, properties passed to ButtonCircleToggle were not passed to ButtonCircle, like the properties to use a Popover triggered by this toggle. 

This PR intends to fix these two things in ButtonCircleToggle.

### DIFF
--- a/src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
+++ b/src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
@@ -8,7 +8,7 @@ import { DEFAULTS as BUTTON_CIRCLE_DEFAULTS } from '../ButtonCircle/ButtonCircle
 import { Size as ButtonCircleSize } from '../ButtonCircle/ButtonCircle.types';
 import { Props } from './ButtonCircleToggle.types';
 import './ButtonCircleToggle.style.scss';
-import { chain } from 'react-aria';
+import { chain } from '@react-aria/utils';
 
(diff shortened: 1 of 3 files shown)


src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -2,14 +2,13 @@ import React, { RefObject, forwardRef, useRef } from 'react';
...
 import { useToggleState } from '@react-stately/toggle';
-import { useToggleButton } from '@react-aria/button';
 
 import { DEFAULTS, STYLE } from './ButtonCircleToggle.constants';
 import { DEFAULTS as BUTTON_CIRCLE_DEFAULTS } from '../ButtonCircle/ButtonCircle.constants';
 import { Size as ButtonCircleSize } from '../ButtonCircle/ButtonCircle.types';
 import { Props } from './ButtonCircleToggle.types';
 import './ButtonCircleToggle.style.scss';
-import { AriaBaseButtonProps } from '@react-types/button';
+import { chain } from 'react-aria';
```
torgeadelin, line 11: import from `@ract-aria/utils`

torgeadelin (changes requested):
one small change and should be good to go

torgeadelin (approved):
giphy (1)

PR #306, opened 2022-09-15
feat(button-pill-toggle): implement button pill toggle

nataliadelmar:
# Description

PR to implement the buttonPillToggle component. This is a component that is based on ButtonPill and adds toggle hooks to it to make it toggable/selectable. 

As for now, only the outline and ghost buttons have a toggable version designed in the component library, which only cover these button colour types in the scope.

image
image

# Links

[*Links to relevent resources.*](https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=4445%3A5884)


maxinteger (changes requested):
image

1. Please add a small header for the 2 sections, like "Toggle off" / "Toggle on"
2. All the buttons look the same in stroybook

torgeadelin (approved):
giphy (2)

PR #307, opened 2022-09-20
fix(autofill-colours): remove yellow autofill box and styling fixes

jor-row:
# Description

This PR styles the autofill icon in Safari so that it is visible in our dark theming - the same changes as made in https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/3032 but generalising them to the momentum component. It also removes the yellow autofill box on Safari (except in the case of a suggested password) and Firefox, making it transparent. In the case of a suggested password in Safari the yellow box is restyled to have a border radius which aligns with that of the TextInput.

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-362673

Safari ->
Before
SAFARI-BEFORE
After 
SAFARI_AFTER

Firefox -> 
Before
FIREFOX-BEFORE
After
FIREFOX-AFTER

### DIFF
--- a/src/components/TextInput/TextInput.stories.tsx
+++ b/src/components/TextInput/TextInput.stories.tsx
@@ -103,14 +103,16 @@ const FormExample: FC<FormProps> = (props: FormProps) => {
       }}
     >
       {displayName && (
-        <TextInput
+        <BetterExample
+          initialValue=""
           placeholder={'Display Name'}
           type="name"
           style={{ marginTop: '1rem', marginBottom: '1rem', width: '16.25rem' }}
         />
       )}
       {emailAddress && (
-        <TextInput
+        <BetterExample
+          initialValue=""
           placeholder={'Email address'}
           type="email"
           style={{
@@ -121,7 +123,8 @@ const FormExample: FC<FormProps> = (props: FormProps) => {
         />
       )}
       {password && (
-        <TextInput
+        <BetterExample
+          initialValue=""
           type="password"
           placeholder={'Password'}
           style={{ marginBottom: '1rem', width: '16.25rem' }}
--- a/src/components/TextInput/TextInput.stories.tsx
+++ b/src/components/TextInput/TextInput.stories.tsx
@@ -1,5 +1,6 @@
 import React, { useState, FC, useRef } from 'react';
 import { MultiTemplate, Template } from '../../storybook/helper.stories.templates';
+import ButtonPill from '../ButtonPill';
 
 import TextInput, { TextInputProps } from './';
 
@@ -59,6 +60,12 @@ interface StoryProps extends TextInputProps {
   initialValue: string;
 }
 
+interface FormProps extends TextInputProps {
+  displayName?: boolean;
+  emailAddress?: boolean;
+  password?: boolean;
+}
+
 const PaddedExample: FC<TextInputProps> = (props: TextInputProps) => {
   return (
     <div style={{ margin: '1rem' }}>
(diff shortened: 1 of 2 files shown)


maxinteger (changes requested):
Could you please create a new story in storybook with a login form to test these scenarios?

for example 
```
const LoginForm = () => (
  <form>
    <PaddedExample label={'Email'} type={'email'} />
    <PaddedExample label={'Password'} type={'password'} />
    <button type={'submit'}>Submit</button>
  </form>
);
```

Probably you can add some text as well to explain why it is needed and you can use the Momentum button for submit :)

Also the fix does not work when you ask Safari to suggest a new password:
image

jor-row:
@maxinteger I've added in two 'form' stories to storybook. This is the first time I've added stories to a component so any improvements or things I've missed are more than welcome!

src/components/TextInput/TextInput.stories.tsx
```diff
@@ -85,10 +92,64 @@ const BetterExample: FC<StoryProps> = (props: StoryProps) => {
   return <TextInput {...mutatedProps} value={value} ref={ref} onChange={onChange} />;
 };
 
+const FormExample: FC<FormProps> = (props: FormProps) => {
+  const { displayName = false, emailAddress = false, password = false } = props;
+
+  return (
+    <>
+      <form>
```
maxinteger, line 100: Please add the following:
```suggestion
    <form
      onSubmit={(e) => {
        e.preventDefault();
        location.reload();
      }}
    >
```
this will prevent to reload the whole storyboard, but it will reload the `iframe` of the story which triggers the browser password manager to store the email / password

src/components/TextInput/TextInput.style.scss
```diff
@@ -39,6 +39,48 @@
     &::placeholder {
       color: var(--textinput-placeholder-text);
     }
+
+    &::-webkit-contacts-auto-fill-button {
```
maxinteger, line 43: I just noticed, the background of the input changes after autofill in chrome:
image

Not sure we need to handle it 
Any idea @nataliadelmar or @torgeadelin ?

maxinteger (changes requested):

src/components/TextInput/TextInput.stories.tsx
```diff
@@ -85,10 +92,64 @@ const BetterExample: FC<StoryProps> = (props: StoryProps) => {
...
+          />
+        )}
+        {password && (
+          <TextInput
+            type="password"
+            placeholder={'Password'}
+            style={{ marginBottom: '1rem', width: '16.25rem' }}
+          />
+        )}
+        <ButtonPill size={32} grown style={{ marginBottom: '1rem', width: '16.25rem' }}>
```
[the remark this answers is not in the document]
jor-row: I've changed to the text to "Submit". My only reasoning for using "Next" was to mimic the log on pages seen on web.webex.

src/components/TextInput/TextInput.stories.tsx
```diff
@@ -85,10 +92,64 @@ const BetterExample: FC<StoryProps> = (props: StoryProps) => {
   return <TextInput {...mutatedProps} value={value} ref={ref} onChange={onChange} />;
 };
 
+const FormExample: FC<FormProps> = (props: FormProps) => {
+  const { displayName = false, emailAddress = false, password = false } = props;
+
+  return (
+    <>
```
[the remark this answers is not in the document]
jor-row: That explains a lot, thanks!

src/components/TextInput/TextInput.style.scss
jor-row: To be honest I haven't been testing these styling changes in Chrome as the option to use autofill wasn't coming up for me. I see how to use it now. 

For some reason the animation won't work in chrome, and from trying a few things quickly can't seem to override the styling on chrome to make it transparent like I have in the other two browsers. 

The awkward thing is I can "change" the colour by overlaying a coloured box-shadow but that of course changes it in Safari (which causes the autofill icon colours to not be visible and the text to be less visible) and Firefox. 

In the meantime while addressing the other comments I have styled it so the edges are at least curved and it fits in the text box. 
 
Screenshot 2022-09-22 at 17 55 22


Any ideas?

maxinteger (approved):

src/components/TextInput/TextInput.style.scss
nataliadelmar: Is that a purple? This color doesn't look off to me like the bright yellow did so I am happy to keep it this way, including the rounded corners. It still passes color contrast and is in tone with the rebrand colors, which are only dark mode because this logged-out part of the app is always in dark mode right?

jor-row: Yes, I've been told that these log on pages are always in the default dark mode as light mode is set as a user preference whilst logged in. I can't think of any time within the app that a form would be used with these kinds of entries that would prompt the browser to use autofill, so I think it should be fine? There may, however, be features within the app that I'm unaware of that would do this. 

As for the X in the input field, I hadn't noticed but yes there should be an X in the black square and there will be when it's used in Cantina. The reason there isn't one in my stories is I haven't used the "value" prop of TextInput which tells the component whether to show it or not. To make it work in my stories I'll have to add separate state to each form field to handle it which I'm working on now as I suppose it's worth including.

nataliadelmar: I understand. In my opinion, the short case that a user will not see an X button until clicking on the field in Chrome if there is autofilled data in Chrome only is small enough to let this case go. What do you think @maxinteger ?

maxinteger: I agree with @nataliadelmar this is kind of an edge.

PR #310, opened 2022-09-23
fix(search-input): fix styling to align with figma

jor-row:
# Description

This PR fixes some of the styling in the SearchInput component:

- Changes weight of icons to "bold"
- Centres the icons and text within the wrapper div 
- Moves the icons closer to the sides of the wrapper
- Adds the two height variations as seen in the Figma and adds a default width
- Replaces the old 'searching' icon with the LoadingSpinner component

I had a look at why the 'searching' icon doesn't seem to work and it's to do with how it's being used in Cantina - the boolean 'searching' prop isn't being used and so the component never changes the icons.

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-356975

https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=25054%3A38561

Before: 
Screenshot 2022-09-23 at 15 31 49

After: 
Screenshot 2022-09-23 at 15 31 58

### DIFF
--- a/src/components/SearchInput/SearchInput.constants.ts
+++ b/src/components/SearchInput/SearchInput.constants.ts
@@ -5,7 +5,7 @@ const CLASS_PREFIX = 'md-search-input';
 
 const HEIGHTS = [28, 32] as const;
 
-const SEARCH_ICON_HEIGHT_MAPPING: Record<SearchFieldHeight, IconScale> = {
+const ICON_HEIGHT_MAPPING: Record<SearchFieldHeight, IconScale> = {
   28: 12,
   32: 16,
 };
@@ -22,4 +22,4 @@ const STYLE = {
   clear: `${CLASS_PREFIX}-clear`,
 };
 
-export { CLASS_PREFIX, HEIGHTS, SEARCH_ICON_HEIGHT_MAPPING, DEFAULTS, STYLE };
+export { CLASS_PREFIX, HEIGHTS, ICON_HEIGHT_MAPPING, DEFAULTS, STYLE };
--- a/src/components/SearchInput/SearchInput.constants.ts
+++ b/src/components/SearchInput/SearchInput.constants.ts
@@ -1,20 +1,18 @@
 import { IconScale } from '../Icon';
+import { SearchFieldHeight } from './SearchInput.types';
 
 const CLASS_PREFIX = 'md-search-input';
 
-const HEIGHTS = {
-  28: 28,
-  32: 32,
-};
+const HEIGHTS = [28, 32] as const;
 
-const SEARCH_ICON_HEIGHT_MAPPING: Record<number, IconScale> = {
+const SEARCH_ICON_HEIGHT_MAPPING: Record<SearchFieldHeight, IconScale> = {
   28: 12,
   32: 16,
 };
 
 const DEFAULTS = {
-  HEIGHT: HEIGHTS[32],
-};
+  HEIGHT: 32,
+} as const;
 
 const STYLE = {
   wrapper: `${CLASS_PREFIX}-wrapper`,
--- a/src/components/SearchInput/SearchInput.constants.ts
+++ b/src/components/SearchInput/SearchInput.constants.ts
@@ -1,6 +1,20 @@
+import { IconScale } from '../Icon';
+
 const CLASS_PREFIX = 'md-search-input';
 
-const DEFAULTS = {};
+const HEIGHTS = {
+  28: 28,
+  32: 32,
+};
+
+const SEARCH_ICON_HEIGHT_MAPPING: Record<number, IconScale> = {
+  28: 12,
+  32: 16,
+};
+
+const DEFAULTS = {
+  HEIGHT: HEIGHTS[32],
+};
 
 const STYLE = {
   wrapper: `${CLASS_PREFIX}-wrapper`,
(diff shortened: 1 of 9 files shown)


src/components/SearchInput/SearchInput.constants.ts
```diff
@@ -1,12 +1,27 @@
+import { IconScale } from '../Icon';
+
 const CLASS_PREFIX = 'md-search-input';
 
-const DEFAULTS = {};
+const HEIGHTS = {
+  28: 28,
+  32: 32,
+};
```
maxinteger, line 8: since the keys and values are the same, better if you are us an array instead 
```suggestion
const HEIGHTS = [28, 32] as const;
```

Note: the `as const` mean TS will infer `[28, 32]` type for the `HEIGHTS` variable instead `number[]`

src/components/SearchInput/SearchInput.constants.ts
```diff
@@ -1,12 +1,27 @@
...
+
+const SEARCH_ICON_HEIGHT_MAPPING: Record<number, IconScale> = {
+  28: 12,
+  32: 16,
+};
+
+const DEFAULTS = {
+  HEIGHT: HEIGHTS[32],
+};
```
maxinteger, line 17: Here just use the number as the default height
Also I recommend to add the `as const`, see the reason above

```suggestion
const DEFAULTS = {
  HEIGHT: 32,
} as const;
```

src/components/SearchInput/SearchInput.stories.args.ts
```diff
@@ -26,4 +27,13 @@ export default {
       category: 'Props',
     },
   },
+  height: {
+    description: 'Changes the height of the component',
+    options: [...Object.values(HEIGHTS)],
```
maxinteger, line 32: If `HEIGHT` is an array you can do this:
```suggestion
    options: HEIGHTS,
```

src/components/SearchInput/SearchInput.stories.tsx
```diff
@@ -122,16 +122,19 @@ const BetterExample: FC<SearchInputExampleProps> = (props: SearchInputExamplePro
...
 /**
  * Common variants story. This renders multiple variants of a single component.
  */
 
 const PaddedExample: FC<SearchInputProps> = (props: SearchInputProps) => {
   return (
-    <div style={{ margin: '1rem' }}>
-      <SearchInput style={{ width: '20rem' }} {...props} />
-    </div>
+    <>
```
maxinteger, line 133: fragment tag is redundant, since the `<div>` is the only root component

src/components/SearchInput/SearchInput.types.ts
```diff
@@ -1,6 +1,8 @@
 import { CSSProperties } from 'react';
 import { AriaSearchFieldProps } from '@react-types/searchfield';
 
+export type SearchFieldHeight = 28 | 32;
```
maxinteger, line 4: The problem is, you defined the height 2 places, here and in the `SearchInput.constants.ts`. You should have a single source of truth. You can reuse the `HEIGHT` constant here:

```suggestion
export type SearchFieldHeight = typeof HEIGHT[number];
```

This is not a perfect solution because we are importing a value into a type definition files, but from the TS point of view both just .ts files so ¯\_(ツ)_/¯

Another solution would be to convert union type to array type (`1 | 2 | 3` => `[1, 2, 3]`) but it involves some type magic... maybe next time

maxinteger (changes requested):

src/components/SearchInput/SearchInput.stories.tsx
jor-row: I think that was left over from when I was testing and had added in another div there. I'll take a not to have one last look over anything like that I've left from messing around with things.

src/components/SearchInput/SearchInput.stories.tsx
```diff
@@ -130,7 +131,7 @@ Example.argTypes = { ...argTypes };
 const PaddedExample: FC<SearchInputProps> = (props: SearchInputProps) => {
   return (
     <div style={{ margin: '1rem' }}>
-      <SearchInput style={{ width: '20rem' }} {...props} />
+      <SearchInput height={32} {...props} />
```
jor-row, line 134: Todo: remove height prop as 32 is the default value

src/components/SearchInput/SearchInput.style.scss
```diff
@@ -1,17 +1,26 @@
...
   background-color: var(--textinput-background);
   border: 0.0625rem solid var(--textinput-border-color);
   box-sizing: border-box;
-  height: 1.75rem;
+  height: 2rem;
+  width: 16.25rem;
   display: flex;
   align-content: center;
+  justify-content: center;
+  align-items: center;
```
jor-row, line 14: Todo: are lines 12-14 all needed?

src/components/SearchInput/SearchInput.constants.ts
```diff
@@ -1,12 +1,25 @@
+import { IconScale } from '../Icon';
+import { SearchFieldHeight } from './SearchInput.types';
+
 const CLASS_PREFIX = 'md-search-input';
 
-const DEFAULTS = {};
+const HEIGHTS = [28, 32] as const;
+
+const SEARCH_ICON_HEIGHT_MAPPING: Record<SearchFieldHeight, IconScale> = {
```
jor-row, line 8: Todo: change naming to ICON_HEIGHT_MAPPING as used for searching and cancel icon too

maxinteger (approved):
Nice one

PR #313, opened 2022-09-26
fix(avatar): add size 64 and change shape of presence icon 

jor-row:
# Description

This PR aligns the avatar component more with what appears on the Figma:

- Add avatar size 64
- Change shape of presence icon wrapper depending on presence used
- Moved story argTypes to a separate file
- Removed unnecessary args from each story in storybook

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-367923

https://www.figma.com/file/vdL18BATeJAIq2JvGAjRPD/Components---Web?node-id=41210%3A65884

Before: 
Screenshot 2022-09-26 at 14 07 15
Screenshot 2022-09-26 at 14 06 34

After: 
Screenshot 2022-09-26 at 14 03 14
Screenshot 2022-09-26 at 14 06 20

### DIFF
--- a/src/components/Avatar/Avatar.unit.test.tsx
+++ b/src/components/Avatar/Avatar.unit.test.tsx
@@ -130,7 +130,7 @@ describe('Avatar', () => {
     it('should pass the size prop', () => {
       expect.assertions(1);
 
-      const size = SIZES[48] as AvatarSize;
+      const size = SIZES[2] as AvatarSize;
 
       const element = mount(<Avatar initials="CW" size={size} />)
         .find(`div.${STYLE.wrapper}`)
--- a/src/components/Avatar/Avatar.constants.ts
+++ b/src/components/Avatar/Avatar.constants.ts
@@ -17,12 +17,12 @@ const SIZES = {
 };
 
 const AVATAR_PRESENCE_ICON_SIZE_MAPPING: Record<number, IconScale> = {
-  24: 8,
-  32: 12,
-  48: 14,
-  72: 16,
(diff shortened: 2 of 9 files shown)


src/components/Avatar/Avatar.unit.test.tsx
```diff
@@ -130,7 +130,7 @@ describe('Avatar', () => {
     it('should pass the size prop', () => {
       expect.assertions(1);
 
-      const size = SIZES[48] as AvatarSize;
+      const size = SIZES[2] as AvatarSize;
```
Coread, line 133: question: Why use SIZES[2] instead of just the number?

jor-row: It doesn't really make a difference. Just the number may be more immediately readable but if it's SIZES[...] in VSCode you can hover over the SIZES and see what other sizes are supported. I see I used just the number in the constants file.

Is there a preference?

Coread: no real preference. Just thought that if we add a new size at the beginning of the array this test will break. But not a big deal, no need to change it

Coread (approved):

src/components/Avatar/Avatar.unit.test.tsx
jor-row: I've kept it as it is. This test in particular won't break as long as it's a valid size, but if I use just the number and that size gets removed from the SIZES in constants it will break I think

PR #316, opened 2022-09-27
fix(space-list-item): cover null case for second-line

jor-row:
# Description

This PR covers the case where secondLine can be passed as null as a prop. I separated out some of the logic to a utils file and wrote a test to cover edge cases.


cscheffauer (approved):
LGTM! great job!

PR #319, opened 2022-10-14
fix: update list accessibility

maxinteger:
- fix shift+tab to the first list item
- can open menu inside list item
- add unit tests
- update list story

### DIFF
--- a/src/components/List/List.unit.test.tsx
+++ b/src/components/List/List.unit.test.tsx
@@ -236,25 +236,25 @@ describe('<List />', () => {
     it('should handle focus on tabbable elements in the list row', () => {
       expect.assertions(11);
 
-      const { getAllByTestId, getAllByRole } = render(
+      const { getAllByRole } = render(
         <List listSize={2}>
           <ListItemBase key="0" itemIndex={0}>
             <ListItemBaseSection>
-              <input data-testid="input" />
-              <button data-testid="button">Button 1</button>
+              <input type="text" />
+              <button>Button 1</button>
             </ListItemBaseSection>
           </ListItemBase>
           <ListItemBase key="1" itemIndex={1}>
             <ListItemBaseSection>
-              <input data-testid="input" />
-              <button data-testid="button">Button 2</button>
+              <input type="text" />
+              <button>Button 2</button>
             </ListItemBaseSection>
           </ListItemBase>
         </List>
       );
       const listItems = getAllByRole('listitem');
-      const inputs = getAllByTestId('input');
-      const buttons = getAllByTestId('button');
+      const inputs = getAllByRole('textbox');
+      const buttons = getAllByRole('button');
 
       expect(listItems[0]).toHaveFocus();
 
@@ -289,7 +289,7 @@ describe('<List />', () => {
       expect(listItems[1]).toHaveFocus();
     });
 
-    it('should handle manu in the list item', async () => {
+    it('should handle menu in the list item', async () => {
       expect.assertions(2);
 
       const { getAllByRole, findAllByText } = render(
--- a/src/components/List/List.unit.test.tsx
+++ b/src/components/List/List.unit.test.tsx
@@ -3,6 +3,13 @@ import { mount } from 'enzyme';
 
 import List, { LIST_CONSTANTS as CONSTANTS } from './';
 import ListItemBase from '../ListItemBase';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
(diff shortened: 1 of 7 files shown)


src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -90,22 +91,16 @@ const ListItemBase = (props: Props, providedRef: RefObject<HTMLLIElement>) => {
...
-      'a[href], button, input, textarea, select, details,[tabindex]:not([tabindex="-1"])'
-    );
-    return Array.from(result).filter(
-      (el) => !el.hasAttribute('disabled') && !el.getAttribute('aria-hidden')
-    );
-  }
-
   useEffect(() => {
-    if (!listContext?.currentFocus) {
+    if (listContext?.currentFocus === undefined) {
```
torgeadelin, line 103: nice catch! I saw this the other day, and the reason we couldn't focus on the first item was due to 0 = false

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -212,7 +207,6 @@ const ListItemBase = (props: Props, providedRef: RefObject<HTMLLIElement>) => {
         data-interactive={interactive}
         className={classnames(className, STYLE.wrapper, { active: isPressed || isSelected })}
         role={role}
-        {...pressProps}
```
torgeadelin, line 215: Why was this removed?

src/components/ListItemBase/ListItemBase.utils.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { getKeyboardFocusableElements } from './ListItemBase.utils';
```
torgeadelin, line 1: Praise for moving out of the component and adding tests!

src/components/List/List.unit.test.tsx
```diff
@@ -139,4 +146,174 @@ describe('<List />', () => {
       expect(element.getAttribute('style')).toBe(styleString);
     });
   });
+
+  describe('list focus', () => {
```
torgeadelin, line 150: Praise for adding these tests.

src/components/ListItemBase/ListItemBase.tsx
torgeadelin: This breaks the ability for users to have onPress on the list item which is not wha we want

torgeadelin (changes requested):

src/components/ListItemBase/ListItemBase.tsx
maxinteger: You are right I removed for testing, but I forgot to add it back

src/components/List/List.unit.test.tsx
```diff
@@ -139,4 +146,174 @@ describe('<List />', () => {
...
+    });
+
+    it('should handle focus on tabbable elements in the list row', () => {
+      expect.assertions(11);
+
+      const { getAllByTestId, getAllByRole } = render(
+        <List listSize={2}>
+          <ListItemBase key="0" itemIndex={0}>
+            <ListItemBaseSection>
+              <input data-testid="input" />
```
cscheffauer, line 243: instead of using data-testid, i would recommend to use getByRole('textbox') and getByRole('button') for quering. using data-testid should be avoided if possible. if u want to narrow it to the specific input/button, u can use the "name" option, see here:
https://testing-library.com/docs/queries/byrole#api

src/components/List/List.unit.test.tsx
```diff
@@ -139,4 +146,174 @@ describe('<List />', () => {
...
+      expect(buttons[1]).toHaveFocus();
+
+      userEvent.tab({ shift: true });
+      expect(inputs[1]).toHaveFocus();
+
+      userEvent.tab({ shift: true });
+      expect(listItems[1]).toHaveFocus();
+    });
+
+    it('should handle manu in the list item', async () => {
```
cscheffauer, line 292: typo `manu`

cscheffauer (changes requested):
Have u tested these changes with Cantina?
like linking locally and check that all e2e tests in Cantina pass etc.?
would be good, since fundamental changes have been made

src/components/List/List.unit.test.tsx
maxinteger: I agree, `data-testid` should be avoided, but in this case the input is created only for the unit test and it is not part of the production code

torgeadelin (approved):
Looking great!

PR #325, opened 2022-11-16
feat(popover): fix click outside bugs

torgeadelin:
# Description
I have managed to find the cause of the click outside bug in Popover combined with mrv2 buttons.
See the comment in code for explanation.

# Links
[SPARK-320405](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-320405)


torgeadelin:
I'm noticing this fix is breaking the `onPress`.
I will investigate more...

torgeadelin:
Ok, I've slightly changed the approach, still similar to one before, but should keep full functionality of the original `onPressDown` implementation inside @react-aria, with one exception for draggable elements which I think can be addressed later if necessary.

torgeadelin:
Please try to test manually if anything else has broken. Hopefully not!

src/components/ButtonSimple/ButtonSimple.tsx
```diff
@@ -22,6 +22,24 @@ const ButtonSimple = forwardRef((props: Props, providedRef: RefObject<HTMLButton
...
+   * The main cause is the fact that they preventDefault the event for any event that is not
+   * draggable. We intercept the event and slightly modify it so that the event is not
+   * prevent defaulted. This does rly heavily on the implementation of how usePress works,
+   * but the other solution would be to fork the package and maintain it ourselves.
+   *
+   * Original implementation of the onPointerDown form @react-aria/interactions
+   * can be found here node_modules/@react-aria/interactions/src/usePress.ts
+   */
+  const onPointerDown: PointerEventHandler<HTMLButtonElement> = useCallback((event) => {
+    if (event?.button !== 0 || !event?.currentTarget.contains(event?.target as HTMLElement)) return;
```
cscheffauer, line 37: should we only do this on buttons or other elements as well? like list items?

torgeadelin: This is a very good point actually.
I'm realising we need the exact same logic wherever we use `usePress` or `useButton`  hook from Aria.
Looks like thats:
- list item base (uses usePress)
- button hyperlink
- select 

Seems super hacky to copy paste the same fix in all these places.

Another solution I was thinking about is to introduce an invisible layer below any overlay displayed on the screen. Clicking anywhere outside would mean clicking the back-overlay and that would save a lot of problems.

I'm just thinking how we should approach the issue where:
1. you click on a popover (gets open)
2. click on a different popover(to open another one)
3. At this point the expected behaviour is (click on the back-overlay detected, close first popover)
4. BUT, assuming the click was practically on top of the second's popover trigger, will that event go down to the second popover trigger and trigger the open of the second one?

Just looking at what the best way to approach this is..

maxinteger: @torgeadelin 
The other problem with this solution is, we do not know where reactAria or other libs using the `draggable` property, so it can have unexpected side effect.

I think adding an invisible backdrop is better solution here. Each popover should has it's own backdrop this way they will be stacked and when the user click outside the popover it will be caught by the most recently opened backdrop and it will close the related popover

This will have UX side effect, if you have a open popover and you want to click on a button outside the popover you have to click 2 times, first will close the popover and the second will hit the button. BUT this is normal, you can see the same effect in the OS or in web UI libs as well for example on GitHub.

cscheffauer: @torgeadelin @maxinteger agree, the invisible backdrop sounds like a more feasible & stable solution!

torgeadelin: ok, i'll go ahead and do that instead.

torgeadelin:
Closing in favour of going for a different solution. Will open new PR once ready.

PR #331, opened 2022-11-24
feat(popover): add backdrop

torgeadelin:
# Description
- added backdrop
- fixed useNativeKeydown bug
- added tests

# Links
[SPARK-320405](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-320405)


src/components/Popover/plugins.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Due to @react-aria's event abstraction layer, we add an invisible backdrop for all popovers
+ * to not interfere with any components built with the hooks from @react-aria.
+ *
+ * This will make sure that popovers are closed on clickoutside and no longer
+ * be stuck open in case the even target was a @react-aria button.
+ */
+export const addBackdrop = {
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
maxinteger, line 12: I removed this line locally and I did not get any error
If you want to be sure you can define the type of the `addBackdrop` variable

```
import type { Plugin } from 'tippy.js';
// ...
export const addBackdrop: Plugin = {
```

src/components/Popover/Popover.unit.test.tsx
```diff
@@ -141,6 +145,36 @@ describe('<Popover />', () => {
 
       expect(container).toMatchSnapshot();
     });
+
+    it('should display only one popover at all time', async () => {
+      expect.assertions(6);
+
+      const { container } = render(
+        <>
+          <Popover triggerComponent={<ButtonSimple>Popover 1</ButtonSimple>} strategy="fixed">
```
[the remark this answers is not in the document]
torgeadelin: added one test for the other strategy as well

src/components/Popover/Popover.unit.test.tsx
```diff
@@ -141,6 +145,36 @@ describe('<Popover />', () => {
...
+            <p>Content</p>
+          </Popover>
+          <Popover triggerComponent={<ButtonSimple>Popover 2</ButtonSimple>} strategy="fixed">
+            <p>Content</p>
```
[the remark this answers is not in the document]
torgeadelin: fixed

src/components/Popover/Popover.unit.test.tsx
```diff
@@ -171,6 +205,67 @@ describe('<Popover />', () => {
...
+      const contentBeforeClickPopover2 = screen.queryByText('Content 2');
+      expect(contentBeforeClickPopover2).not.toBeInTheDocument();
+
+      await openPopoverByClickingOnTriggerAndCheckContent(/Popover 1/i, /Content 1/i);
+
+      await openPopoverByClickingOnTriggerAndCheckContent(/Popover 2/i, /Content 2/i);
+
+      // assert that first popover has closed, and only second one is open
+      const contentAfterClickingBoth = screen.queryByText('Content 1');
+      expect(contentAfterClickingBoth).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
maxinteger: you are right, but what about the check after the click in line 271

src/components/Popover/tippyPlugins.ts
```diff
@@ -22,3 +24,35 @@ export const hideOnEscPlugin: Plugin = {
...
 };
+
+/**
+ * Due to @react-aria's event abstraction layer, we add an invisible backdrop for all popovers
+ * to not interfere with any components built with the hooks from @react-aria.
+ *
+ * This will make sure that popovers are closed on clickoutside and no longer
+ * be stuck open in case the even target was a @react-aria button.
+ */
+export const addBackdrop: Plugin = {
```
maxinteger, line 35: could you please add unit test(s)?

src/components/Popover/tippyPlugins.ts
```diff
@@ -22,3 +24,35 @@ export const hideOnEscPlugin: Plugin = {
...
+  name: 'addBackdrop',
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
maxinteger, line 37: I think you do not need this comment anymore

maxinteger (changes requested):

src/components/Popover/plugins.ts
torgeadelin: removed

src/components/Popover/Popover.unit.test.tsx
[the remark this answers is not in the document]
torgeadelin: the click on 271 is just a button which does nothing.
I've added a check to make sure the first popover is not visible either (altho i've checked that before clicking)

src/components/Popover/tippyPlugins.ts
torgeadelin: there is a small unit test already, not sure how much more I can test this, it's technically just an object.

torgeadelin: im open to suggestions tho

maxinteger: the `onMount` and the `onHide` methods are contains some logic and you can test that

For example `onMount` does nothing when the trigger is `mouseenter` or the `onHide` remove only one layer of backdrop

```ts
document.body.innerHTML = '<div [data-tippy-root]' />'
const tippyInstance = { props: { trigger: 'click'}}
const pluginInstance = addBackdro.fn(tippyInstance)
pluginInstance.onMount()
expect(/**backdrop exists*/)
pluginInstance.onHide()
expect(/**backdrop does not exists*/)
```

maxinteger (approved):

PR #332, opened 2022-11-28
Feat/list/fix list focus on change

torgeadelin:
# Description
- fixed list focus auto-focus on every update.
- added story with search + list to illustrate the focus works
- update rtl user event to new version as it was buggy
- lots of updates on tests due to rtl upgrade
   - userEvent recommended to be setup at beginning of test
   - use use object returned and simulate events on it
   - new methods returns promise hence the `await` everywhere

# Links


src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -121,13 +122,34 @@ const ListItemBase = (props: Props, providedRef: RefObject<HTMLLIElement>) => {
...
+   * case the index of the element focused before the list shrink is now outside
+   * the size of the new list size (shrinked size)
+   */
+  useEffect(() => {
+    if (listContext?.currentFocus >= listContext?.listSize) {
+      // set focus to first item
+      listContext.setContext(0);
+      updateTabIndexes();
+      // mark the update as silent (don't actually DOM focus the element)
+      setShouldIgnoreFocusUpdate(true);
```
maxinteger, line 145: this is incorrect, in the search use case the list can shrink when user add more chars to the search term but it can grow as well when suer remove chars from the term

Here you try to exclude cases where focus should not happen, maybe you can try the opposite and the find the cases when focus must be called

If I understand correctly focus should be called only when 
- the user select a row by keyboard or mouse or
- you reset the selected index And the last focused item was inside the list
Probably you can add an extra `isNeedFocus` parameter to the `setContext`

src/components/ListItemBase/ListItemBase.utils.tsx
```diff
@@ -6,3 +8,18 @@ export function getKeyboardFocusableElements<T extends HTMLElement>(root: T): El
     (el) => !el.hasAttribute('disabled') && el.getAttribute('aria-hidden') !== 'true'
   );
 }
+/**
+ * Same as useEffect but ignores first render.
+ * @param fn - use effect callback
+ * @param inputs dependencies
+ */
+export const useDidUpdateEffect = (fn: EffectCallback, inputs: DependencyList): void => {
```
maxinteger, line 16: missing unite test

maxinteger (changes requested):
Added few observation

src/components/Popover/Popover.unit.test.tsx
```diff
@@ -303,10 +316,12 @@ describe('<Popover />', () => {
       expect(props.onDestroy).toBeCalled();
     });
 
-    it('should show/hide Popover on tab + enter', async () => {
+    it.only('should show/hide Popover on tab + enter', async () => {
```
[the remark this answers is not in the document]
torgeadelin: hehe, removed

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -121,13 +122,34 @@ const ListItemBase = (props: Props, providedRef: RefObject<HTMLLIElement>) => {
...
+  }, [listContext?.currentFocus]);
+
+  /**
+   * When the items inside the list context gets smaller (search/filter applied)
+   * we want to silently update the currentFocus back to the first element in
+   * case the index of the element focused before the list shrink is now outside
+   * the size of the new list size (shrinked size)
+   */
+  useEffect(() => {
+    if (listContext?.currentFocus >= listContext?.listSize) {
```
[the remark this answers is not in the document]
torgeadelin: yeah, seems kinda cryptic.
added as separate check

src/components/ListItemBase/ListItemBase.tsx
torgeadelin: Focus happens at the item level, not in the context. the context only stores the current focused item and the list item base will automatically focus depending on that current focus variable form the context. in short, the state is lifted but the focusing mechanism is still in the list item.

It does feel correct to me because:
- when the list grows, that means that the focus must have been within the range of the previous list size, and the new items added (even though they can appear at the beginning, or end of list) can't really be refocused based on what was focused previously because I'm not storing the mapping of `itemIndex` -> `itemFocused`.
- when the list shrinks, this is where resetting to 0 when the focus overflows always makes sense, and also, if it doesn't it will focus back on the same item (in terms of which one was focused before based on position, not the actual item). Again, I don't think I can focus back exactly to the one that was focused before because I'm not storing the mapping of `itemIndex` -> `itemFocused`.

With the current solution, did you find any bugs?
If not, I might suggest to leave it as is due to time constraints.

src/components/ListItemBase/ListItemBase.utils.tsx
torgeadelin: added

src/components/ListItemBase/ListItemBase.tsx
maxinteger: t looks good in storybook, so you can leave it as it is. Probably we can revisit this problem later and fix with ordering.

maxinteger (approved):

PR #334, opened 2022-12-05
Feat/components/theme provider/add momentum design themes

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to add support for the new Momentum Design themes and scheme within the `<ThemeProvider />` component. These themes are not yet mapped to the components directly, but the changes provided here will allow for the adoption of these newer, automated tokens.

### DIFF
--- a/config/storybook/cjs/globals.js
+++ b/config/storybook/cjs/globals.js
@@ -1,6 +1,6 @@
 import {
   DEFAULTS,
-  THEME_NAMES,
+  THEMES,
 } from '../../../dist/cjs/components/ThemeProvider/ThemeProvider.constants';
 
 const globals = {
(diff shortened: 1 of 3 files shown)


src/components/ThemeProvider/ThemeProvider.types.ts
```diff
@@ -14,6 +14,8 @@ export type ThemeNames =
   | 'lightRose'
   | 'lightWebex';
 
+export type MDTheme = 'webex-dark' | 'webex-light';
```
cscheffauer, line 17: this could be typed more stable against changes, with:
`export type MDTheme = typeof MD_THEMES[keyof typeof MD_THEMES];`

InteractiveTimmy: This produces a typeof `string`, which isn't explicit.

Screenshot 2022-12-05 at 6 14 25 PM

The preferred route is to cast the constant as the type, but, this removes the ability to auto-cast non-expansive constants using `keyof typeof` due to a circular reference between a scoped `*.constants.ts` and `*.types.ts`.

Screenshot 2022-12-05 at 6 16 36 PM

Let me know if you have an alternative solution I could apply to this scope.

torgeadelin (approved):
lgtm

PR #339, opened 2022-12-07
Feat: Update All Tokens to Automated Theme Tokens

InteractiveTimmy:
# Description

The scope of the changes in this pull request are to update all tokens used within this project to the new, automated/stable theme tokens from design.

The new theme tokens are defined with the prefix of `--mds-color-theme`, and can be mounted along-side existing tokens with the appended changes to `<ThemeProvider />`.

These changes also modify the css variable validator. From this point on, component tokens will no longer be consumed within components within MRv2, but can be consumed in downstream dependencies.


nataliadelmar (changes requested):

nataliadelmar (approved):
LGTM!!!!!

PR #345, opened 2022-12-16
feat(add-offset-popover): Add offset prop to Popover

nataliadelmar:
# Description

Add offset to the Popover component to allow the consumer to manipulate the position in which the Popover appears regarding its trigger.

Tippy includes this property (See https://atomiks.github.io/tippyjs/v6/all-props/#offset), which is why we needed to add it to our implementation and pass it through.

This work is done on behalf of [SPARK-382377](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-382377), in order to be able to trigger a Popover from the scheduled meeting bricklet that appears in the center of the bricklet rather than outside of it, as the spec states. 

image

# Links

[SPARK-382377](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-382377)


src/components/Popover/Popover.tsx
```diff
@@ -78,6 +79,16 @@ const Popover: FC<Props> = (props: Props) => {
...
 
+  const distance = offset[0] ? offset[0] : DEFAULTS.OFFEST_DISTANCE;
+
+  let skidding: number | undefined;
+
+  if (offset[1]) {
+    skidding = showArrow ? ARROW_HEIGHT + offset[1] : offset[1];
+  } else {
+    skidding = showArrow ? ARROW_HEIGHT + DEFAULTS.OFFEST_SKIDDING : DEFAULTS.OFFEST_SKIDDING;
+  }
```
torgeadelin, line 90: can you add this to the utils file or create one if doesn't exist?
Easier to test.

src/components/MenuTrigger/MenuTrigger.unit.test.tsx.snap
```diff
@@ -15853,8 +15853,8 @@ exports[`<MenuTrigger /> - Enzyme snapshot should match snapshot with variant 1`
           interactive={true}
           offset={
             Array [
-              0,
               5,
+              0,
```
torgeadelin, line 15857: why has this order changed?

torgeadelin (changes requested):
looking good, small adjustments

src/components/Popover/Popover.tsx
```diff
@@ -141,7 +152,7 @@ const Popover: FC<Props> = (props: Props) => {
       delay={delay}
       plugins={addTippyPlugins(hideOnEsc, addBackdrop)}
       // add arrow height to default offset if arrow is shown:
-      offset={[0, showArrow ? ARROW_HEIGHT + OFFSET : OFFSET]}
+      offset={[distance, skidding]}
```
maxinteger, line 155: this should be 
```suggestion
      offset={[skidding, distance]}
```
see in the docs: https://atomiks.github.io/tippyjs/v6/all-props/#offset

src/components/Popover/Popover.types.ts
```diff
@@ -157,4 +157,10 @@ export interface Props extends PopoverCommonStyleProps, Partial<LifecycleHooks>
    * @default `absolute`
    */
   strategy?: PositioningStrategy;
+
+  /**
+   * Array with two number values: the distance (in px) from the reference and the skidding (in px) along the reference.
+   * @default 'undefined'
+   */
+  offset?: Array<number | undefined>;
```
maxinteger, line 165: I think the tippy API is not very nice here and we should not copy it.

Instead create 2 optional props like `offsetDistance?: number` and `offsetSkid?: number` or with similar name.

maxinteger (changes requested):

src/components/Popover/Popover.tsx
nataliadelmar: Leaving this logic in the LazyTippy's offset prop as it got reduced by separating distance from skidding
Is that oki? :)

src/components/Popover/Popover.constants.ts
```diff
@@ -20,6 +20,8 @@ const DEFAULTS = {
   VARIANT: 'small',
   TRIGGER: 'click',
   PLACEMENT: PLACEMENTS.AUTO as string,
+  OFFEST_DISTANCE: 5,
```
cscheffauer, line 23: typo (and line below)

src/components/MenuTrigger/MenuTrigger.unit.test.tsx.snap
cscheffauer: it appears that there is a change on the offset prop DEFAULTS, probably because of that.
@nataliadelmar can u pls check this again? like there is something different now in terms of order, pls check if we don't break anything existing.

cscheffauer (changes requested):

torgeadelin (approved):
lgtm

src/components/MenuTrigger/MenuTrigger.unit.test.tsx.snap
nataliadelmar: oh yeah; it was due to the incorrect swapping of properties in offest; it is now back to normal :)

maxinteger (approved):
Nice one

cscheffauer (approved):

PR #347, opened 2022-12-20
feat(popover): add optional prop to remove backdrop

jor-row:
# Description

This PR is to add an additional optional prop to Popover to remove the backdrop in cases where clicking outside the popover is not an intended way of closing it (e.g. Popovers with ```trigger='manual'``` that are hidden programatically without user intervention). 

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-392243


maxinteger (approved):

src/components/Popover/Popover.utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { Plugin } from 'tippy.js';
+import { addBackdropPlugin, hideOnEscPlugin } from './tippyPlugins';
+
+export const addTippyPlugins = (hideOnEsc: boolean, addBackdrop: boolean): Array<Plugin> => {
+  if (hideOnEsc && addBackdrop) {
```
cscheffauer, line 5: this can be simplified by doing something like (will also help for later if there are more plugins):
```
const plugins = [];
if(hideOnEsc) {
  plugins.push(hideOnEscPlugin);
}
if(addBackdrop) {
  plugins.push(addBackdropPlugin);
}
return plugins;
```

src/components/Popover/Popover.types.ts
```diff
@@ -72,6 +72,13 @@ export type PopoverCommonStyleProps = {
    * @default true
    */
   hideOnEsc?: boolean;
+
+  /**
+   * Whether the popover should hide on clicking outside the popover
+   *
+   * @default true
+   */
+  addBackdrop?: boolean;
```
cscheffauer, line 81: the comment in your JSDoc (whether popover should hide on clicking outside) doesn't really match `addBackdrop`

if consumers are just reading the prop name, they could also mean that the prop allows u to just add a backdrop (without the click outside user behaviour).

maybe we can name this to what it does or adjust the comment to what it really does?

cscheffauer (changes requested):

PR #349, opened 2022-12-29
fix(buttongroup): added role prop for accessibility

deepak-rohilla:
# Description

PR to add role prop for ButtonGroup and ButtonSimple components for accessibility fixes related to [Jira](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-367395). 
Right now the tablist and tab roles are not getting reflected due to which screen readers cannot correctly work on the Space List Header buttons (All, Direct and Spaces).

### DIFF
--- a/src/components/ButtonSimple/ButtonSimple.tsx
+++ b/src/components/ButtonSimple/ButtonSimple.tsx
@@ -15,7 +15,7 @@ import './ButtonSimple.style.scss';
  */
 
 const ButtonSimple = forwardRef((props: Props, providedRef: RefObject<HTMLButtonElement>) => {
-  const { children, className, isDisabled, id, style, title, useNativeKeyDown } = props;
+  const { children, className, isDisabled, id, style, title, useNativeKeyDown, role } = props;
   const internalRef = useRef();
   const ref = providedRef || internalRef;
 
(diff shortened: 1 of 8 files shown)


src/components/ButtonSimple/ButtonSimple.tsx
```diff
@@ -34,6 +34,7 @@ const ButtonSimple = forwardRef((props: Props, providedRef: RefObject<HTMLButton
         {...hoverProps}
         // override of onKeyDown to ensure the standard html button behavior (on enter => onClick) happens
         onKeyDown={useNativeKeyDown ? props.onKeyDown : buttonProps.onKeyDown}
+        role={role}
```
cscheffauer, line 37: I think this is not necessary, since the `props` are going through the useButton and therefore the `buttonProps` should contain the role, if it is passed in.
Pls doublecheck, thanks!

Note: if that works, u don't need to destructure `role` in line 18 i think

src/components/ButtonSimple/ButtonSimple.types.ts
```diff
@@ -34,4 +34,9 @@ export interface Props extends AriaButtonProps, HoverProps {
    * This is necessary since `react-aria` supress that behaviour by design.
    */
   useNativeKeyDown?: boolean;
+
+  /**
+   * Role for adding accessibility
+   */
+  role?: string;
```
cscheffauer, line 41: AriaButtonProps (line 5) should contain the `role` prop i think

cscheffauer (changes requested):

src/components/ButtonSimple/ButtonSimple.tsx
cscheffauer: checked with @deepak-rohilla - role is not passed through, so it has to stay here as it is.

src/components/ButtonSimple/ButtonSimple.types.ts
cscheffauer: checked with @deepak-rohilla - role is not part of AriaButtonProps, so it has to stay here as it is.

cscheffauer (approved):
lgtm! 💯

PR #350, opened 2023-01-19
SPARK-393277: Input hover color broken

nataliadelmar:
# Description

It looks like a recent change to momentum (looks like https://github.com/momentum-design/momentum-react-v2/pull/339/) has left the hover color with an unset variable (--mds-color-theme-background-solid-primary-normal-hovered). This means that in dark mode it falls back to the light theme color.

# Links

[SPARK-393277](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-393277)


chrisadubois (approved):
there were a few other missing css values that were similar to this, did you happen to find them?

maxinteger (approved):

PR #351, opened 2023-01-23
feat(color-bug-fixes): fix missing mds color tokens

nataliadelmar:
# Description

Now that we have switched the source of tokens to be our new MonoRepo --mds tokens, I wanted to make sure that all --mds tokens exist in the node-module's momentum-design dist. This is to avoid bugs like the one encountered in the input legacy component, that used a non-existent -mds color.

# Links

New source of tokens: https://github.com/momentum-design/momentum-design/tree/main/packages/%40momentum-design/tokens/src/stable/theme


InteractiveTimmy (approved):
LGTM, good iterative change.

PR #352, opened 2023-01-24
feat(listitem): navigation with child elements

prsaroha:
# Description
Change keyboard navigation of ListItems with interactive children to use left/right arrow keys. Since there's no guidelines on interactive elements within interactive list items (discouraged actually) other similar structures have been used for reference, such as Grid.

As per new accessibility requirements, if a list item has interactive children then pressing left/right arrow keys should be how they are navigated, instead of the current behavior which utilizes tab. Up/down arrow keys will still navigate from one list item to other, including when pressed on a child element within a list item. Pressing tab/shift+tab on a list item will move the focus out to the list item, further tab/shift+tab on a list item will move focus out of the list, and when navigating back to the list then the last active list item will receive focus.

I made tabindex of all navigable child elements "-1" so they can receive focus via element.focus() but tab key will never focus on them. An issue with react-aria meant that pressing tab on a child element with "-1" tabindex moved the focus to nearest "0" tabindex element, which was always the parent list item. For this reason, the tabindex of a list item is only changed to "0" when exiting the list either by pressing tab, shift+tab, or enter (e.g. opening a space). Any other navigation action that keeps the focus within the list context changes a "0" tabindex to "-1" on the list item. These include pressing arrow keys on a list item with "0" tabindex to move to a different list item or within the same list item.

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-398376
https://www.w3.org/WAI/ARIA/apg/patterns/listbox/
https://www.w3.org/WAI/ARIA/apg/example-index/grid/LayoutGrids.html

[ListItem Navigation.webm](https://user-images.githubusercontent.com/50709949/216297373-b36de624-bc24-4424-898c-b8a118e1c00c.webm)

### DIFF
--- a/src/components/ListItemBase/ListItemBase.tsx
+++ b/src/components/ListItemBase/ListItemBase.tsx
@@ -11,7 +11,7 @@ import classnames from 'classnames';
 
 import './ListItemBase.style.scss';
 import { ContextMenuState, Props } from './ListItemBase.types';
-import { DEFAULTS, SHAPES, SIZES, STYLE } from './ListItemBase.constants';
+import { DEFAULTS, KEYS, SHAPES, SIZES, STYLE } from './ListItemBase.constants';
 import ListItemBaseSection from '../ListItemBaseSection';
 import { verifyTypes } from '../../helpers/verifyTypes';
 import FocusRing from '../FocusRing';
@@ -48,6 +48,7 @@ const ListItemBase = (props: Props, providedRef: RefObject<HTMLLIElement>) => {
 
   const internalRef = useRef<HTMLLIElement>();
   const ref = providedRef || internalRef;
+  const navigableChildren = useRef<Element[]>();
 
   if (shape === SHAPES.isPilled && (size === SIZES[40] || size === SIZES[70])) {
     console.warn(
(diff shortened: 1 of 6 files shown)


src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -92,9 +93,54 @@ const ListItemBase = (props: Props, providedRef: RefObject<HTMLLIElement>) => {
...
+          if (index > -1) {
+            if (index > 0) {
+              newTarget = navigableChildren.current[index - 1];
+            } else {
+              newTarget = navigableChildren.current[navigableChildren.current.length - 1];
+            }
+          } else {
+            newTarget = navigableChildren.current[navigableChildren.current.length - 1];
+          }
+          event.preventDefault();
```
torgeadelin, line 139: Could we please extract these 2 out of this file if possible?

torgeadelin (commented):
looks great otherwise.

src/components/ListItemBase/ListItemBase.tsx
prsaroha: moved this to utils and added tests

torgeadelin (approved):
LGTM

PR #354, opened 2023-01-26
feat(tokens-checker): Complete automatic tokens checker

nataliadelmar:
# Description

As discussed offline, now that tokens have a new source (momentum-design), we want to implement a token checker that:
1. Make sure anything inside a var(*) is a valid token
2. Make sure anything inside a var(*) comes only from our new source momentum-design if it is a file from /src (non-legacy file)
3. Make sure all --mds tokens exist in momentum-design package
4. Make sure all --mdl tokens exist in src/examples/ThemeSelect/themes.js. I changes tokens inside themes.js to have an --mdl prefix instead of a --md prefix to avoid confusion. The 'l' comes from **l**egacy


src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+  lines.forEach((currentLine, index) => {
+    const matcher = /var[(](.*?)[)]/g;
+    const matches = currentLine.matchAll(matcher);
+
+    for (const match of matches) {
+      let tokens = match[0].replace(/rgba[(](.*?)[)]/g, '**').slice(4, -1);
+      tokens.split(', ').forEach((token) => {
+        if (
+          !variableTokens.has(token) &&
+          !token.includes('#') &&
```
chrisadubois, line 101: config powered as above

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+      }
+    }
+  });
+};
+
+const checkVarTokens = function (source) {
+  const lines = source.split('\n');
+
+  const variableTokens = new Set([
+    '--mds-color-theme-text-${level}-normal',
```
chrisadubois, line 88: love this with the interpolation. IMO we want to tokenize these values in the config and allow them to be parameterized

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+    '--mds-color-theme-text-team${teamColorForToken}-normal',
+  ]);
+
+  lines.forEach((currentLine, index) => {
+    const matcher = /var[(](.*?)[)]/g;
+    const matches = currentLine.matchAll(matcher);
+
+    for (const match of matches) {
+      let tokens = match[0].replace(/rgba[(](.*?)[)]/g, '**').slice(4, -1);
+      tokens.split(', ').forEach((token) => {
```
chrisadubois, line 98: constantize the delimiter? why are we re-splitting something that should have already been split? perhaps it would be better to store them in a map as opposed to a string that we have to reprocess?

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+        if (
+          !variableTokens.has(token) &&
+          !token.includes('#') &&
+          !token.includes('*') &&
+          !token.includes('inherit')
```
chrisadubois, line 103: where do the list of strings actually come from? is there an MDN source? are we liable to have new features come out in CSS that we will have to account for here?

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+        ) {
+          try {
+            isVarTokenValid(token);
+          } catch (error) {
+            throw `line ${index + 1}: ${error}`;
```
chrisadubois, line 108: repeated code

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+const checkAllSrcFiles = function () {
+  isSrc = true;
   glob('src/**/*.*', {}, (_er, files) => {
     const errors = [];
     files.forEach((filePath) => {
       try {
-        if (filePath != 'src/helpers/cssTokenCheck.js') {
-          readAndCheckCSS(filePath);
+        if (
+          filePath != 'src/helpers/cssTokenCheck.js' &&
```
chrisadubois, line 130: there is something definitely funky here. Is this really necessary ? IMO this whole check wouldn't be a problem if we ran this from a CLI and the process was running in a different context

src/helpers/cssTokenCheck.js
```diff
@@ -76,19 +176,27 @@ const checkAllLegacyFiles = function () {
...
-    readAndCheckCSS(file);
-  } else {
-    checkAllFiles();
-    checkAllLegacyFiles();
-  }
+  getMDLTokens().then((resultMdlTokens) => {
+    mdlTokens = resultMdlTokens;
+
+    if (file) {
+      isSrc = file.includes('src/') ? true : false;
```
chrisadubois, line 193: src should not be hard coded, we should pass this in

src/helpers/cssTokenCheck.js
```diff
@@ -76,19 +176,27 @@ const checkAllLegacyFiles = function () {
...
+      readAndCheckTokens(file);
+      isSrc = false;
+    } else {
+      checkAllSrcFiles();
```
chrisadubois, line 197: we should offer a plugin model, right? that would allow somebody to define their own function/glob and this could just be an array, so that we don't need to call multiple functions but instead loop through a defined list, better follows the open/closed principle

src/helpers/getTokenCssVars.js
```diff
@@ -21,4 +21,23 @@ const getTokenCssVars = function () {
     .then((jsonString) => JSON.parse(jsonString));
 };
 
-module.exports = getTokenCssVars;
+const getMDLTokens = function () {
```
chrisadubois, line 24: should take a parameter of the themes file it's looking at

src/helpers/getTokenCssVars.js
```diff
@@ -21,4 +21,23 @@ const getTokenCssVars = function () {
...
+  return fs
+    .readFile(path.join(__dirname, '../../src/examples/ThemeSelect/themes.js'))
+    .then((buffer) => buffer.toString('utf-8'))
+    .then((data) =>
+      data
+        .split('\n')
+        .filter((line) => line.includes('--'))
+        .map((line) =>
+          line.replace("'", '"').replace("': '", '": "').replace("',", '",').replace("'", '"')
```
chrisadubois, line 33: i think this line needs comments

src/helpers/getTokenCssVars.js
```diff
@@ -21,4 +21,23 @@ const getTokenCssVars = function () {
...
 
-module.exports = getTokenCssVars;
+const getMDLTokens = function () {
+  return fs
+    .readFile(path.join(__dirname, '../../src/examples/ThemeSelect/themes.js'))
+    .then((buffer) => buffer.toString('utf-8'))
+    .then((data) =>
+      data
+        .split('\n')
+        .filter((line) => line.includes('--'))
```
chrisadubois, line 31: is there any other way of detecting if a line is a css var? wouldn't it be .includes(var) && inclues(--)

src/helpers/getTokenCssVars.js
```diff
@@ -21,4 +21,23 @@ const getTokenCssVars = function () {
...
+        )
+        .join('\n')
+    )
+    .then((formattedLines) => {
+      return ['{', formattedLines, '}'].join('\n');
+    })
+    .then((jsonString) => JSON.parse(jsonString));
+};
+
+module.exports = { getMDSTokens, getMDLTokens };
```
chrisadubois, line 43: Is there an output mode/dry-run mode?

src/helpers/cssTokenCheck.js
```diff
@@ -1,20 +1,44 @@
 const fs = require('fs'); // eslint-disable-line
 const glob = require('glob'); // eslint-disable-line
-const getTokenCssVars = require('./getTokenCssVars'); // eslint-disable-line
+const { getMDSTokens, getMDLTokens } = require('./getTokenCssVars'); // eslint-disable-line
```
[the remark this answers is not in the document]
nataliadelmar: You mean the csstokensCheck.js file? Yes.... butttt there are lots of things in this file that are really adhoc to MRV2 repo. Like, lots of exceptions were inspired in all the legacy spaghetti code that we do not want to touch too much.

src/helpers/cssTokenCheck.js
```diff
@@ -1,20 +1,44 @@
...
 
-let colors;
+let mdlTokens;
+let mdsTokens;
+let isSrc;
 
-const getColor = function (cssVar) {
+const isMDSTokenValid = function (token) {
+  if (!mdsTokens[token]) {
+    throw `MDS token not found for ${token}`;
```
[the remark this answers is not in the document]
nataliadelmar: Oh yes! woops; corrected :)

src/helpers/cssTokenCheck.js
```diff
@@ -1,20 +1,44 @@
...
+    token.match('transparent') ||
+    token.match('--null') ||
+    token.match('$')
+  ) {
     return;
   }
-  if (!colors[cssVar]) {
-    throw `var not found for ${cssVar}`;
+  if (!mdsTokens[token] && !mdlTokens[token]) {
+    throw `Var token not found for ${token}`;
```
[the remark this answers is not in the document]
nataliadelmar: It prints the filepath where the incorrect token was and the line where it was found thanks to line :136
```
      } catch (error) {
        errors.push(`In file ${filePath}:\n ${error}\n`);
      }
```

src/helpers/cssTokenCheck.js
```diff
@@ -27,10 +51,10 @@ const checkCSSVars = function (source) {
     const matches = currentLine.matchAll(matcher);
 
     for (const match of matches) {
-      const cssVar = match[0].slice(0, -1);
-      if (!variableTokens.has(cssVar)) {
+      const token = match[0].slice(0, -1);
```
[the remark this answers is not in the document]
nataliadelmar: Good catch; it took me some time to figure those alterations, so I've added the comment of my logic for those matches.

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
-const readAndCheckCSS = function (filePath) {
+const checkMDLTokens = function (source) {
+  const lines = source.split('\n');
+
+  lines.forEach((currentLine, index) => {
+    const matcher = /--mdl(.*?)[,;)]/g;
+    const matches = currentLine.matchAll(matcher);
+
+    for (const match of matches) {
+      const token = match[0].slice(0, -1);
```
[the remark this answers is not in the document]
nataliadelmar: yes :) Ive abstracter the checkAllFiles and the checkTokens functions; lmk what you think

src/helpers/cssTokenCheck.js
```diff
@@ -39,18 +63,74 @@ const checkCSSVars = function (source) {
...
+      try {
+        isMDLTokenValid(token);
+      } catch (error) {
+        throw `line ${index + 1}: ${error}`;
```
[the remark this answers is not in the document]
nataliadelmar: Yes, Im printing the file path too before the error:
```
      } catch (error) {
        errors.push(`In file ${filePath}:\n ${error}\n`);
      }
```

src/helpers/cssTokenCheck.js
nataliadelmar: string.matchAll()[0] gives you a string. So Im decomposing that string so that it gives me all the tokens found there in an array

src/helpers/cssTokenCheck.js
nataliadelmar: We are liable; indeed... I cannot find a public method to get the updated list of css keywords available from https://developer.mozilla.org/en-US/docs/Web/CSS/Reference

Is it good for now?

src/helpers/cssTokenCheck.js
nataliadelmar: I still do not want to check this current file for tokens, and this file was originally located in src/helpers, and I didn't want to move it from there... even if running from cli to check this repo, wouldn't we still have the problem that src/helpers/cssTokenCheck.js will be evaluated?

src/helpers/getTokenCssVars.js
nataliadelmar: Added :)

src/helpers/getTokenCssVars.js
nataliadelmar: -- seems the main way of differentiating whether a line is referring to the token or not. The token source never has var() on it
image
image

src/helpers/getTokenCssVars.js
nataliadelmar: mmm not sure i understand this one

src/helpers/cssTokenCheck.js
```diff
@@ -1,20 +1,44 @@
...
+  if (!mdlTokens[token]) {
+    throw `MDL token not found for ${token}`;
+  }
+};
+
+const isVarTokenValid = function (token) {
   // exclude locally defined css vars
-  if (cssVar.match('--md-globals') || cssVar.match('--local')) {
+  if (
+    token.match('--md-globals') ||
```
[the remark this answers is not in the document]
nataliadelmar: adding them to config/cssTokenCheck.js

PR #357, opened 2023-01-30
feat(reaction): add onComplete

torgeadelin:
# Description

Add onComplete event handler for Reaction component.
Fix loading spinner.
Add error glyph.

# Links

### DIFF
--- a/src/components/Reaction/Reaction.tsx
+++ b/src/components/Reaction/Reaction.tsx
@@ -3,11 +3,11 @@ import classnames from 'classnames';
 import lottie, { AnimationItem } from 'lottie-web';
 import { useDynamicJSONImport } from '../../hooks/useDynamicJSONImport';
 
-import Icon from '../Icon';
 import { DEFAULTS, STYLE } from './Reaction.constants';
 import { Props } from './Reaction.types';
 import './Reaction.style.scss';
 import LoadingSpinner from '../LoadingSpinner';
+import { GLYPH_NOT_FOUND } from '../Icon/Icon.constants';
 
 const Reaction: FC<Props> = (props: Props) => {
   const { autoPlay, className, id, loop, name, size, style, onComplete } = props;
@@ -27,13 +27,17 @@ const Reaction: FC<Props> = (props: Props) => {
       });
       animation.addEventListener('complete', onComplete);
     }
-    () => {
-      animation?.removeEventListener('complete', onComplete);
+    return () => {
+      if (animation) {
+        animation.removeEventListener('complete', onComplete);
+        animation.destroy();
+      }
     };
   }, [svgContainer, animationData]);
 
   if (error) {
-    return <Icon name={'warning'} scale={16} />;
+    console.warn('Reaction load failed:', error);
+    return <div className={STYLE.notFound}>{GLYPH_NOT_FOUND}</div>;
   }
   if (animationData) {
     return (
--- a/src/components/Reaction/Reaction.tsx
+++ b/src/components/Reaction/Reaction.tsx
@@ -40,7 +40,6 @@ const Reaction: FC<Props> = (props: Props) => {
   }, [svgContainer, animationData]);
 
   if (error) {
-    console.warn('Reaction load failed:', error);
     return <div className={STYLE.notFound}>{GLYPH_NOT_FOUND}</div>;
   }
   if (animationData) {
--- a/src/components/Reaction/Reaction.tsx
+++ b/src/components/Reaction/Reaction.tsx
@@ -25,11 +25,15 @@ const Reaction: FC<Props> = (props: Props) => {
         animationData: animationData,
         name,
       });
-      animation.addEventListener('complete', onComplete);
+      if (onComplete) {
+        animation.addEventListener('complete', onComplete);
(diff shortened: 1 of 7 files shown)


maxinteger (changes requested):
Could you please add a unit test

torgeadelin:
Added, sorry forgot to push last commit @maxinteger

src/components/Reaction/Reaction.unit.test.tsx
```diff
@@ -42,9 +51,34 @@ describe('<Reaction/>', () => {
...
+      const container = await mountAndWait(<Reaction name="haha" size={16} />);
+
+      expect(container).toMatchSnapshot();
+    });
   });
 
   describe('attributes', () => {
+    beforeEach(() => {
+      jest.spyOn(jsonImport, 'useDynamicJSONImport').mockReturnValue({ animationData: 'data' });
+      console.warn = jest.fn();
```
nataliadelmar, line 79: none of this should give you an error; Do we still need this console.warn?

src/components/Reaction/Reaction.unit.test.tsx
```diff
@@ -99,4 +133,50 @@ describe('<Reaction/>', () => {
       expect(element.getAttribute('data-size')).toBe(`${size}`);
     });
   });
+
+  describe('actions', () => {
+    beforeEach(() => {
+      jest.spyOn(jsonImport, 'useDynamicJSONImport').mockReturnValue({ animationData: 'data' });
```
nataliadelmar, line 139: why don't you pass this mock to the first test only as you overwrite it in the next 2 tests

src/components/Reaction/Reaction.unit.test.tsx
```diff
@@ -99,4 +133,50 @@ describe('<Reaction/>', () => {
...
+
+    it('should set/remove listeners correctly', async () => {
+      const addEventListener = jest.fn();
+      const removeEventListener = jest.fn();
+      const destroy = jest.fn();
+      const onComplete = jest.fn();
+
+      jest
+        .spyOn(lottie, 'loadAnimation')
+        .mockReturnValue({ addEventListener, removeEventListener, destroy } as never);
```
nataliadelmar, line 151: why not using the AnimationItem type from lottie for this?

src/components/Reaction/Reaction.tsx
```diff
@@ -1,32 +1,43 @@
...
-      lottie.loadAnimation({
+      animation = lottie.loadAnimation({
         container: svgContainer.current, // the dom element that will contain the animation
         renderer: 'svg',
         loop: loop,
         autoplay: autoPlay,
         animationData: animationData,
+        name,
       });
+      animation.addEventListener('complete', onComplete);
```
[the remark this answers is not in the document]
torgeadelin: yes, addressed

src/components/Reaction/Reaction.unit.test.tsx
```diff
@@ -42,9 +51,34 @@ describe('<Reaction/>', () => {
...
     });
+
+    it('should match snapshot with glyph not found if error', async () => {
+      expect.assertions(1);
+
+      jest.spyOn(jsonImport, 'useDynamicJSONImport').mockReturnValue({ error: new Error('error') });
+
+      const container = await mountAndWait(<Reaction name="haha" size={16} />);
+
+      expect(container.find(`.${STYLE.notFound}`).text()).toEqual(GLYPH_NOT_FOUND);
```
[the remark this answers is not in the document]
torgeadelin: we have both snapshots and manually checking the DOM

src/components/Reaction/Reaction.unit.test.tsx
torgeadelin: it does, in the case where I force it to go in error state, it will console.warn `Reaction load failed`

src/components/Reaction/Reaction.unit.test.tsx
torgeadelin: it was already mocked, i didn't realise it. removed

src/components/Reaction/Reaction.unit.test.tsx
torgeadelin: it's not really and AnimationItem because it's missing a lot of other properties.

src/components/Reaction/Reaction.tsx
```diff
@@ -1,32 +1,43 @@
...
+      if (animation) {
+        animation.removeEventListener('complete', onComplete);
+        animation.destroy();
+      }
+    };
   }, [svgContainer, animationData]);
 
   if (error) {
-    return <Icon name={'warning'} scale={16} />;
+    console.warn('Reaction load failed:', error);
```
maxinteger, line 39: MR2 is a library which is consumed by Apps. Tis means the lib should not do unchangeable thing for example pollute the console with warning. Logging should be happening in the App.

Either remove this warning or add an `onLoadFailed` prop, so the app react on the error.

Also much easier to test

maxinteger (changes requested):

maxinteger (approved):

PR #358, opened 2023-01-31
fix(button-simple): added tabindex for accessibility

deepak-rohilla:
# Description

PR to add tabIndex prop to ButtonSimple for accessibility as the react-aria useButton doesn't forwards the received tabIndex prop and always returns undefined for it.

# Links

[*useButton tabIndex issue.*](https://github.com/adobe/react-spectrum/issues/2799)

### DIFF
--- a/src/components/SearchInput/SearchInput.tsx
+++ b/src/components/SearchInput/SearchInput.tsx
@@ -77,7 +77,6 @@ const SearchInput = (props: Props, ref: RefObject<HTMLInputElement>): ReactEleme
           className={STYLE.clear}
           {...clearButtonProps}
           aria-label={clearButtonAriaLabel}
-          tabIndex={-1}
         >
           <Icon weight="bold" scale={ICON_HEIGHT_MAPPING[height]} name="cancel" />
         </ButtonSimple>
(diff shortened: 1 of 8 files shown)


cscheffauer (approved):

torgeadelin:
There is a test failing, otherwise looks ok.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -77,6 +77,7 @@ const SearchInput = (props: Props, ref: RefObject<HTMLInputElement>): ReactEleme
           className={STYLE.clear}
           {...clearButtonProps}
           aria-label={clearButtonAriaLabel}
+          tabIndex={-1}
```
cscheffauer, line 80: this button inside of the searchInput should be focusable, so setting the tabIndex to -1 here i think isn't correct and has an effect on Cantina. 
@torgeadelin can u pls confirm the accessibility behaviour we are expecting from that clear button in searchInputs?

cscheffauer (changes requested):

src/components/SearchInput/SearchInput.tsx
deepak-rohilla: Have updated the tabIndex handling of ButtonSimple and removed the modifications from other files, thanks for the help @cscheffauer.

torgeadelin (approved):
lgtm

cscheffauer (approved):

PR #359, opened 2023-02-01
Feat/reactions/add new reactions

torgeadelin:
# Description
- update to consume latest animations
- add types and constants with all available animations for reactions
- - original (messaging)
- - on video

**I have renamed some constants which might be breaking consumers that are importing them.**

# Links
https://jira-eng-gpk2.cisco.com/jira/secure/RapidBoard.jspa?rapidView=4992&view=detail&selectedIssue=SPARK-400192&sprint=27627&quickFilter=44812


src/components/Reaction/Reaction.style.scss
```diff
@@ -1,18 +1,13 @@
-.md-reaction-wrapper {
-  height: 100%;
+.md-reaction-wrapper,
+.md-reaction-not-found {
+  height: 10rem;
   display: flex;
 
   &[data-size='12'] {
-    width: 0.75rem;
```
torgeadelin, line 6: Forgot to revert this.

src/components/Reaction/Reaction.style.scss
```diff
@@ -1,18 +1,13 @@
...
+    width: 10rem;
   }
 
   &[data-size='16'] {
-    width: 1rem;
```
torgeadelin, line 10: And this.

src/components/Reaction/Reaction.util.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    const result = cartesian(['prayer', 'raise-hand'], ['yellow', 'dark']);
+    expect(result).toEqual([
+      { type: 'prayer', skinTone: 'yellow' },
+      { type: 'prayer', skinTone: 'dark' },
+      { type: 'raise-hand', skinTone: 'yellow' },
+      { type: 'raise-hand', skinTone: 'dark' },
+    ]);
+  });
+
+  it('returns the correct value when second param is empty', () => {
```
cscheffauer, line 14: i would also consider testing if the first param is empty and both params are empty

src/components/Reaction/Reaction.tsx
```diff
@@ -40,13 +40,18 @@ const Reaction: FC<Props> = (props: Props) => {
   }, [svgContainer, animationData]);
 
   if (error) {
-    return <div className={STYLE.notFound}>{GLYPH_NOT_FOUND}</div>;
+    return (
+      <div data-name={name} className={STYLE.notFound}>
```
cscheffauer, line 44: depending on whats the intention of data-name: if data-name should be a helper prop to identify reactions easily in unit tests, this data-name here could actually cause false positives, since it will still say the real data-name of the reaction, even though it wasn't found. 
I think there are probably 2 solutions for here:
1. Change `data-name={name}` to `data-name={"not-found"}` or similar
2. Add a `data-error` attribute with `not-found` or similar to explicitly indicate that this reaction isn't found

cscheffauer (changes requested):

src/components/Reaction/Reaction.tsx
torgeadelin: Oh, that's a very good point!
I've added it because if you load a bunch of icons and some fail, it's difficult to know which ones failed from looking at the DOM. I will replace it with `data-error`.

src/components/Reaction/Reaction.constants.ts
```diff
@@ -14,14 +24,72 @@ const STYLE = {
...
+  wow: 'wow',
+  haha: 'haha',
+  celebrate: 'celebrate',
+  heart: 'heart',
+  fire: 'fire',
+  'speed-up': 'speed-up',
+  'slow-down': 'slow-down',
+};
+
+const ORIGINAL_REACTIONS = {
```
cscheffauer, line 58: can u pls add a TSDoc to this and the on video reactions with an example of how this constant will look like then?
OR a snapshot test of these constants might be valuable as well (making sure the constants stay as they are intended to be)

thanks!

cscheffauer (approved):
small comment - rest looks really good! 💯

PR #365, opened 2023-02-09
fix(toast-notification): accept non string content

pradeep0601:
# Description:
The changes are comprised of update of the below components - 
1. ToastNotification
   - accept both string & non-string content 

# Links

### DIFF
--- a/src/components/AlertBanner/AlertBanner.style.scss
+++ b/src/components/AlertBanner/AlertBanner.style.scss
@@ -1,7 +1,7 @@
 .md-alert-banner-wrapper {
   display: flex;
   border-radius: 0.5rem;
-  margin: 0.5rem 0rem;
+  margin: 0.5rem;
   transition: background-color 0.2s, color 0.2s, border-color 0.2s;
   width: 25rem;
 
--- a/src/components/AlertBanner/AlertBanner.style.scss
+++ b/src/components/AlertBanner/AlertBanner.style.scss
@@ -1,7 +1,7 @@
 .md-alert-banner-wrapper {
   display: flex;
   border-radius: 0.5rem;
-  margin: 0.5rem;
+  margin: 0.5rem 0rem;
   transition: background-color 0.2s, color 0.2s, border-color 0.2s;
   width: 25rem;
 
--- a/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
+++ b/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
@@ -1,4 +1,4 @@
-import React, { CSSProperties, ReactNode } from 'react';
+import React, { CSSProperties, ReactNode, isValidElement } from 'react';
 import { act, render, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
@@ -17,7 +17,7 @@ type PrepareForSnapshotProps = {
   attention?: AttentionType;
   zIndex?: number;
   limit?: number;
-  enterAnimation?: string;
+  enterAnimation?: Parameters<typeof NotificationSystem>[0]['enterAnimation'];
   newestOnTop?: boolean;
   containerClassName?: string;
   bodyClassName?: string;
@@ -39,6 +39,11 @@ describe('<NotificationSystem />', () => {
     attention,
     zIndex,
     limit,
+    enterAnimation,
+    newestOnTop,
+    containerClassName,
+    bodyClassName,
+    toastClassName,
   }: PrepareForSnapshotProps) => {
     const { container } = render(
       <NotificationSystem
@@ -48,6 +53,11 @@ describe('<NotificationSystem />', () => {
         position={position}
         zIndex={zIndex}
         limit={limit}
+        enterAnimation={enterAnimation}
+        newestOnTop={newestOnTop}
+        containerClassName={containerClassName}
+        bodyClassName={bodyClassName}
+        toastClassName={toastClassName}
       />
     );
 
@@ -60,7 +70,12 @@ describe('<NotificationSystem />', () => {
     });
 
     // wait till the toast shows up on the screen:
-    await screen.findByRole('button', { name: textContent });
+    const isReactElement = isValidElement(content);
+    if (isReactElement) {
+      await screen.findByRole('button', { name: textContent });
+    } else {
+      await screen.findByText(content as string);
+    }
     return { container };
   };
 
@@ -199,22 +214,12 @@ describe('<NotificationSystem />', () => {
     });
 
     it('should match snapshot if notification content is a free string', async () => {
-      expect.assertions(3);
-      const { container } = render(<NotificationSystem />);
+      expect.assertions(1);
 
-      const toastId = '12345';
-      act(() => {
-        NotificationSystem.notify(<NotificationTemplate content={textContent} />, {
-          autoClose: false,
-          toastId,
-        });
+      const { container } = await waitForNotificationToAppear({
+        content: textContent,
       });
 
-      // wait till the toast shows up on the screen:
-      const toast = await screen.findByText(textContent);
-
-      expect(toast).toBeVisible();
-      expect(toast).toHaveTextContent(textContent);
       expect(container).toMatchSnapshot();
     });
 
@@ -299,36 +304,6 @@ describe('<NotificationSystem />', () => {
         'md-toast-notification-wrapper'
       );
       expect(notificationContainerToasts.length).toBe(toastLimit);
-    });
-
-    it('should accept a free string as a notification content', async () => {
-      expect.assertions(2);
-
-      const { container } = render(<NotificationSystem />);
-
-      const toastId = '12345';
-      act(() => {
-        NotificationSystem.notify(<NotificationTemplate content={textContent} />, {
-          autoClose: false,
-          toastId,
-        });
-      });
-
-      // wait till the toast shows up on the screen:
-      const toast = await screen.findByText(textContent);
-
-      expect(toast).toBeVisible();
-      expect(container).toHaveTextContent(textContent);
-    });
-
-    it('should accept a react element as a notification content', async () => {
-      expect.assertions(1);
-
-      const { container } = await waitForNotificationToAppear({
-        content,
-      });
-
-      expect(container).toHaveTextContent(textContent);
     });
   });
 
--- a/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
+++ b/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
@@ -147,7 +147,7 @@ describe('<NotificationSystem />', () => {
     expect.assertions(1);
 
     const { container } = await waitForNotificationToAppear({
-      notificationText,
+      content,
       enterAnimation: 'slideInBottom',
     });
 
@@ -158,7 +158,7 @@ describe('<NotificationSystem />', () => {
     expect.assertions(1);
 
     const { container } = await waitForNotificationToAppear({
-      notificationText,
+      content,
       newestOnTop: false,
     });
 
@@ -169,7 +169,7 @@ describe('<NotificationSystem />', () => {
     expect.assertions(1);
 
     const { container } = await waitForNotificationToAppear({
-      notificationText,
+      content,
       toastClassName: 'toast',
     });
 
@@ -180,7 +180,7 @@ describe('<NotificationSystem />', () => {
     expect.assertions(1);
 
     const { container } = await waitForNotificationToAppear({
-      notificationText,
+      content,
       bodyClassName: 'body-toast',
     });
 
@@ -191,7 +191,7 @@ describe('<NotificationSystem />', () => {
     expect.assertions(1);
 
     const { container } = await waitForNotificationToAppear({
-      notificationText,
+      content,
       bodyClassName: 'container',
     });
 
--- a/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
+++ b/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
@@ -1,4 +1,4 @@
-import React, { CSSProperties } from 'react';
+import React, { CSSProperties, ReactNode } from 'react';
 import { act, render, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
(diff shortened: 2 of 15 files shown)


torgeadelin:
Why do you need a callback on click to the text?

pradeep0601:
> Why do you need a callback on click to the text?

@torgeadelin - This is required to deal with a scenario where we want to open the chat panel in case of clicking on the in-meeting-chat notification.

image

chburket (approved):
LGTM

src/components/ToastNotification/ToastNotification.unit.test.tsx
```diff
@@ -204,5 +269,65 @@ describe('<ToastNotification />', () => {
...
+      expect.assertions(1);
+
+      const mockCallback = jest.fn();
+
+      const wrapper = await mountAndWait(
+        <ToastNotification onToastMessageClick={mockCallback} content={exampleContent} />
+      );
+      const component = wrapper.find('div').filter({ className: STYLE.body });
+
+      component.props().onClick({
```
torgeadelin, line 283: Why is this needed?

torgeadelin: I'm referring to the big click event object.

torgeadelin (changes requested):

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -23,15 +50,18 @@ const ToastNotification: FC<Props> = (props: Props) => {
...
-          <Text className={classnames(className, STYLE.text)} type="body-primary">
-            {text}
+        {isString(content) ? (
+          <Text className={classnames(className, STYLE.content)} type="body-primary">
+            {content}
           </Text>
+        ) : (
+          <div className={classnames(className, STYLE.content)}>{content}</div>
         )}
+
```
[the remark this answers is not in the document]
pradeep0601: @torgeadelin - Do you mean instead of wrapping the non-string content, I should convert it to string and wrap it into the `Text` component?

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -1,19 +1,46 @@
...
+  const divProps: HTMLAttributes<HTMLDivElement> = {
+    className: STYLE.body,
+    ...(onToastMessageClick && {
+      onClick: onToastMessageClick,
+      onKeyDown: handleOnKeyDown,
+      role: 'button',
+      tabIndex: 0,
+    }),
+  };
```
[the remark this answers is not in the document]
pradeep0601: There are two issues with using of ButtonSimple component - 

1. Button Simple comes with the default CSS of a button (background, active, focus etc.)- which can be fixed by overriding it.
2. As we already have a button (leadingVisula) as a descendant of it, so it throws a console error -  `validateDOMNesting(...): <button> cannot appear as a descendant of <button>
`

src/components/ToastNotification/ToastNotification.tsx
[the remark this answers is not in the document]
pradeep0601: done

src/components/ToastNotification/ToastNotification.types.ts
```diff
@@ -44,4 +44,9 @@ export interface Props {
    * Callback when user clicks on the close button.
    */
   onClose?: (event: PressEvent) => void;
+
+  /**
+   * Callback when user clicks on the notification.
+   */
+  onToastMessageClick?: (event: React.MouseEvent<HTMLElement>) => void;
```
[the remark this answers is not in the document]
pradeep0601: updated

src/components/ToastNotification/ToastNotification.types.ts
```diff
@@ -12,9 +12,9 @@ export type SupportedButtonGroup = ButtonPillProps | ButtonCircleProps;
 
 export interface Props {
   /**
-   * The text that appears inside the Notification toast.
+   * The free string or component that appears inside the Notification toast.
    */
-  text: string;
+  content: string | ReactElement | ReactElement[];
```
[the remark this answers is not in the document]
pradeep0601: updated, thanks

src/components/ToastNotification/ToastNotification.unit.test.tsx
pradeep0601: The big object has been auto-generated, and without that it complains - `Argument of type '{}' is not assignable to parameter of type 'MouseEvent<{}, MouseEvent>'.
  Type '{}' is missing the following properties from type 'MouseEvent<{}, MouseEvent>': altKey, button, buttons, clientX, and 29 more.`

src/components/AlertBanner/AlertBanner.style.scss
```diff
@@ -1,7 +1,7 @@
 .md-alert-banner-wrapper {
   display: flex;
   border-radius: 0.5rem;
-  margin: 0.5rem;
+  margin: 0.5rem 0rem;
```
cscheffauer, line 4: why was this changed?

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
```diff
@@ -26,11 +26,11 @@ type PrepareForSnapshotProps = {
 
 // pin the toast id to make the snapshots reliable:
 const fixedToastId = '123456';
-const notificationText = 'This is a test';
+const content = 'This is a test';
```
cscheffauer, line 29: i think its worth having a test where we pass in the content as a 'string' and JSX, to ensure both work

src/components/NotificationSystem/NotificationTemplate.tsx
```diff
@@ -9,9 +9,9 @@ import './NotificationTemplate.style.scss';
 
 interface NotificationTemplateProps {
   /**
-   * Text content to be shown in notification
+   * The content to be shown in notification
    */
-  notificationText: string;
+  content: string | ReactElement | ReactElement[];
```
cscheffauer, line 14: I think this could also be typed as `React.ReactNode` instead

src/components/ToastNotification/ToastNotification.stories.tsx
```diff
@@ -147,7 +153,8 @@ Common.parameters = {
       ),
     },
     {
-      text: 'Very long text that can occupy two, three or even more lines. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation.',
+      content:
+        'Very long text that can occupy two, three or even more lines. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation.',
```
cscheffauer, line 157: i don't see an example in storybook where the content is NOT a string - like we should also show this as an option, like how to pass in something like a JSX

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -1,19 +1,44 @@
...
+    content,
+    leadingVisual,
+    buttonGroup,
+    onClose,
+    onToastPress,
+  } = props;
+
+  const handleOnKeyDown = (e: React.KeyboardEvent<HTMLElement>) => {
+    if (e.key === KEYS.ENTER_KEY) {
+      onToastPress(e as unknown as React.MouseEvent<HTMLElement, MouseEvent>);
```
cscheffauer, line 29: whats on toastPress for?

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -1,19 +1,44 @@
...
+    }
+  };
+
+  const divProps: HTMLAttributes<HTMLDivElement> = {
+    className: STYLE.body,
+    ...(onToastPress && {
```
cscheffauer, line 35: like what is the intention of this?
if `onToastPress` is pass in, you want the div to be a button which is clickable?

If yes, some notes on this:
- Its discouraged from web standards to put a `role=button` on a div, instead of using a native button element
- If this is a real use case, where u want a interactable toast, we have to talk and find a proper architecture allowing it.

cscheffauer (changes requested):
lets chat about this

src/components/ToastNotification/ToastNotification.tsx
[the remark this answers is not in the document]
cscheffauer: @torgeadelin we need to discuss this - i would recommend to use a div instead

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
pradeep0601: added unit tests

src/components/ToastNotification/ToastNotification.tsx
pradeep0601: As per our discussion, I removed the click callback.

src/components/AlertBanner/AlertBanner.style.scss
pradeep0601: It was required to remove the extra margin from the persistent chat banner.
Removing it from here and will override it in the cantina.

cscheffauer (changes requested):

src/components/ToastNotification/ToastNotification.tsx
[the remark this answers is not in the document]
cscheffauer: we discussed this and came to the conclusion to not make the whole toast clickable - instead its recommended to pass in a buttonGroup

src/components/ToastNotification/ToastNotification.unit.test.tsx
cscheffauer: as mentioned in comment above, we will not make the toast clickable - therefore this comment can be resolved, since its not applicable anymore

src/components/ToastNotification/ToastNotification.stories.tsx
pradeep0601: Added storybook with an option to select string or non-string content in the example section.

src/components/ToastNotification/ToastNotification.unit.test.tsx
```diff
@@ -63,17 +65,27 @@ describe('<ToastNotification />', () => {
...
 
       expect(container).toMatchSnapshot();
     });
 
-    it('should match snapshot with text', async () => {
+    it('should match snapshot with non string content', async () => {
       expect.assertions(1);
 
-      const text = 'Example text';
+      const content = <>{exampleContent}</>;
```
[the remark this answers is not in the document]
pradeep0601: updated

src/utils/component-conversions.tsx
```diff
@@ -38,7 +38,9 @@ const PrimitiveConverter: FC<PrimativeConverterProps> = (props: PrimativeConvert
   }
 
   return isElement ? (
-    React.cloneElement(children, { ...addedProps })
+    children.type === Fragment && className ?
+     React.cloneElement(<div />, { ...addedProps }, children) :
```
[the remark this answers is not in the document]
pradeep0601: It was needed for the case where PrimitiveConvertor receive a fragment with className (observed in the case of in-meeting chat).
Now, as I'm not using it and handling the case of string and non-string content in the ToastNotification component itself, so reverted the code.

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
```diff
@@ -124,78 +125,106 @@ describe('<NotificationSystem />', () => {
...
-    expect.assertions(1);
+    it('should match snapshot with enter animation', async () => {
+      expect.assertions(1);
 
-    const { container } = await waitForNotificationToAppear({
-      notificationText,
-      limit: 5,
+      const { container } = await waitForNotificationToAppear({
+        content,
+        enterAnimation: 'slideInBottom',
```
cscheffauer, line 140: so I have double checked this and this test is not going to work.
The `enterAnimation` prop is passed to `waitForNotificationToAppear`, but in that function its not passed on to the NotificationSystem itself.
That could have been also catched by looking at the snapshots, which we SHOULD do - otherwise they are useless.

cscheffauer: actually there are other props also not being passed through in the `waitForNotificationToAppear` function - please check that and update the snapshots accordingly.

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
```diff
@@ -124,78 +125,106 @@ describe('<NotificationSystem />', () => {
...
-      notificationText,
-      bodyClassName: 'body-toast',
+      expect(container).toMatchSnapshot();
     });
 
-    expect(container).toMatchSnapshot();
-  });
+    it('should match snapshot if notification content is a free string', async () => {
+      expect.assertions(3);
+      const { container } = render(<NotificationSystem />);
```
cscheffauer, line 203: why can for this whole block here the `waitForNotificationToAppear` util not be used? I would recommend to reuse it as well (u will have to tweak line 63 - if content is a text, just assert for the text to show up - otherwise wait for the button to show up)

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
```diff
@@ -234,43 +263,73 @@ describe('<NotificationSystem />', () => {
...
 
-    act(() => {
-      NotificationSystem.notify(
-        <NotificationTemplate notificationText="additional toast to test the limit works" />,
-        {
-          toastId: '1234589012',
-          notificationSystemId: '1234567',
-        }
-      );
+    it('should accept a free string as a notification content', async () => {
```
cscheffauer, line 304: if you change the test util `waitForNotificationToAppear` as suggested in my comment above, this test and the test below about (passing in a react element) are both not necessary, since thats what the util already tests implicitly.

cscheffauer (changes requested):
just some small things I spotted, which need to be addressed to be merged

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
pradeep0601: updated

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
pradeep0601: done, thanks

src/components/NotificationSystem/NotificationSystem.unit.test.tsx
pradeep0601: done

cscheffauer (approved):

torgeadelin (approved):

PR #369, opened 2023-02-14
feat(reactions): fix sizes, export types, fix autoplay

torgeadelin:
# Description

- export types
- fix and refactor reaction sizes (to reuse icons scales)
- add extra scale options
- fix changing autoplay, loop props not updating the animation
- improve stries

# Links

[*Links to relevent resources.*
](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-404245)


src/components/index.ts
```diff
@@ -42,8 +42,7 @@ export { default as MeetingListItem, MeetingMarker } from './MeetingListItem';
 export { default as TextInput } from './TextInput';
 export { default as SpaceListItem } from './SpaceListItem';
 export { default as List } from './List';
-export { default as Reaction } from './Reaction';
-export { REACTIONS, SKIN_TONES } from './Reaction';
+export * from './Reaction';
```
nataliadelmar, line 45: We do not export constants/types for any other component? if we do, how do we export it here?

nataliadelmar (commented):
LGTM, just one question

src/components/index.ts
torgeadelin: I'm not really sure why we don't normally export all types and constants in here, but I do want to export everything from Reaction/index.tsx.

@InteractiveTimmy ?

InteractiveTimmy: Type interpolation should still occur even if the types aren't exported at the top level. Attempt to build with declarations set to true and mount it within web client. This should propagate the type tree when using a component with `props`.

If we would like users to consume the type tree when creating objects/typed values in their down-stream applications, we will need to elevate all props to the root export for external imports. OR, we could create isolated module references:

```ts
// Isolated module references
import Component, { ComponentProps } from '@momentum-design/momentum-react-v2/Component';

// Root exports
import Component, { ComponentProps } from '@momentum-design/momentum-react-v2';
```

Root exports requires a bunch of index modifications, isolated module trees require `package.json` modifications.

We should discuss this before moving forwards in a meeting.

InteractiveTimmy: >I'm not really sure why we don't normally export all types and constants in here, but I do want to export everything from Reaction/index.tsx. - @torgeadelin 

We should only use explicit exports, so `export * from ./directory` is dangerous in some consuming TS applications, as it introduces more room for error in export overlaps.

Convert:

```ts
export * from './directory';
```

to:

```ts
import Component from './directory';

export type {
  ComponentProps,
  ComponentOtherType,
} from './directory';

export {
  CONSTANTS as COMPONENT_CONSTANTS,
  someUtils as componentSomeUtils,
} from './directory';
```

Or something similar.

If we want to start/stop using wild-card imports/exports, we need to plan for it and discuss.

InteractiveTimmy (approved):
LGTM!

src/components/Icon/Icon.style.scss
```diff
@@ -1,157 +1,200 @@
...
-
-.md-icon-wrapper {
-  align-items: center;
-  display: flex;
-  height: fit-content;
-  justify-content: center;
-  width: fit-content;
-
+.md-icon-auto-scales {
+  &[data-autoscale='true'],
```
maxinteger, line 2: lot of repeat code here, you could use a loop here, for example

```scss
$scale-sizes: ("true": 1rem, "auto": "auto", "inherit": "inherit");

@each $scale, $size in $scale-sizes {
 
  &[data-autoscale=#{$scale}],
  > *[data-autoscale=#{$scale}] {
    width: $size;
    height: $size;
  }
}
```

torgeadelin: We agreed to not use any scss mixins and utils when started this project, so will leave it as is.
I don't like that it repeats so much, but I don't want to introduce them unless we agree as a team.

PR #370, opened 2023-02-17
SPARK-350148: implement shallow disable in ButtonPill

nataliadelmar:
# Description

Button Pill's disable state doesn't allow onPress actions to be passed. Sometimes, the spec asks us to add a popover to a disabled button (to explain to the user, for example, why is the button disabled). For that, we need to implement a button that allows a shallowDisabled prop to be passed to make the button LOOK disabled, but not to make it be in reality (for html, it is still a non-disabled button), so that we can pass a onHover or onPress action. 

# Screenshots

image


InteractiveTimmy:
This change raises some concern around how the button operates with accessibility.

* Does this impact WAI-ARIA spec?
* If a button is `shallow-disabled`, does it impact tab-order and screen readers?
* Could this be done has a hover?

chburket:
> This change raises some concern around how the button operates with accessibility.
> 
> * Does this impact WAI-ARIA spec?
> * If a button is `shallow-disabled`, does it impact tab-order and screen readers?
> * Could this be done as a hover?

I can answer a few:  Yes it impacts tab order and screen readers (that's one of the main use cases).  We currently only plan to use the mouseenter event to trigger the popover.

nataliadelmar:
> This change raises some concern around how the button operates with accessibility.
> 
> * Does this impact WAI-ARIA spec?
> * If a button is `shallow-disabled`, does it impact tab-order and screen readers?
> * Could this be done as a hover?

It can be done also as hover (trigger of popover = 'mouseenter'

Coread (approved):

PR #376, opened 2023-02-27
Aria toolbar component

Coread:
# Description

This implements the aria toolbar pattern.

# Links

https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/

### DIFF
--- a/src/components/AriaToolbar/AriaToolbar.stories.args.ts
+++ b/src/components/AriaToolbar/AriaToolbar.stories.args.ts
@@ -1,0 +1,24 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+
+const ariaToolbarArgTypes = {
+  // NOTE: Below is an example. See [Storybook argTypes documentation]{@link https://storybook.js.org/docs/react/api/argtypes}.
+  children: {
+    description: 'Provides the child nodes for this element.', // NOTE: Description of this prop.
+    control: { type: 'text' }, // NOTE: Control type for this prop.
+    table: {
+      type: {
+        summary: 'ReactNode', // NOTE: Explicit type of this prop.
+      },
+      defaultValue: {
+        summary: 'undefined', // NOTE: Default value for this prop.
+      },
+    },
+  },
+};
+
+export { ariaToolbarArgTypes };
+
+export default {
+  ...commonStyles,
+  ...ariaToolbarArgTypes,
+};
--- a/src/components/AriaToolbar/AriaToolbar.stories.args.ts
+++ b/src/components/AriaToolbar/AriaToolbar.stories.args.ts
@@ -1,16 +1,27 @@
 import { commonStyles } from '../../storybook/helper.stories.argtypes';
 
 const ariaToolbarArgTypes = {
-  // NOTE: Below is an example. See [Storybook argTypes documentation]{@link https://storybook.js.org/docs/react/api/argtypes}.
   children: {
-    description: 'Provides the child nodes for this element.', // NOTE: Description of this prop.
-    control: { type: 'text' }, // NOTE: Control type for this prop.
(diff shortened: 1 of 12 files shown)


src/components/AriaToolbar/AriaToolbar.stories.args.ts
```diff
@@ -0,0 +1,24 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+
+const ariaToolbarArgTypes = {
```
nataliadelmar, line 3: Is this missing?

src/components/AriaToolbar/AriaToolbar.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <ButtonCircle onPress={onPressHandler}>1</ButtonCircle>
+        <ButtonCircle onPress={onPressHandler}>2</ButtonCircle>
+        <ButtonCircle onPress={onPressHandler}>3</ButtonCircle>
+      </AriaToolbar>
+      <input type="text" id="textInput" aria-label="A text input" />
+    </>
+  );
+};
+
+// NOTE: Export stories here. The first export should be `Example`, and the last export should be `Common`.
```
nataliadelmar, line 77: remove comment

src/components/AriaToolbar/AriaToolbar.unit.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+      <AriaToolbar ariaLabel="test">
+        <ButtonSimple onPress={onPress}>test button</ButtonSimple>
+      </AriaToolbar>
+    );
+
+    triggerPress(element.find(ButtonSimple));
+
+    expect(onPress).toBeCalled();
+  });
+});
```
nataliadelmar, line 207: we should add test for the `Arrow Right-Left` and `Arrow Up-Down` actions

Coread: These are tested in the snapshot tests above. The reasoning is that there is no external effect of the arrow keys other than the change in tab index.

src/components/AriaToolbar/AriaToolbar.stories.args.ts
Coread: added these to the story

src/components/AriaToolbar/AriaToolbar.stories.tsx
Coread: removed

src/components/AriaToolbar/AriaToolbar.types.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * The HTML ID of the element that this toolbar relates to
+   */
+  ariaControls?: string;
+
+  /**
+   * aria-label attribute for the toolbar
+   */
+  ariaLabel: string;
```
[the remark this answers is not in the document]
Coread: Yes, a requirement of the pattern is to provide either aria-label or aria-labelledby. Since we don't need the latter at present, I've made ariaLabel required.

nataliadelmar (approved):
Just one small comment/question

src/components/AriaToolbar/AriaToolbar.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <ButtonCircle onPress={onPressHandler}>3</ButtonCircle>
+      </AriaToolbar>
+      <input type="text" id="textInput" aria-label="A text input" />
+    </>
+  );
+};
+
+Horizontal.argTypes = argTypes;
+
+const Vertical = () => {
```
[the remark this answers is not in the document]
Coread: only the first story appears on the storybook on the docs page where the controls are

PR #383, opened 2023-03-03
Popover first focus

Coread:
# Description

This allows specifying the element that a popover should focus on render using a function ref. See storybook for how this works.

Also in this PR are modifications to the AriaToolbar to
1) not render elements like null
2) handle being focused from outside

### DIFF
--- a/src/components/AriaToolbar/AriaToolbar.tsx
+++ b/src/components/AriaToolbar/AriaToolbar.tsx
@@ -1,9 +1,10 @@
-import React, { FC, ReactNode, useRef, useState, useEffect } from 'react';
+import React, { FC, useRef, useState, useEffect } from 'react';
 import classnames from 'classnames';
 
 import { DEFAULTS, STYLE } from './AriaToolbar.constants';
 import { Props } from './AriaToolbar.types';
 import { useKeyboard } from '@react-aria/interactions';
+import { map } from 'lodash';
 
 /**
  * The AriaToolbar component. A style-less component implementing the Aria Toolbar pattern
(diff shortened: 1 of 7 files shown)


src/components/AriaToolbar/AriaToolbar.tsx
```diff
@@ -66,18 +68,25 @@ const AriaToolbar: FC<Props> = (props: Props) => {
...
     >
-      {React.Children.map<ReactNode, ReactNode>(children, (child, index) => {
-        return React.cloneElement(child as React.ReactElement<any>, {
+      {map(validChildren, (child, index) => {
+        return React.cloneElement(child, {
           tabIndex: index === (currentFocus || 0) ? 0 : -1,
           ref: (e) => {
             buttonRefs.current[index] = e;
+            if (child.ref) {
+              child.ref(e);
```
Coread, line 77: both the AriaToolbar and the Popover will need to use refs so this passes the ref on to the original user. This requires the use of function refs.

src/components/AriaToolbar/AriaToolbar.unit.test.tsx
```diff
@@ -96,6 +109,32 @@ describe('<AriaToolbar />', () => {
 
       expect(container).toMatchSnapshot();
     });
+
+    it('should update order if triggered externally', () => {
+      expect.assertions(4);
```
torgeadelin, line 114: it would be really good to have a test with checking the actual DOM, and not exclusively depend on the snapshots

torgeadelin (approved):
just one small non-blocking comment + fix the failing unit tests

src/components/AriaToolbar/AriaToolbar.unit.test.tsx
Coread: I've switched to RTL for this test and now also check the DOM

PR #405, opened 2023-05-31
feat(slider): added new slider component

cscheffauer:
# Description

- Added new slider component (only for horizontal orientation right now since this is what we need atm)
- Added unit tests
- Accessibility comes from React-aria

### DIFF
--- a/src/components/Slider/Slider.hooks.test.ts
+++ b/src/components/Slider/Slider.hooks.test.ts
@@ -1,0 +1,148 @@
+import { useSliderSideEffects, useThumbSideEffects } from './Slider.hooks';
+import { renderHook } from '@testing-library/react-hooks';
+import { useRef } from 'react';
+import { useNumberFormatter, useSlider } from 'react-aria';
+import { useSliderState } from '@react-stately/slider';
+import type { SliderHookArgs, ThumbHookArgs } from './Slider.types';
+
+describe('useSliderSideEffects', () => {
+  let result;
+
+  const render = (args: SliderHookArgs) => {
+    ({ result } = renderHook((props) => useSliderSideEffects(props), {
+      initialProps: args,
+    }));
+  };
+
+  it.each([
+    {
+      value: 1,
+      minValue: 0,
+      maxValue: 10,
+      step: 1,
+      expectedValuePercentage: 10,
+    },
+    {
+      value: 9,
+      minValue: 0,
+      maxValue: 10,
+      step: 1,
+      expectedValuePercentage: 90,
+    },
+    {
+      value: 105,
+      minValue: 100,
+      maxValue: 110,
+      step: 1,
+      expectedValuePercentage: 50,
+    },
+    {
+      value: 11,
+      minValue: 0,
+      maxValue: 10,
+      step: 1,
+      expectedValuePercentage: 100,
+    },
+    {
+      value: 4,
+      minValue: 5,
+      maxValue: 10,
+      step: 1,
+      expectedValuePercentage: 0,
+    },
+  ])(
+    'should return as expected for %o',
+    ({ value, minValue, maxValue, step, expectedValuePercentage }) => {
+      render({
+        value,
+        minValue,
+        maxValue,
+        step,
+        onChange: jest.fn(),
+        isDisabled: false,
+        ariaLabel: 'test label',
+      });
+
+      expect(result.current).toStrictEqual({
+        groupProps: {
+          'aria-label': 'test label',
+          'aria-labelledby': undefined,
+          id: 'test-ID',
+          role: 'group',
+        },
+        trackProps: {
+          onKeyDown: expect.any(Function),
+          onMouseDown: expect.any(Function),
+          onPointerDown: expect.any(Function),
+          onTouchStart: expect.any(Function),
+        },
+        state: expect.any(Object),
+        trackRef: expect.any(Object),
+        trackStyle: {
+          '--local-value': `${expectedValuePercentage}%`,
+        },
+      });
+    }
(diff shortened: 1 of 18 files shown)


src/components/Slider/Slider.hooks.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  let result;
+
+  const render = (args: SliderHookArgs) => {
+    ({ result } = renderHook((props) => useSliderSideEffects(props), {
+      initialProps: args,
+    }));
+  };
+
+  it.each([
+    {
```
jor-row, line 18: question: I'm assuming it's not possible to have a value that isn't reachable with the given step? e.g. What happens if ```value: 3, minValue: 0, maxValue: 10, step: 3```?

src/components/Slider/Slider.types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Custom style for overriding this component's CSS.
+   */
+  style?: CSSProperties;
+
+  /**
+   * Whether Slider is disabled
+   */
+  isDisabled: boolean;
```
jor-row, line 24: issue: I notice there is no styling that changes based on whether this component is disabled or not. Am I just missing it or does that need to be added? I can see that there are tokens and such for that state in the Figma

src/components/Slider/Slider.unit.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+      expect(sliderInput.getAttribute('value')).toBe(defaultProps.value.toString());
+
+      const newValue = 8;
+      rerender(<ControlledComponent newValue={newValue} />);
+
+      const sliderInputAfterChange = screen.getByRole('slider');
+      expect(sliderInputAfterChange.getAttribute('value')).toBe(newValue.toString());
+    });
+
+    it('should be change value as expected if pressing left and down arrow key', async () => {
```
jor-row, line 175: nitpick: typo 
```suggestion
    it('should change value as expected when left or down arrow key is pressed', async () => {
```

src/components/Slider/Slider.unit.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+      await user.keyboard('{ArrowDown}');
+      const newValueAfterDown = newValue - defaultProps.step;
+
+      const sliderInputAfterPressingDown = screen.getByRole('slider');
+      expect(sliderInputAfterPressingDown).toHaveFocus();
+      expect(sliderInputAfterPressingDown.getAttribute('value')).toBe(newValueAfterDown.toString());
+      expect(defaultProps.onChange).toHaveBeenLastCalledWith(newValueAfterDown);
+    });
+
+    it('should be change value as expected if pressing right and up arrow key', async () => {
```
jor-row, line 199: nitpick: typo
```suggestion
    it('should change value as expected when right or up arrow key is pressed', async () => {
```

src/components/Slider/Slider.unit.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+      await user.keyboard('{ArrowUp}');
+      const newValueAfterUp = newValue + defaultProps.step;
+
+      const sliderInputAfterPressingUp = screen.getByRole('slider');
+      expect(sliderInputAfterPressingUp).toHaveFocus();
+      expect(sliderInputAfterPressingUp.getAttribute('value')).toBe(newValueAfterUp.toString());
+      expect(defaultProps.onChange).toHaveBeenLastCalledWith(newValueAfterUp);
+    });
+
+    it('should be stick to minValue even after pressing more often left arrow', async () => {
```
jor-row, line 223: nitpick: typo
```suggestion
    it('should stick to minValue even after pressing left arrow key more', async () => {
```

src/components/Slider/Slider.unit.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+      await user.keyboard('{ArrowLeft}');
+      const newValueAfterLeft = defaultProps.minValue;
+
+      const sliderInputAfterPressingLeft = screen.getByRole('slider');
+      expect(sliderInputAfterPressingLeft).toHaveFocus();
+      expect(sliderInputAfterPressingLeft.getAttribute('value')).toBe(newValueAfterLeft.toString());
+      expect(defaultProps.onChange).toHaveBeenLastCalledWith(newValueAfterLeft);
+    });
+
+    it('should be stick to maxValue even after pressing more often right arrow', async () => {
```
jor-row, line 250: nitpick:  typo
```suggestion
    it('should stick to maxValue even after pressing right arrow key more', async () => {
```

jor-row (changes requested):
A few nitpicks, questions for my benefit and one step of the build seems to be failing

src/components/Slider/Slider.stories.args.ts
```diff
@@ -0,0 +1,83 @@
...
+      type: {
+        summary: 'function',
+      },
+      defaultValue: {
+        summary: 'undefined',
+      },
+    },
+  },
+};
+
```
jor-row, line 77: suggestion: Maybe adding ```isDisabled``` as an arg for the storybook would be nice

src/components/Slider/Slider.hooks.test.ts
cscheffauer: it is possible at the moment to set any value from the top, there is no clamping / fallback handling in place at the moment. its not necessary to have this right now for the feature we need the slider for, so i'll keep this as a tech debt, if thats fine for u!

jor-row (approved):

PR #408, opened 2023-06-15
fix: removed react aria from slider

cscheffauer:
# Description

- Had to remove react-aria from the slider cause of several dependency issue when consuming it.
- Switched the slider to native html element `<input type="range" .../>` and styled it accordingly.


src/components/Slider/Slider.tsx
```diff
@@ -4,36 +4,35 @@ import classnames from 'classnames';
...
+      type="range"
+      value={value}
+      min={minValue}
+      max={maxValue}
+      aria-label={ariaLabel}
+      onChange={handleChange}
+      step={step}
+      className={classnames(className, STYLE.wrapper)}
+      id={id}
+      style={style}
```
nataliadelmar, line 33: does this make any difference if you are setting classname to STYLE.wrapper already?

src/components/Slider/Slider.tsx
```diff
@@ -4,36 +4,35 @@ import classnames from 'classnames';
...
+      disabled={isDisabled}
```
nataliadelmar, line 34: disabled should have a default value of false

src/components/Slider/Slider.tsx
```diff
@@ -4,36 +4,35 @@ import classnames from 'classnames';
...
-    </div>
+    <input
+      ref={inputRef}
+      type="range"
+      value={value}
+      min={minValue}
+      max={maxValue}
+      aria-label={ariaLabel}
+      onChange={handleChange}
+      step={step}
```
nataliadelmar, line 30: step, value, min and max should prob have default values as well

src/components/Slider/Slider.tsx
```diff
@@ -4,36 +4,35 @@ import classnames from 'classnames';
...
-      </div>
-    </div>
+    <input
+      ref={inputRef}
+      type="range"
+      value={value}
+      min={minValue}
+      max={maxValue}
+      aria-label={ariaLabel}
+      onChange={handleChange}
```
nataliadelmar, line 29: types says that onChange is mandatory to be passed. Is it?

src/components/Slider/Slider.stories.tsx
```diff
@@ -41,6 +41,7 @@ Example.args = {
   maxValue: 100,
   ariaLabel: 'slider',
   value: 0,
+  isDisabled: false,
```
nataliadelmar, line 44: so we are assuming that the value will be controlled from outside, is that right? can storybook args pass functions? so, in that input box, can I put say '(value) => {console.log(value)}' and that would be rendered every time handleChange is called?

src/components/Slider/Slider.tsx
cscheffauer: this is the inline style, which can be passed in by the consumer. Different then the STYLE.wrapper, which is a css `class`

src/components/Slider/Slider.stories.tsx
cscheffauer: yes, the slider is supposed to be a controlled component.
Check the stories file, i have done a wrapper component, which handles exactly this.
In storybook itself u will see "onChange" in the actions tab (next to the controls)

nataliadelmar (approved):

PR #421, opened 2023-07-31
feat(button-inverted): delete button solid as a prop

nataliadelmar:
# Description

Button Circle and Button Pill used to have the following contradictive properties:
- ghost
- outline
- solid

It was not clear what each of them, in combination, meant. That gave place to many bugs in MRV2 and in Cantina.
To fix this, we have agreed that:
- ghost: background is transparent
- outline: button has a border/background
- inverted (previously 'solid'): background is white on light mode and black on dark mode.

This results in a combination of styles like this:
image

# Links


src/components/Chip/Chip.stories.tsx
```diff
@@ -52,6 +52,7 @@ Common.parameters = {
       leftIcon: exampleIcon,
       rightIcon: exampleRightIcon,
       outline: true,
+      inverted: true,
```
jor-row, line 55: Screenshot 2023-08-01 at 11 09 43

The red stories with the red outline don't appear in the figma - might be worth just either doing a 'ghost outline color' example instead for the story

src/components/ButtonPill/ButtonPill.stories.tsx
```diff
@@ -226,30 +201,176 @@ Disabled.parameters = {
...
+delete NotGhostOutlineInverted.argTypes.ghost;
+delete NotGhostOutlineInverted.argTypes.inverted;
+
+NotGhostOutlineInverted.args = {
+  outline: true,
+  ghost: false,
+  inverted: true,
+};
+
+NotGhostOutlineInverted.parameters = {
```
jor-row, line 365: suggestion:

for MultiTemplateWithPseudoStates you could make a chance on line 102 to 

```
<div style={{ display: 'grid', gridTemplateColumns: repeat(${variants.length}, 1fr)` }}>
       {items}
</div>
```

which makes the stories a little easier to look at all at once but this also affects other stories so you don't have to change it if you feel like it's not a good change

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -27,10 +27,8 @@ const ButtonCircleToggle = forwardRef((props: Props, providedRef: RefObject<HTML
 
   const state = useToggleState(props);
 
-  if (ghost === false && outline === false) {
-    console.warn(
-      'ButtonCircleToggle is only designed for outline or ghost circle buttons. Outline and ghost properties cannot be false at the same time.'
-    );
+  if (outline === false) {
+    console.warn('ButtonCircleToggle is only designed for ghost circle buttons.');
```
jor-row, line 31: question: this warning only checks for outline but the message says it's only designed for ghost, is that right?

src/components/ButtonCircle/ButtonCircle.tsx
```diff
@@ -15,11 +15,15 @@ const ButtonCircle = forwardRef((props: Props, providedRef: RefObject<HTMLButton
...
     outline,
-    solid,
+    inverted,
     size,
     ...otherProps
   } = props;
 
+  if (ghost && inverted) {
+    console.warn('ButtonCircle: Ghost inverted button circle is not allowed in Momentum.');
+  }
```
jor-row, line 25: suggestion: A wording similar to the toggle warn would be better
or something along the lines of "Momentum React V2 does not support a ghost inverted ButtonCircle"

"not allowed" doesn't sound right

jor-row (changes requested):

src/components/ToastNotification/ToastNotification.stories.tsx
```diff
@@ -118,7 +118,7 @@ ButtonGroup.parameters = {
       buttonGroup: (
         <>
           <ButtonPill size={28}>Button</ButtonPill>
-          <ButtonPill outline size={28}>
+          <ButtonPill outline ghost size={28}>
```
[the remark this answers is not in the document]
nataliadelmar: yes;
outline alone --> change to outline + ghost
solid --> change to inverted

src/components/ButtonPill/ButtonPill.stories.tsx
jor-row: That looks much better I think!

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -27,10 +27,8 @@ const ButtonCircleToggle = forwardRef((props: Props, providedRef: RefObject<HTML
 
   const state = useToggleState(props);
 
-  if (ghost === false && outline === false) {
-    console.warn(
-      'ButtonCircleToggle is only designed for outline or ghost circle buttons. Outline and ghost properties cannot be false at the same time.'
-    );
+  if (ghost === false) {
+    console.warn('MRV2: Momentum does not support non-ghost inverted ButtonCircleToggle.');
```
jor-row, line 31: This mentions ```inverted```, is that right? or does the toggle just not support non-ghost?

src/components/ButtonCircle/ButtonCircle.stories.tsx
```diff
@@ -35,66 +39,155 @@ Example.args = {
...
+NotGhostOutlineInverted.args = {
+  outline: true,
+  ghost: false,
+  inverted: true,
 };
 
-States.argTypes = { ...argTypes };
-delete States.argTypes.disabled;
-delete States.argTypes.ghost;
+NotGhostOutlineInverted.parameters = {
```
jor-row, line 183: Screenshot 2023-08-01 at 17 39 03
question: I don't think this variant is used anywhere in Cantina, so non-blocking, but do you think the hover and active state for this (in ButtonCircle and ButtonPill) should behave similarly to ghost outline not inverted and go all green when white text? So the only difference is the normal/disabled state? or should we keep it like this? 

It's not technically in the Figma so I guess it's just up to us how we think it should work

jor-row (approved):
🔥 Nice work, LGTM! Two non-blocking comments that I'm not going to request changes for, but if you can address them before merging that would be great

PR #423, opened 2023-08-20
feat(fixed-width-button-content): add component

jor-row:
# Description

Adding a component to enable us to fix the width of a button with variable string contents. Example use case: mute/unmute in-meeting buttons

sb1

# Links

[Figma](https://www.figma.com/file/Qd7VK2mGCDXSX6tJ8OXYqr/Archived_Rebrand_DSK?type=design&node-id=540-89132&mode=design&t=C4S1EK1HSMjuJj8Z-0)


nataliadelmar (commented):
I like this implementation. Could we pass it to be part of the ButtonPill and ButtonPillToggle components.

So, adding the stringContentVariations array as a prop to ButtonPill and ButtonPillToggle. It should be undefined by default. But if it is defined, then we should invisibly print all the variations in hidden mode as you did in this implementation. So the consumer would use it like this:
```
<ButtonPill
    scale={20}
    stringContentVariations = ['long', 'longer', 'longest']
>
   <Icon scale={28}/>
   <Text>Hello</Text>
</ButtonPill>
```

ButtonPill will know its own size. So from there, you can deduce the icon scale and text scale the invisible prints should have. 

What do you think?

PR #425, opened 2023-08-23
feat(button-pill-fixed-width-content): add component

jor-row:
# Description

Add component

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-333073


src/components/ButtonPill/ButtonPill.types.ts
```diff
@@ -1,6 +1,11 @@
 import { ReactNode } from 'react';
 import { ButtonSimpleProps } from '../ButtonSimple';
 
+export type ContentVariations = {
+  stringContentVariations: string[];
+  includeIcon: boolean;
```
nataliadelmar, line 6: we should leave includeIcon as optional, so even if the object `contentVariations = {stringContentVariations: ['a', 'b', 'c,]}`, the we don't need to force the consume to specify includeIcon=true/false

the default should be false, as the default of ContentVariations states

src/components/ButtonPill/ButtonPill.tsx
```diff
@@ -17,8 +19,10 @@ const ButtonPill = forwardRef((props: Props, providedRef: RefObject<HTMLButtonEl
     outline,
     size,
     inverted,
+    contentVariations = DEFAULTS.CONTENT_VARIATIONS,
     ...otherProps
   } = props;
+  const { stringContentVariations, includeIcon } = contentVariations;
```
nataliadelmar, line 25: related to my comment in ButtonPill/types, we should leave includeIcon as optional. So, if the user sends a contentVariations object, but without an includeIcon, we can default includeIcon to false.

```
const { stringContentVariations, includeIcon = DEFAULTS.CONTENT_VARIATIONS.includeIcon } = contentVariations;
```

src/components/ButtonPill/ButtonPillFixedWidthContent/ButtonPillFixedWidthContent.style.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  .md-button-pill-fixed-width-content-button-content {
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    align-items: center;
+    flex-wrap: nowrap;
+    gap: 0.375rem;
+
+    &[data-size='40'] {
```
nataliadelmar, line 17: if this string will be inside a ButtonPill that already has said in its style 'as I am X in size, then any text inside me should be of fontSize: Y, height: Zrem, line-height: Wrem...

Is it necessary to repeat this here?

src/components/ButtonPill/ButtonPillFixedWidthContent/ButtonPillFixedWidthContent.tsx
```diff
@@ -0,0 +1,49 @@
...
+            data-hidden={true}
+          >
+            {includeIcon && <Icon name="placeholder" scale={iconScale} />}
+            <Text type={textType}>{string}</Text>
+          </div>
+        ))}
+      <div
+        className={classnames(className, STYLE.buttonContent)}
+        data-size={buttonPillSize}
+        data-hidden={false}
```
nataliadelmar, line 41: do we need to set data-hidden to false here? isn't it implicitly not hidden?

src/components/ButtonPill/ButtonPillFixedWidthContent/ButtonPillFixedWidthContent.types.ts
```diff
@@ -0,0 +1,28 @@
+import { ReactNode } from 'react';
+
+export interface Props {
+  /**
+   * Child components of this ButtonPillFixedWidthContent.
+   */
+  children: ReactNode;
```
nataliadelmar, line 7: ButtonPill children is an optional prop, so we should also set it optional here

src/components/ButtonPill/ButtonPillFixedWidthContent/ButtonPillFixedWidthContent.utils.ts
```diff
@@ -0,0 +1,42 @@
+import { IconScale } from '../../Icon/Icon.types';
+import { FontStyle } from '../../Text/Text.types';
+
+export const getTextTypeFromButtonPillSize = (buttonPillSize: number): FontStyle => {
+  let textType;
+
+  switch (buttonPillSize) {
+    case 20:
```
nataliadelmar, line 8: I don't see a ButtonPill 20 in the Figma but I see it in the ButtonPill css :/ Let's leave it there just in case

src/components/ButtonPill/ButtonPillFixedWidthContent/ButtonPillFixedWidthContent.types.ts
jor-row: This component makes no sense without children. I'm leaving it as is.

PR #427, opened 2023-09-28
feat(overlay-alert): focus lock

Coread:
# Description

- Added focusLockProps to Overlay. If present, the overlay will render inside a FocusLock
- Used these focusLockProps for OverlayAlert. OverlayAlert can still be configured to use different focusLockProps but the default is to returnFocus

Why didn't I just put the FocusLock in OverlayAlert? Because, like with legacy overlay component, overlays naturally want to restrict the focus to what they show in front of them and not behind them. I left this configurable in case there are exceptions to that.

### DIFF
--- a/src/components/Overlay/Overlay.types.ts
+++ b/src/components/Overlay/Overlay.types.ts
@@ -1,4 +1,5 @@
-import { CSSProperties, ReactNode } from 'react';
+import { CSSProperties, ReactNode, ComponentProps } from 'react';
+import FocusLock from 'react-focus-lock';
 
 export type Color = 'primary' | 'secondary';
 
@@ -32,4 +33,9 @@ export interface Props {
    * Custom style for overriding this component's CSS.
    */
   style?: CSSProperties;
+
+  /**
+   * Props to be passed to FocusLock
+   */
+  focusLockProps?: Omit<ComponentProps<typeof FocusLock>, 'children'>;
 }
--- a/src/components/Overlay/Overlay.unit.test.tsx
+++ b/src/components/Overlay/Overlay.unit.test.tsx
@@ -1,5 +1,8 @@
 import React from 'react';
 import { mount } from 'enzyme';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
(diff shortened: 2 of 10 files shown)


src/components/Overlay/Overlay.types.ts
```diff
@@ -32,4 +33,9 @@ export interface Props {
    * Custom style for overriding this component's CSS.
    */
   style?: CSSProperties;
+
+  /**
+   * Props to be passed to FocusLock
+   */
+  focusLockProps?: Omit<ComponentProps<typeof FocusLock>, 'children'>;
```
cscheffauer, line 40: Nice typing! i didn't know about `ComponentProps` :)

src/components/Overlay/Overlay.unit.test.tsx
```diff
@@ -60,6 +60,14 @@ describe('<Overlay />', () => {
...
     });
+
+    it('should match snapshot with focusLockProps', () => {
+      expect.assertions(1);
+
+      const container = mount(<Overlay focusLockProps={{ returnFocus: true }} />);
+
+      expect(container).toMatchSnapshot();
+    });
   });
```
cscheffauer, line 71: This component and also the `OverlayAlert` are both lacking of actionability tests (which could be done with React Testing Library). Would be great to have 2 simple tests - one with focus lock enabled, one disabled.
The enabled case you gonna focus on a element inside of the modal and then press tab to get to a element outside of it (shouldn't be possible). For disabled opposite scenario - you should be able to focus.
Interactions like `Tab` from [user-event](https://testing-library.com/docs/user-event/convenience#tab) could be used.

cscheffauer (changes requested):
missing interaction tests, otherwise looking good!

src/components/Overlay/Overlay.unit.test.tsx
Coread: I trust that FocusLock works as expected. There is little value in testing a component that is not part of this library

cscheffauer (approved):
looking good 💯

PR #428, opened 2023-10-02
fix(overlay-alert): fix spacing and corners of overlay alert

Coread:
# Description

Update the overlay alert to be in line with the latest UX figma

Before
overlay_alert_before

After
alert_example

Figma Design
figma_overlay_alert

Figma storybook example
Screenshot 2023-10-03 104838


nataliadelmar:
Can we also correct that the description (or secondary text) should have the color TextSecondary-normal as stated on the Figma? :)

cscheffauer (approved):

nataliadelmar (approved):

PR #431, opened 2023-10-11
fix(select): fix select a11y issues and more

cscheffauer:
# Description

- Made use of the new Popover for the Select listbox
- Fixed accessibility issues
- Added new scenarios to the common story
- Added new prop `listboxMaxHeight` to override the height of it

### DIFF
--- a/src/components/Select/Select.unit.test.tsx
+++ b/src/components/Select/Select.unit.test.tsx
@@ -4,8 +4,18 @@ import { Item } from '@react-stately/collections';
 import { DIRECTIONS, STYLE } from './Select.constants';
 import { mountAndWait, triggerPress, waitForComponentToPaint } from '../../../test/utils';
 import ListBoxBase from '../ListBoxBase';
+import Popover from '../Popover';
+
+import '@testing-library/jest-dom';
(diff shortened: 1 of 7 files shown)


src/components/Select/Select.unit.test.tsx
jor-row, line 1: Is it possible to add tests that assert what which list item is in focus when the list box is opened? There would be three tests:
1. assert that the first list item is in focus on open if nothing is selected
2. assert that the selected list item is in focus on open
3. assert that if the first item is disabled that the next focusable item is in focus on open

Also, maybe not needed, but a test that shows the expected behaviour if all of the list items are disabled, what should happen there? Is the whole select disabled and can't be opened?

jor-row, line 1: Should we be checking in the tests that the arrow icon is in the correct state? or are the snapshot tests enough for this? (I'm not sure if anyone looks into them if they fail and just overwrite them...)

cscheffauer: i guess snapshots tests are sufficient for the time being

jor-row (approved):

PR #436, opened 2023-10-25
feat: better handling of tippy instances

martroos:
# Description

This change improves tippy instance handling (specifically via the `setInstance` prop) of popovers so that it no longer relies on accessing a DOM node in order to retrieve the current tippy instance. This means that nested popovers now (correctly) get their own instance, and much of the popover internals are simplified with regards to refs.

Note that I haven't yet added any tests for this. Since this is not a new behaviour (rather a refactor of an old one), I expect the existing set of tests should cover the changes. I am, of course, happy to add tests if wanted.

This change also makes the `applyRef` utility function unused, however I have left this in as it may have use in future. Again if this should be removed I am happy to do so.

### DIFF
--- a/src/components/MenuTrigger/MenuTrigger.unit.test.tsx.snap
+++ b/src/components/MenuTrigger/MenuTrigger.unit.test.tsx.snap
@@ -81,6 +81,7 @@ exports[`<MenuTrigger /> - Enzyme snapshot should match snapshot 1`] = `
         }
       }
       render={[Function]}
+      setInstance={[Function]}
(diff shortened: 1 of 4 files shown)


jor-row (approved):
LGTM 👍🏻 

Would have been nice to see a very explicit storybook example where this functionality could be shown. e.g. One trigger with two nested popovers which have different instances. Have buttons to the side which control the popovers (open/close popover # 1/2). But that's is a nice to have and, in the interest of time, I'm not going to block this PR for it.

PR #443, opened 2023-10-30
feat: added listboxWidth to select

cscheffauer:
# Description

- Added `listboxWidth` prop to select to allow for popover `strategy: fixed`. This will fix a bug where when the select gets used in a scroll box, it can't break out of the scroll box.


jor-row (approved):
Looks good, just the pipeline failing to fix before we can merge

PR #445, opened 2023-11-01
fix(list-item-base): make non-interactive list items have tabIndex -1

jor-row:
# Description

Currently if a ListHeader is used outside of a ListNext context then it is included in the tab order of the page despite it not being interactive. This change makes it so that interactive=false implies tabIndex=-1 in ListItemBase

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-471953


maxinteger (approved):
Looks good

PR #449, opened 2023-11-06
fix(card): static as div

InteractiveTimmy:
# Description

The scope of the changes in this pull request are focused on fixing the `static` value to not render a usable button, but instead, render the card as a `<div />`

# Links

* [SPARK-464621](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-464621)

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -29,23 +29,36 @@ const Card: FC<Props> = (props: Props) => {
     ...otherProps
   } = props;
 
+  const wrapperProps = {
+    className: classnames(className, STYLE.wrapper),
+    'data-color': color,
+    'data-height': height,
+    'data-outline': outline,
+    'data-rounding': rounding,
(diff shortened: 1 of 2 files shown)


stevem63 (approved):

src/components/Card/Card.tsx
```diff
@@ -29,23 +29,36 @@ const Card: FC<Props> = (props: Props) => {
...
+    <>
+      {
+        isStatic ?
+          <div {...wrapperProps} {...otherProps}>
+            <CardStatus color={statusColor} striped={statusStriped} />
+            {children}
+          </div> :
+          <ButtonSimple
+            {...wrapperProps}
+            isDisabled={isStatic || isDisabled}
```
chrisadubois, line 54: isStatic will never be true here so why is it included in the conditional?

chrisadubois (approved):

PR #451, opened 2024-01-10
feat(text-input): add max len property

nataliadelmar:
# Description

To address, [SPARK-488197](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-488197), I added a inputMaxLen optional property to TextInput to allow consume to specify tha max amount of chars the UI should allow the user to input.


chrisadubois (approved):
🙌

PR #452, opened 2024-01-24
fix(csp): use lottie-light to avoid eval()

robstax:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description
lottie uses `eval()` which is not considered secure and will cause issues with CSP. it uses this to evaluate expressions on animations. i don't believe we use any of that functionality and lottie-light does not use eval()

# Links
https://github.com/airbnb/lottie-web/issues/2173

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
+    "source.fixAll.eslint": "explicit"
   }
 }
--- a/src/components/Reaction/Reaction.tsx
+++ b/src/components/Reaction/Reaction.tsx
@@ -1,6 +1,6 @@
 import React, { FC, useRef, useEffect } from 'react';
 import classnames from 'classnames';
-import lottie, { AnimationItem } from 'lottie-web';
+import lottie, { AnimationItem } from 'lottie-web/build/player/lottie_light';
 import { useDynamicJSONImport } from '../../hooks/useDynamicJSONImport';
 
 import { DEFAULTS, STYLE } from './Reaction.constants';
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
+    "source.fixAll.eslint": "explicit"
```
torgeadelin, line 3: this has been haunting me jesus

src/components/Reaction/Reaction.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { FC, useRef, useEffect } from 'react';
 import classnames from 'classnames';
-import lottie, { AnimationItem } from 'lottie-web';
+import lottie, { AnimationItem } from 'lottie-web/build/player/lottie_light';
```
torgeadelin, line 3: not a fan of importing things like this, is it not distributed in any other way?

torgeadelin (commented):
lgtm, just some small comment on the import. will need to test locally and then can approve

src/components/Reaction/Reaction.tsx
robstax: i agree. i could not find any documentation on how else to do this. most GH issues were doing it this way or through webpack like `'lottie-web': path.resolve('node_modules/lottie-web/build/player/lottie_light.js'),`

torgeadelin: if there's no official way it's fine..

InteractiveTimmy: After inspecting the package definition, it appears that this is the best approach to using `lottie_light`, as it is not defined in the entry point or from the package.json.

torgeadelin (approved):

PR #456, opened 2024-03-07
Fix(Toast): Added aria-live prop to Toast

Ashwin7mak:
# Description
I added a new prop aria-live to the Toast component to allow the screen reader to announce the Toast at regular intervals.

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505652


src/components/Toast/Toast.types.ts
```diff
@@ -8,6 +8,12 @@ export type SupportedChildren = ToastContentProps | ToastDetailsProps;
 export type SupportedControlButtons = ButtonControlProps;
 
 export interface Props {
+
+  /**
+   * Aria-live prop as a string. This will be used by a screen reader for live changes.
+   */
+  ariaLive?: "off" | "polite" | "assertive";
```
InteractiveTimmy, line 15: This works great! note that you can also pull these props from `react` as:

```ts
import type { AriaAttributes } from 'react';

let ariaLive: AriaAttributes['aria-live'];
```

This will lock it to the possible values provided by react when populating the `aria-live` attribute on a JSX/TSX element.

**NOTE**: no change required, just a suggestion.

InteractiveTimmy (approved):
LGTM! Great test coverage too!

src/components/Toast/Toast.types.ts
Ashwin7mak: Thats an excellent suggestion :) Never about AriaAttributes from react !!

PR #457, opened 2024-03-26
ci(legacy): remove percy and cypress

cscheffauer:
# Description

- Percy will not be used anymore, therefore removing it. 
- Since Cypress is only used in conjunction with Percy, Cypress will be removed as well

# Links


Coread (approved):
Looks like github actions needs updating as well

PR #467, opened 2024-05-22
Spark 525339 popover

shnaaz:
# Description

- Added role and aria-modal to ModalContainer
- Added aria-labelledby to ModalConatiner if interactive
- aria-haspopup and id to triggerComponent

# Links
[Tech Breakdown](https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/Bucket+Breakdown+-+SPARK-522085+Overlays)
[SPARK-525339](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-525339)

### DIFF
--- a/src/components/ModalContainer/ModalContainer.unit.test.tsx
+++ b/src/components/ModalContainer/ModalContainer.unit.test.tsx
@@ -37,7 +37,7 @@ describe('<ModalContainer />', () => {
     it('should match snapshot with role', () => {
       expect.assertions(1);
 
-      const role = 'dialog';
+      const role = 'tooltip';
 
       const container = mount(<ModalContainer role={role} />);
 
--- a/src/components/ModalContainer/ModalContainer.unit.test.tsx
+++ b/src/components/ModalContainer/ModalContainer.unit.test.tsx
@@ -34,6 +34,26 @@ describe('<ModalContainer />', () => {
       expect(container).toMatchSnapshot();
     });
 
+    it('should match snapshot with role', () => {
+      expect.assertions(1);
+
+      const role = 'dialog';
+
+      const container = mount(<ModalContainer role={role} />);
+
+      expect(container).toMatchSnapshot();
+    });
+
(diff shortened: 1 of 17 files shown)


src/components/ModalContainer/ModalContainer.unit.test.tsx
```diff
@@ -34,6 +34,26 @@ describe('<ModalContainer />', () => {
       expect(container).toMatchSnapshot();
     });
 
+    it('should match snapshot with role', () => {
+      expect.assertions(1);
+
+      const role = 'dialog';
+
+      const container = mount(<ModalContainer role={role} />);
```
jor-row, line 42: This snapshot test should pass something other than role=dialog (as it's the default and will be in every other snapshot anyway)

src/components/Popover/Popover.tsx
```diff
@@ -82,17 +87,35 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
...
+  const handleOnPopoverShow = useCallback(() => {
+    setIsVisible(true);
+  },[]);
+
   useEffect(() => {
     firstFocusElement?.focus();
   }, [firstFocusElement]);
 
+  const triggerComponentCommonProps = {
+    'aria-haspopup': 'dialog',
```
jor-row, line 107: aria-haspopup can have other values like "listbox" or "menu" which I think we might want to use in other components (e.g. SelectNext might need "listbox"). I think this should be a prop that defaults to 'dialog'

shnaaz: As discussed we will check with buttons team if they are handling this.

jor-row: Sounds good. Depending on what they say you might not have to remove this prop if interactive=false but regardless, I think aria-haspopup should be the same value as the modalcontainer role (as per those mdn docs we looked at)

jor-row (approved):
Nice, thanks for making those changes

PR #470, opened 2024-05-23
fix(selectNext): adding aria-label as prop

shivani1211:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description

added aria-label to selectNext component as an optional prop so SR can announce meaningful label for drop down list when focused.  

To give precedence to aria-label in useButton props - deleted aria-labelledby since otherwise SR announces first selection in the dropdown as label since the id for aria-labelledby is linked like that. 

https://react-spectrum.adobe.com/react-aria/useButton.html

added role=combobox so SR can announce dropdowns as comboboxes (expected behavior) and not menu buttons (current behavior). 

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505183
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-502434

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -39,6 +39,7 @@ function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement
     listboxMaxHeight,
     isInForm = DEFAULTS.IS_IN_FORM,
     listboxWidth,
+    ariaLabel,
   } = props;
   const [popoverInstance, setPopoverInstance] = useState<PopoverInstance>();
   const hasBeenOpened = useRef<boolean>(false);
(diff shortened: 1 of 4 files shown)


src/components/Select/Select.tsx
```diff
@@ -48,9 +49,15 @@ function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement
...
-  const { buttonProps } = useButton({ ...triggerProps, isDisabled }, selectRef);
+  const { buttonProps } = useButton(
+    { ...triggerProps, isDisabled, 'aria-label': ariaLabel },
+    selectRef
+  );
   delete buttonProps.color;
   delete buttonProps.onKeyDown;
+  // with aria-labelledby likned to the ID for the <label> below - SR announces frst selection as the label for the dropdown.
+  // aria-labelledby overrides aria-label so if custom aria-label needs to be passed thru cantina aria-labelledby needs to be deleted.
+  delete buttonProps['aria-labelledby'];
```
Ashwin7mak, line 60: We are deleting only if aria-label is supplied, what if we don't have aria-label supplied ? We need to consider this use case as well !!

shivani1211: more research and testing shows I don't need to add ariaLabel as prop - I can just send aria-label from cantina and it gets picked up correctly by SR.

shivani1211: so not deleting aria-labelledby anymore

Ashwin7mak: I am just concerned if in case where ariaLabel is not supplied. I would still add condition if ariaLabel is present, then delete aria-labelledby, else do not.

shivani1211: I deleted the prop ariaLabel and not deleting aria-labelledby anymore. I tested with cantina and I can just just set the aria-label in cantina and no need to send ariaLbel as a prop.

Ashwin7mak (approved):
LGTM !!

PR #473, opened 2024-05-27
feat(avatar): add accessibility-related attributes

Tianhui-Han:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description

- Add props nameLabel, actionLabel, pictureLabel, presenceLabel for accessibility
- Add a series of tests, including the case of various inputs
- Added accessibility related stories

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-527788

### DIFF
--- a/src/components/Avatar/Avatar.stories.args.tsx
+++ b/src/components/Avatar/Avatar.stories.args.tsx
@@ -208,7 +208,20 @@ export default {
   },
   typingLabel: {
     defaultValue: DEFAULTS.LABEL,
-    description: "The label that represents the user is typing,  so that it is used in the aria-label of the container, as part of it. e.g 'is typing' ",
+    description: "The label that represents the user is typing,  so it is used in the aria-label of the container. e.g 'is typing' ",
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  extraLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: "The label that represents additional description, it will be spliced to the end of the container aria-label ",
     control: { type: 'text' },
     table: {
       type: {
--- a/src/components/Avatar/Avatar.stories.args.tsx
+++ b/src/components/Avatar/Avatar.stories.args.tsx
@@ -74,7 +74,7 @@ export default {
   title: {
     defaultValue: undefined,
     description:
-      'Name of person/space. The component will extract initials from this value and display accordingly',
+      "Name of person/space. The component will extract initials from this value and display accordingly, it is used in the aria-label of this component.  e.g 'John Doe'",
     control: { type: 'text', required: true },
     table: {
       type: {
@@ -103,6 +103,20 @@ export default {
   icon: {
     defaultValue: undefined,
     description: 'Name of the icon to be displayed inside the Avatar. Must be a valid icon name.',
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: undefined,
+      },
+    },
+  },
+  iconOnHover: {
+    defaultValue: undefined,
+    description:
+      'Name of the icon to be displayed inside the Avatar when hovered or focused. Must be a valid icon name.',
     control: { type: 'text' },
     table: {
       type: {
@@ -150,6 +164,58 @@ export default {
       },
       defaultValue: {
         summary: false,
+      },
+    },
+  },  
+  actionLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: "The name of the action to be performed by this Avatar onPress, so that it is used in the aria-label of this component. e.g 'Open contact card of ' ",
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  pictureLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: "The name of the graphic of the avatar in case there it does not trigger any action, so that it is used in the aria-label of this component. e.g 'Avatar picture of ' ",
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  presenceLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: "The name of the presence of the avatar's person, so that it is used in the aria-label of this component. e.g 'Busy' ",
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  typingLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: "The label that represents the user is typing,  so that it is used in the aria-label of the container, as part of it. e.g 'is typing' ",
+    control: { type: 'text' },
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
       },
     },
   },
--- a/src/components/Avatar/Avatar.stories.args.tsx
+++ b/src/components/Avatar/Avatar.stories.args.tsx
@@ -221,7 +221,7 @@ export default {
   },
   extraLabel: {
     defaultValue: DEFAULTS.LABEL,
-    description: "The label that represents additional description, it will be spliced to the end of the container aria-label ",
+    description: 'The label that represents additional description, it will be spliced to the end of the container aria-label',
     control: { type: 'text' },
     table: {
       type: {
--- a/src/components/Avatar/Avatar.stories.tsx
+++ b/src/components/Avatar/Avatar.stories.tsx
@@ -14,8 +14,9 @@ import {
 
 import Documentation from './Avatar.documentation.mdx';
 import argTypes from './Avatar.stories.args';
-import { PresenceType } from './Avatar.types';
+import { PresenceType , Props } from './Avatar.types';
 import { AVATAR_COLORS, SIZES } from './Avatar.constants';
+
 
 const DocsPage: FC = () => (
   <>
@@ -43,7 +44,6 @@ const Template: Story<AvatarProps> = (args) => <Avatar title="Cisco Webex" {...a
 
 const MultiTemplate: Story<AvatarProps> = (args: AvatarProps, { parameters }) => {
   const { variants } = parameters;
-
   const items = variants.map((variant, index: number) => (
     <div key={index}>
       <Avatar title="Cisco Webex" {...args} {...variant} />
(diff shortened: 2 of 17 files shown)


src/components/Avatar/Avatar.stories.args.tsx
```diff
@@ -152,5 +152,61 @@ export default {
         summary: false,
       },
     },
+  },  
+  nameLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: 'Alternative text for avatar image.',
```
nataliadelmar, line 158: "The name of the person or space that the Avatar makes reference to, so that it is used in the aria-label of this component"
Let's add this description to the types file too

src/components/Avatar/Avatar.stories.args.tsx
```diff
@@ -152,5 +152,61 @@ export default {
...
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  actionLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: 'The aria-label for avatar container button when onPress is provided. (e.g open contact card of',
```
nataliadelmar, line 172: "The name of the action to be performed by this Avatar, so that it is used in the aria-label of this component. e.g 'Open contact card of ' "
Let's add this description to the types file too

src/components/Avatar/Avatar.stories.args.tsx
```diff
@@ -152,5 +152,61 @@ export default {
...
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  pictureLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: 'The aria-label for avatar container when onPress is not provided. (e.g avatar picture for)',
```
nataliadelmar, line 186: "The name of the graphic of the avatar in case there it does not trigger any action, so that it is used in the aria-label of this component. e.g 'Avatar picture of ' "
Let's add this description to the types file too

src/components/Avatar/Avatar.stories.args.tsx
```diff
@@ -152,5 +152,61 @@ export default {
...
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: DEFAULTS.LABEL,
+      },
+    },
+  },
+  presenceLabel: {
+    defaultValue: DEFAULTS.LABEL,
+    description: 'The aria-label for presence related elements.',
```
nataliadelmar, line 200: "The name of the presence of the avatar's person, so that it is used in the aria-label of this component. e.g 'Busy ' "
Let's add this description to the types file too

src/components/Avatar/Avatar.stories.tsx
```diff
@@ -43,10 +43,9 @@ const Template: Story<AvatarProps> = (args) => <Avatar title="Cisco Webex" {...a
 
 const MultiTemplate: Story<AvatarProps> = (args: AvatarProps, { parameters }) => {
   const { variants } = parameters;
-
   const items = variants.map((variant, index: number) => (
     <div key={index}>
-      <Avatar title="Cisco Webex" {...args} {...variant} />
+      <Avatar title="Cisco Webex" role={variant.onPress?'button':'group'} {...args} {...variant} />
```
nataliadelmar, line 48: can we add this logic to the avatar's component tsx?

src/components/Avatar/Avatar.tsx
```diff
@@ -61,7 +65,7 @@ const Avatar: React.FC<Props> = (props: Props) => {
       {src && (
         <img
           src={src}
-          alt={alt}
+          alt={nameLabel}
```
nataliadelmar, line 68: we should not add labels to the contents of the avatar.
- If sr can access this image, then we should add aria-hidden so that it doesn't

src/components/Avatar/Avatar.tsx
```diff
@@ -72,6 +76,7 @@ const Avatar: React.FC<Props> = (props: Props) => {
         <Icon
           className={classnames(STYLE.wrapperChildren, STYLE.iconWrapper)}
           name={icon}
+          ariaLabel={nameLabel}
```
nataliadelmar, line 79: same with this icon: we should not add labels to the contents of the avatar. 
- If sr can access them, we should add aria-hidden so that it doesn't
- the meaning of this icon is important. so we should add a property of iconAriaLabel, and add that value to our avatar's containerAriaLabel. That way it is read with the Avatar

src/components/Avatar/Avatar.tsx
```diff
@@ -92,14 +97,15 @@ const Avatar: React.FC<Props> = (props: Props) => {
           presenceIcon={presenceIcon}
           isCircularWrapper={isCircularWrapper}
           size={size}
+          ariaLabel={presenceLabel}
```
nataliadelmar, line 100: i think we should not add labels to the contents of the avatar.

- If sr can access them, we should add aria-hidden so that it doesn't
- the meaning of this presence is important and it is conveyed in the avatars aria label

src/components/Avatar/Avatar.unit.test.tsx
```diff
@@ -2,23 +2,108 @@ import Avatar from '.';
...
+  const checkAvatarAccessibility = ({
+    withOnPress,
+    withPresence,
+    isPerson,
+    avatarType,
+  }) => {
+    const onPress = withOnPress
+      ? () => {
+          // eslint-disable-next-line no-console
+          console.log('trigger -----');
```
nataliadelmar, line 32: this could log trigger----- in our console at the time of running the tests. Which we should avoid

src/components/Avatar/Avatar.unit.test.tsx
```diff
@@ -2,23 +2,108 @@ import Avatar from '.';
...
+  
+    expect.assertions(assertCount);
+  
+    const wrapper = mockAvatar.find('.md-avatar-wrapper');
+    expect(wrapper.props().title).toEqual(sampleProps.title);
+  
+    if (withOnPress) {
+      const button = mockAvatar.find('button');
+      expect(button.props()['aria-label']).toEqual(
+        `${withOnPress ? sampleProps.actionLabel : sampleProps.pictureLabel} ${sampleProps.nameLabel} ${withPresence ? sampleProps.presenceLabel : ''}`,
```
nataliadelmar, line 69: instaed of constructing the same logic as we have in our implementation, we should be passing an 'expectedAriaLabel' prop to this check function. We should write down manually on each test what is the aria label we expect will be resulting from each case

src/components/Avatar/Avatar.unit.test.tsx
```diff
@@ -2,23 +2,108 @@ import Avatar from '.';
...
+      );
+    } else {
+      const container = mockAvatar.find('.md-avatar-wrapper');
+      expect(container.props()['aria-label']).toEqual(
+        `${withOnPress ? sampleProps.actionLabel : sampleProps.pictureLabel} ${sampleProps.nameLabel} ${withPresence ? sampleProps.presenceLabel : ''}`,
```
nataliadelmar, line 74: same here: instaed of constructing the same logic as we have in our implementation, we should be passing an 'expectedAriaLabel' prop to this check function. We should write down manually on each test what is the aria label we expect will be resulting from each case

src/components/Avatar/Initials.tsx
```diff
@@ -29,7 +29,7 @@ const Initials: FC<InitialsProps> = (props: InitialsProps) => {
       );
   }
 
-  return <span className={className}>{initials}</span>;
+  return <span className={className} aria-hidden="true">{initials}</span>;
```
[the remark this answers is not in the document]
nataliadelmar: confirmed by the ux team: sr should not read the initials

nataliadelmar (commented):
I see the Avatar, apart from its aria-label, it is described by 3 elements that could be on it: icon, isTyping and presence.
image

The first 2 elements are on top of the avatar, so they should no be accessed by sr to not mess up with the tabbing logic. But the presence is on the right corner of the avatar, so (sorry I asked differntly before) we should see if making it sr-accessible makes sense).

To address this description mess, I suggest:

1. We could open a const called avatarDescriptions
const avatarDescriptions = [];

2. We add the ids of icon, presence and isTyping, if they are truthy. We should generate unique ids to make the avatar descriptions of Bob do not mess up with the avatar descriptions of John, for example.
- if icon then we push 'unique-avatar-icon-id' to avatarDescriptions
- if presence, then we push 'unique-avatar-presence-id' to avatarDescriptions
- if isTyping, then we push 'unique-avatar-typing-id' to avatarDescriptions

3. in the avatar (the div or the button), we add
aria-describedby={onPress ? '' : avatarDescriptions.join(' ')} // so that all the ids end in a string separated by a space aria-describedby='id-1 id-2 id-3'

4. We add props to the icon, istyping and presence
- in the <Icon>, we keep the aria-hidden, we add ariaLabel = {iconAriaLabel} // that is a new prop we would need to add. It needs to come from cantina as it should be translated to the language of the user.
- in the presence, we should probably remove aria-hidden=true (sorry that i suggested this before, I think it makes sense to allow sr to focus on that presence as it doesn't interfere with the avatar's image as the icon and istyping). Then we add to the presence ariaLabel = {presenceLabel}
- in the isTyping span, same as icon: we keep the aria-hidden, and we add ariaLabel = {typingAriaLabel} // new prop too

src/components/Avatar/Avatar.tsx
nataliadelmar:
```
        <img
          src={src}
          alt=""
          onLoad={handleOnLoad}
          onError={handleOnError}
          className={classnames(STYLE.wrapperChildren, { [STYLE.imageHidden]: !imageLoaded })}
        />
```

src/components/Avatar/Avatar.tsx
```diff
@@ -46,12 +48,41 @@ const Avatar: React.FC<Props> = (props: Props) => {
...
+  const containerAriaLabel = rest['aria-label'] || avatarLabels.join(' ');
+  delete rest['aria-label'];
+  
   const content = (
     <div
       className={classnames(STYLE.wrapper, className)}
       data-size={size}
       data-color={color}
+      aria-label={onPress ? '' : containerAriaLabel}
       title={!hideDefaultTooltip ? title : ''}
```
nataliadelmar, line 84: this is interfering for some reason with the screen reader; when I try to go inside the avatar, it goes to the title

image

src/components/Avatar/Avatar.types.ts
```diff
@@ -70,6 +66,37 @@ export interface Props extends Omit<AriaButtonProps, 'type'> {
...
+
+  /**
+   * The name of the graphic of the avatar in case there it does not trigger any action, so that it is used in the aria-label of this component.
+   * e.g 'Avatar picture of '
+   */
+  pictureLabel?: string;
+
+  /**
+   * The name of the person or space that the Avatar makes reference to, so that it is used in the aria-label of this component.
+   * e.g 'Alice's avatar'
```
nataliadelmar, line 83: e.g 'John Doe'

src/components/Avatar/Avatar.types.ts
```diff
@@ -70,6 +66,37 @@ export interface Props extends Omit<AriaButtonProps, 'type'> {
...
+  /**
+   * The name of the presence of the avatar's person, so that it is used in the aria-label of this component.
+   * e.g 'Busy'
+   */
+  presenceLabel?: string;
+
+
+  /**
+   * The label that represents the user is typing,  so that it is used in the aria-label of the container, as the end part of it.
+   * e.g 'isTyping'
```
nataliadelmar, line 96: e.g 'is typing'

src/components/Avatar/Avatar.tsx
```diff
@@ -46,12 +48,46 @@ const Avatar: React.FC<Props> = (props: Props) => {
...
+
+  if(presence && presenceLabel){
+    avatarLabels.push(presenceLabel);
+  }
+
+  if(isTyping && typingLabel){
+    avatarLabels.push(typingLabel);
+  }
+
+  const containerAriaLabel = rest['aria-label'] || avatarLabels.join(',');
```
nataliadelmar, line 80: we should join them with `.join(', ') // comma and space`

nataliadelmar (approved):

nataliadelmar (approved):

PR #474, opened 2024-05-28
fix: popover's esc key behaviour

maxinteger:
# Description

Popover closes when the user presses Esc key. But when there are multiple open popover then Esc closes all of them. After this modification, Esc closes only the lastly opened (most nested) popover and the user must press it again if s/he want to close next level.

issue: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-529166


test/utils.ts
```diff
@@ -56,3 +56,60 @@ export const simulateMouseLeave = (component: ReactWrapper): void => {
...
+ *
+ * When `spyOnEvents` is not specified, it will spy on all events
+ *
+ * Returns with the add/removeEventListener, they are spying only on the specified events
+ * and the eventHandlers object which contains the registered event handlers
+ *
+ * @param target - Event target
+ * @param spyOnEvents - List of event names to spy on
+ */
+export const sypOnEventListener = (target: EventTarget, spyOnEvents?: string[]) => {
```
torgeadelin, line 74: praise: nice util :D

torgeadelin (approved):
I've tested manually, and works :D

src/components/Popover/Popover.stories.tsx
```diff
@@ -181,6 +181,50 @@ MultiplePopovers.args = {
...
+  children: (
+    <p>
+      Popover level 1
+      <div>
+        <Popover
+          trigger="mouseenter"
+          placement={PLACEMENTS.BOTTOM}
+          showArrow
+          interactive
+          triggerComponent={<ButtonSimple>Hover or click me!</ButtonSimple>}
```
cscheffauer, line 202: the way the nested popover is done in the Client is that the Nested Popover is passed in through the triggerComponent, not as a children. Please edit this and test if it works, since its a different way of implementation

cscheffauer (changes requested):

src/components/Popover/Popover.stories.tsx
jor-row: When Laszlo says nested popover he means a popover that contains a tooltip, not a button with a tooltip and a popover on it (that don't open at the same time)

cscheffauer (approved):

PR #475, opened 2024-05-28
do-no-merge feat(search-input): make clearButtonAriaLabel required

nataliadelmar:
BREAKING CHANGE: DO NOT MERGE IF YOU ARE NOT THE AUTHOR

# Description

On the SearchInput component, we need to require the consumer to pass a clearButtonAriaLabel property for when the user has added value to the input. This is an accessibilty requirement, as the clear button will always be a ButtonCircle with an X, so aria-label is needed to describe to the impaired user what would happen if he clicks on that X.

Cantina PR to handle breaking-changes:

### DIFF
--- a/src/components/List/List.stories.tsx
+++ b/src/components/List/List.stories.tsx
@@ -197,7 +197,7 @@ const ListSearchWrapper = () => {
 
   return (
     <>
-      <SearchInput value={query} onChange={setQuery} />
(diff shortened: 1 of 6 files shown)


maxinteger (approved):
Looks good

torgeadelin (approved):

PR #478, opened 2024-05-29
fix(focus-lock): focus lock on popover, modal etc

cscheffauer:
# Description

- Moved Focus Lock into ModalContainer
- Removed logic from MenuTrigger to tab outside of the modal, since this is not applicable anymore according to WCAG
- CSS changes necessary since DOM structure changed due to the focus lock movement
- Adjusted unit tests in all applicable spots


src/components/Popover/Popover.tsx
jor-row, line 1: I'm seeing an interesting change in behaviour that I'm not sure we want. when linking locally with cantina in the call controls on production I can tab through each button as expected, but on local host when tabbing through call controls I have to press tab to dismiss the tooltips (the ones on start/stop video and mute/unmute in "non expanded (icon only) mode"). Also, some tooltips aren't opening at all with keyboard.

Is this due to these changes?

https://github.com/momentum-design/momentum-react-v2/assets/86778628/8df81f97-c0a4-4588-afd4-53a9badb878a

jor-row, line 1: Focus enters popovers as expected

jor-row: I tested with the latest momentum and looks like this behaviour is due to this PR so we should fix that

jor-row (changes requested):

src/components/Popover/Popover.tsx
```diff
@@ -68,10 +68,8 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
   const triggerComponentId = triggerComponent.props?.id || uuidV4();
 
   const modalConditionalProps = {
-    ...(interactive && {'aria-labelledby': triggerComponentId}),
+    ...(interactive && { 'aria-labelledby': triggerComponentId, focusLockProps: { returnFocus: true,  } }),
```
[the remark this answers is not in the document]
cscheffauer: good catch - we haven't discussed this during tech breakdown, going to set it to the same value as `focusBackOnTrigger`

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -92,11 +91,6 @@ const MenuTrigger: FC<Props> = (props: Props) => {
       if (event.key === 'Escape') {
         closeMenuTrigger();
       }
-      // When there are more than one menus inside the menu trigger, we should not close the overlay
-      // according to W-ARIA
-      if (state.isOpen && event.key === 'Tab' && menus.length === 1) {
-        closeMenuTrigger();
-      }
```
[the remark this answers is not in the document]
cscheffauer: yes, clarified in the space

src/components/MenuTrigger/MenuTrigger.unit.test.tsx
```diff
@@ -331,174 +335,149 @@ describe('<MenuTrigger /> - React Testing Library', () => {
...
-          <Menu selectionMode="multiple" key="4" aria-label="Multi Menu">
-            <Item key="asd">Four</Item>
-            <Item key="ff">Five</Item>
-            <Item key="d">Six</Item>
-          </Menu>,
-        ],
-      };
-
-      render(<MenuTrigger {...defaultProps} />);
+    describe('keyboard accessibility', () => {
```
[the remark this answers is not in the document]
cscheffauer: that was ordered that way previously - have cleaned it up now

src/components/Popover/Popover.tsx
cscheffauer: fixed

src/components/ModalContainer/ModalContainer.types.ts
```diff
@@ -1,5 +1,6 @@
-import { CSSProperties, ReactNode } from 'react';
+import { CSSProperties, ReactNode, ComponentProps } from 'react';
 import type { PlacementType } from '../ModalArrow/ModalArrow.types';
+import FocusLock from 'react-focus-lock';
```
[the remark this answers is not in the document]
cscheffauer: yes, there have been issues with it (can't recall the exact issues). We could have also used focus scope, but we would have needed a lot more of work arounds than we do need for the react-focus-lock.

src/components/Popover/Popover.tsx
jor-row, line 1: Linked locally with Cantina and here are some issues I'm seeing:

- OverlayAlert no longer returns focus properly. This works as expected on web-unstable. Examples to test: Profile Popover -> Settings -> Report an issue / Health Checker, Space settings -> View space policy
- In-meeting reaction popover you can no longer press Tab to cycle through all the options, but can cycle through them all with alt-tab
- In-meeting reaction popover, can no longer press esc to close the popover when focused on any of the reaction buttons
- End meeting button, the thing that takes focus when the popover is opened is not the button and cannot be pressed 

Accessibility bug that we thought would be fixed with this but isn't yet
- Video options popover doesn't automatically take focus when opened

jor-row: 4. Looks like this can be solved in Cantina - remove FocusTrappableElement and add autoFocus={false} to the PopoverNext props

jor-row (approved):

PR #480, opened 2024-05-31
feat(lightbox): add theme wrapper

torgeadelin:
# Description
- add theme wrapper for lighbox
- required as part of:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-503208
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-503209
- fixes focus styling for all components inside

### DIFF
--- a/src/legacy/Lightbox/index.js
+++ b/src/legacy/Lightbox/index.js
@@ -4,11 +4,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Modal from 'react-aria-modal';
 import { Spinner, Tooltip, Icon } from '@momentum-ui/react-collaboration';
+import MomentumThemeProvider from '../../components/ThemeProvider';
+
 import { v4 as uuidv4 } from 'uuid';
 
 /**
-* @deprecated - Components in the legacy folder (/src/legacy) are deprecated. Please use a component from the components folder (/src/components) instead. Legacy components may not follow accessibility standards.
-**/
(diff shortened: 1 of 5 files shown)


src/legacy/Lightbox/index.js
```diff
@@ -496,54 +498,59 @@ class Lightbox extends React.Component {
         underlayClass="md-lightbox__container"
         aria-labelledby={this.nameId}
       >
-        <div className="md-lightbox__header" ref={(ref) => (this.lightBox = ref)}>
-          <div className="md-lightbox__header-item--left">
-            <div className="md-lightbox__header-meta">
-              <div className="md-lightbox__header-sharer">{info.sharedBy}</div>
-              <div className="md-lightbox__header-timestamp">{info.sharedOn}</div>
+        {/* adding this so that focus styling applies to components inside the modal. */}
+        <MomentumThemeProvider theme={this.props.theme}>
```
cscheffauer, line 502: I think this is not necessary - in the app the MomentumThemeProvider is mounted on the top and this Lightbox is not in a iFrame, so it should pick up the theme correctly. 
If not, can u please explain why its not picking up the theme?

cscheffauer (changes requested):

src/legacy/Lightbox/index.js
torgeadelin: Lighbox uses an external dependency for the Modal, and it build the dom elements outside of the theme provider. I've attached pic.
image


Legacy Modal does the same thing.

maxinteger (approved):

cscheffauer (approved):

PR #482, opened 2024-06-03
feat(clear-button-index): overwrite react aria excludeFromTabOrder

nataliadelmar:
# Description

We are using react-aria's useSearchField for the GlobalSearchInput component. However, this hook is returning a static true on the clear button excludeFromTabOrder, which makes this button not focusable and violating accessibilty rules.

I am overwriting this prop from react aria to false before passing it to the clear button. 

# Links

[*Links to relevent resources.*](https://github.com/adobe/react-spectrum/blob/02f242f41f0347b3f11e50b019c197127b7d0a08/packages/%40react-aria/searchfield/src/useSearchField.ts#L118)


torgeadelin (approved):
lgtm

PR #484, opened 2024-06-04
fix: ariatoolbar propagate keydown

cscheffauer:
# Description

- Fixed the aria toolbar, by allowing the keyDown event to propagate up (this is for fixing accessibility issues, when using a aria toolbar within a Popover)


jor-row (approved):
Linked to Cantina and works as expected

PR #488, opened 2024-06-05
feat(button): add aria-disabled for shallow disabled

torgeadelin:
# Description
- add aria-disabled for shallow disabled

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-367651

### DIFF
--- a/src/components/ButtonCircle/ButtonCircle.tsx
+++ b/src/components/ButtonCircle/ButtonCircle.tsx
@@ -38,6 +38,7 @@ const ButtonCircle = forwardRef((props: Props, providedRef: RefObject<HTMLButton
       data-inverted={inverted || DEFAULTS.INVERTED}
       data-disabled={disabled || DEFAULTS.DISABLED}
       data-shallow-disabled={shallowDisabled || DEFAULTS.SHALLOW_DISABLED}
+      aria-disabled={shallowDisabled || DEFAULTS.SHALLOW_DISABLED}
       isDisabled={disabled}
       {...otherProps}
     >
--- a/src/components/ButtonCircle/ButtonCircle.unit.test.tsx
+++ b/src/components/ButtonCircle/ButtonCircle.unit.test.tsx
@@ -308,6 +308,19 @@ describe('<ButtonCircle />', () => {
 
       expect(container.find(ButtonSimple).exists()).toBe(true);
     });
+
+    it('should pass shallow disabled prop', () => {
+      expect.assertions(2);
+
+      const shallowDisabled = !DEFAULTS.SHALLOW_DISABLED;
+
+      const element = mount(<ButtonCircle shallowDisabled={shallowDisabled} />)
+        .find(ButtonCircle)
+        .getDOMNode();
+
+      expect(element.getAttribute('data-shallow-disabled')).toBe(`${shallowDisabled}`);
+      expect(element.getAttribute('aria-disabled')).toBe('true');
+    });
   });
 
   describe('actions', () => {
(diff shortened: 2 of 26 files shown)


src/components/ButtonCircle/ButtonCircle.tsx
```diff
@@ -38,6 +38,7 @@ const ButtonCircle = forwardRef((props: Props, providedRef: RefObject<HTMLButton
       data-inverted={inverted || DEFAULTS.INVERTED}
       data-disabled={disabled || DEFAULTS.DISABLED}
       data-shallow-disabled={shallowDisabled || DEFAULTS.SHALLOW_DISABLED}
+      aria-disabled={shallowDisabled || DEFAULTS.SHALLOW_DISABLED}
```
nataliadelmar, line 41: what is prop disabled = true ?

src/components/ButtonSimple/ButtonSimple.tsx
```diff
@@ -27,6 +27,12 @@ const ButtonSimple = forwardRef((props: Props, providedRef: RefObject<HTMLButton
     ? { ...buttonProps, ...hoverProps, role, tabIndex }
     : { ...buttonProps, ...hoverProps, role };
 
+  // props should override buttonProps for aria-disabled because react-aria has it's own logic for aria-disabled.
+  // so if we want to pass it in via props, we need to make sure it's not overridden.
+  if (props['aria-disabled'] === true) {
```
maxinteger, line 32: based on the react-aria code if you set `isDisabled` prop on the button then it will update the `aria-disabled` as well.
https://react-spectrum.adobe.com/react-aria/useButton.html#disabled


Can we use the `isDisabled` prop in cantina instead of fight with react-aria

Probably we can remove the `disabled` prop from the type def to avoid confusion

maxinteger (changes requested):

src/components/ButtonCircle/ButtonCircle.tsx
torgeadelin: What do you mean?

src/components/ButtonSimple/ButtonSimple.tsx
torgeadelin: No because we don't actually want the button to be disabled, we want it to be shallow disabled (semantically disabled)

src/components/ButtonCircle/ButtonCircle.unit.test.tsx
```diff
@@ -308,6 +308,19 @@ describe('<ButtonCircle />', () => {
 
       expect(container.find(ButtonSimple).exists()).toBe(true);
     });
+
+    it('should pass shallow disabled prop', () => {
```
nataliadelmar, line 312: I see a too similar test (the name is prob wrong) that we could reuse called 'should pass sallow disabled prop' a few lines up there (line 220 of this file)

nataliadelmar (approved):

src/components/ButtonCircle/ButtonCircle.tsx
torgeadelin: I don't think so. aria-disabled is just a semantically equivalent of disabled but with the exception that it's still interactive.

If the button is disabled, adding aria-disabled seems paradoxal to me, it's like forcing disabled but then reverting the true disabled to be interactive and just be semantically disabled (for SR).

src/components/ButtonCircle/ButtonCircle.unit.test.tsx
torgeadelin: good eye, have fixed this

maxinteger (approved):

PR #489, opened 2024-06-05
feat(icon-aria-label): allow aria-label to be used as accessible name

nataliadelmar:
# Description

I made a PR in order to pass ariaLabel correctly to an icon. I am also allowing the passing of aria-label to the icon, which should be treated the same way as aria-label. This is to not break the various places in Cantina where aria-label is already passed to an IconNext.

# Links

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -81,7 +81,9 @@ const Icon: React.FC<Props> = (props: Props) => {
 
   const { inheritedColors, styleColors } = getColors();
 
-  const accessibleName = ariaLabel || title;
+  const accessibleName = ariaLabel || otherProps['aria-label'] || title;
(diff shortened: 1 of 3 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -81,7 +81,9 @@ const Icon: React.FC<Props> = (props: Props) => {
 
   const { inheritedColors, styleColors } = getColors();
 
-  const accessibleName = ariaLabel || title;
+  const accessibleName = ariaLabel || otherProps['aria-label'] || title;
```
maxinteger, line 84: We should allow `aria-label` or `ariaLabel` but not both.

To speed up things, I will approve this

Tech debt issue raised

maxinteger (approved):

PR #491, opened 2024-06-06
feat(tooltip-popover-combo): add new component

jor-row:
# Description

This creates a component that allows you to have a tooltip and a popover on the same trigger component. The tooltip and popover cannot be open at the same time.

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-525737

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -24,7 +24,7 @@ const Tooltip = forwardRef(
   ) => {
     // Update aria props manually, because
     // "The `aria` attribute is reserved for future use in React."
-    const updateInstance = useCallback(
+    const setInstance = useCallback(
       (popoverInstance: PopoverInstance | undefined) => {
         popoverInstance?.setProps({
           // see https://atomiks.github.io/tippyjs/v6/all-props/#aria
@@ -33,8 +33,9 @@ const Tooltip = forwardRef(
             content: isDescription ? 'describedby' : 'labelledby',
           },
         });
+        otherProps?.setInstance?.(popoverInstance);
       },
-      [isDescription]
+      [isDescription, otherProps?.setInstance]
     );
 
     const trigger = isToggletip ? 'click' : 'mouseenter focus';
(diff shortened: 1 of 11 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -23,17 +23,29 @@ const Tooltip = forwardRef(
       [isDescription]
     );
 
+    // Modified tooltipSetInstance to call both setInstance and updateInstance
+    const setTooltipInstance = useCallback(
+      (popoverInstance: PopoverInstance | undefined) => {
+        if (otherProps?.setInstance) {
+          otherProps.setInstance(popoverInstance);
+        }
```
maxinteger, line 31: I do not think we need the extra `useCallback` here.

If you rename `updateInstance` to `setInstance` then you do it there as well

also you can make it in a single line:
```suggestion
        otherProps?.setInstance?.(popoverInstance);
```

src/components/TooltipPopoverCombo/TooltipPopoverCombo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const [tooltipInstance, setTooltipInstance] = useState<PopoverInstance>();
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+
+  // Modified tooltipSetInstance to call both setInstance and updateInstance
+  const setMergedTooltipInstances = useCallback(
+    (popoverInstance: PopoverInstance | undefined) => {
+      if (otherTooltipProps?.setInstance) {
+        otherTooltipProps?.setInstance(popoverInstance);
+      }
```
maxinteger, line 29: nitpicking:

```suggestion
      otherTooltipProps?.setInstance?.(popoverInstance);
```

and the same for the rest

maxinteger (approved):
Looks good

src/components/TooltipPopoverCombo/TooltipPopoverCombo.tsx
jor-row: made suggested change

PR #494, opened 2024-06-10
feat: add toggletip component

maxinteger:
# Description

Created new `<Toggletip/>` component.

Also, removed the `<Tooltip\>`'s toggletip functionality.

# Links

Implemented based on [Tooltips & Toggletips](https://inclusive-components.design/tooltips-toggletips/)


maxinteger:
Note: In Chrome (on Mac) VoiceOver announced the live area text with significant delay when the DevTools was open. Make sure it closed during testing.

src/components/Popover/Popover.stories.args.ts
```diff
@@ -142,14 +142,14 @@ const popoverArgTypes = {
...
-    description: `Determines wether the focus should return to the trigger element when the popover is closed`,
+    description: `Determines whether the focus should return to the trigger element when the popover is closed`,
     control: { type: 'boolean' },
     table: {
       type: {
         summary: 'boolean',
       },
       defaultValue: {
-        summary: DEFAULTS.FOCUS_BACK_ON_TRIGGER_COMPONENT,
+        summary: DEFAULTS.FOCUS_BACK_ON_TRIGGER_COMPONENT_NON_INTERACTIVE,
```
torgeadelin, line 152: Why was thins changed?

src/components/Toggletip/Toggletip.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <>
+        <Popover
+          ref={ref}
+          appendTo={() => liveAriaRef.current}
+          trigger="click"
+          triggerComponent={React.cloneElement(triggerComponent, { ref: triggerComponentRef })}
+          showArrow
+          interactive={false}
+          addBackdrop={false}
+          role="generic"
```
torgeadelin, line 74: Not sure about this. 
But I'm assuming it had to be done because passing `tooltip` would make tippy add role=`tooltip` but also the `aria-describedby` - which doesn't really work with SR?

Or what was the intent here?

The reason I'm concerned is because: https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/ in the role and aria section it mentioned we need `role = tooltip` and `aria-describedby`

torgeadelin (commented):
Some questions, but otherwise looks great.

src/components/Popover/Popover.stories.args.ts
maxinteger: because the `FOCUS_BACK_ON_TRIGGER_COMPONENT` constant is does not exists,
not sure how it passed the build previously

src/components/Toggletip/Toggletip.tsx
maxinteger: Toggletip works completely different way then Tooltip.

While Tooltip is well defined in ARIA toggletip is not, for example there is no `toggletip` role.

I think we could use the `tooltip` role if we could refer to the content via `aria-describedby`. Unfortunately it is impossible, so we have to use live area to let SR know about the content of the tooltip.

As you can see the popover is appended to a `div` with `role=status`, if we keep the `role=tooltip` on the popover then we will have this:
```html
<div role="status">
   <div role="tooltip">
      ...
   </div>
</div>
```
and it is not clear how SR will behave in this case. It is safer to use the generic role for the content

torgeadelin (approved):

PR #495, opened 2024-06-11
fix(tooltip-popover-combo): fix id for aria labelled by

jor-row:
# Description

Make sure the aria-labelledby id matches the id on the button

Screenshot 2024-06-11 at 15 17 28


shnaaz (approved):
Resolve the conflict

PR #499, opened 2024-06-13
feat: change tooltip api and label tooltip behaviour

maxinteger:
# Description

Improve Tooltip component
- new, more explicit API
- label tooltip is more accessible


src/components/Tooltip/Tooltip.types.ts
```diff
@@ -31,7 +31,18 @@ export interface Props
...
+  /**
+   * This prop applicable only when `type` is `label`!
+   *
+   * When it is `true` Tooltip set the `aria-labelledby` on the trigger component which will shadow
+   * the label and/or aria-label,
+   * otherwise does nothing and the tooltip content will be hidden for the Screen Reader
+   *
+   * @default true
+   */
+  overwriteAccessibleLabel?: boolean;
```
lismith2-cisco, line 47: I think it would be simpler and less confusing just to have one parameter 'label'|'description'|'none'

nataliadelmar (approved):

PR #503, opened 2024-06-15
fix(PopoverNext): pass zIndex param to tippy, allow ariaLabel param to override default labelling

lismith2-cisco:
# IMPORTANT

# Description

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505185
https://sqbu-github.cisco.com/WebExSquared/webex-web-client/pull/5917

This is used for the positioning-only trigger for webex assistant. zIndex puts the assistant response popover slightly behind the webex assistant control popover. Added to MenuTrigger, just because other style-related props are. ariaLabel provides a label directly, as the positioning-only trigger cannot be used as aria-labelledby.

# Links

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -66,6 +66,8 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
     appendTo = DEFAULTS.APPEND_TO,
     continuePropagationOnTrigger,
     'aria-labelledby': providedAriaLabelledby,
+    zIndex,
+    'aria-label': ariaLabel,
     ...rest
   } = props;
 
@@ -84,7 +86,8 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
 
   const modalConditionalProps = {
     ...(interactive && {
-      'aria-labelledby': ariaLabelledby,
+      ...((providedAriaLabelledby || !ariaLabel) && { 'aria-labelledby': ariaLabelledby }),
+      ...(ariaLabel && { 'aria-label': ariaLabel }),
       focusLockProps: { restoreFocus: focusBackOnTrigger, autoFocus },
     }),
   };
@@ -123,7 +126,7 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
   if (interactive) {
     triggerComponentCommonProps['aria-haspopup'] =
       triggerComponent?.props?.['aria-haspopup'] || 'dialog';
-    if (!providedAriaLabelledby) {
+    if (!providedAriaLabelledby && !ariaLabel) {
       triggerComponentCommonProps['id'] = ariaLabelledby;
     }
   }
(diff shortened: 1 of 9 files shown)


src/components/Popover/Popover.types.ts
```diff
@@ -197,4 +202,9 @@ export interface Props extends PopoverCommonStyleProps, Partial<LifecycleHooks>
    * The element to append the popover to.
    */
   appendTo?: AppendToType;
+
+  /**
+   * Aria label for the popover. By default, the popover will be labelled by the triggerComponent.
+   */
+  ariaLabel?: string;
```
cscheffauer, line 209: there are attribute unit tests missing for the aria-label change

src/components/Popover/Popover.types.ts
```diff
@@ -197,4 +202,9 @@ export interface Props extends PopoverCommonStyleProps, Partial<LifecycleHooks>
    * The element to append the popover to.
    */
   appendTo?: AppendToType;
+
+  /**
+   * Aria label for the popover. By default, the popover will be labelled by the triggerComponent.
```
cscheffauer, line 207: maybe worth adding here to the comment to only use this ariaLabel in edge cases, like when its not possible to have discernible text on the trigger component (for example the trigger is something else, a different aria-label needs to be passed in when having multiple popovers attached to the same trigger, e.g.).

cscheffauer (changes requested):
looking good, 2 small comments

src/components/Popover/Popover.tsx
```diff
@@ -79,7 +81,8 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
 
   const modalConditionalProps = {
     ...(interactive && {
-      'aria-labelledby': triggerComponentId,
+      'aria-labelledby': ariaLabel ? undefined : triggerComponentId,
```
torgeadelin, line 84: to make props testing easier, i would love to only append if `aria-labelledby` ariaLabel is defined, instead of passing undefined when `ariaLabel` is falsy.

torgeadelin: this will prevent prop testing like this:
```js
expect(btn.props()).toEqual({
  ...
  "aria-labelledby": undefined,
})
```

src/components/Popover/Popover.types.ts
lismith2-cisco: oops

src/components/Popover/Popover.tsx
lismith2-cisco: sure it was just easier this way, but I can change it if you think it will help

lismith2-cisco: I think when I've seen similar before, it considers an unset key to be equal to an explicitly undefined key though

src/components/Select/Select.tsx
```diff
@@ -48,7 +48,10 @@ function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement
   const state = useSelectState(props);
```
lismith2-cisco, line 48: this file is just lint changes due to merge

src/components/Popover/Popover.types.ts
lismith2-cisco: added unit tests

src/components/Popover/Popover.tsx
lismith2-cisco: changed

cscheffauer (approved):

PR #505, opened 2024-06-17
fix(RadioGroup): add aria-describedby

stanjiawang:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

Add ```aria-describedby``` for ```RadioGroup``` and ```RadioSimpleGroup```

# Description

Add ```aria-describedby``` for ```RadioGroup``` and ```RadioSimpleGroup```
```descriptionProps``` is not supported in the current ```react-aria``` version, we cannot use ```descriptionProps``` returned by the hook ```useRadioGroup```. So we have to manually add ```aria-describedby``` to the radio group.

Note: we need to remove this once we upgrade the ```react-aria``` version.

# Links

https://react-spectrum.adobe.com/react-aria/useRadioGroup.html

### DIFF
--- a/src/components/RadioGroup/RadioGroup.tsx
+++ b/src/components/RadioGroup/RadioGroup.tsx
@@ -1,5 +1,6 @@
 import React, { FC } from 'react';
 import classnames from 'classnames';
+import { v4 as uuidV4 } from 'uuid';
 import { useRadioGroup } from '@react-aria/radio';
 import { useRadioGroupState } from '@react-stately/radio';
 
@@ -27,11 +28,23 @@ const RadioGroup: FC<RadioGroupProps> = (props: RadioGroupProps) => {
 
   const state = useRadioGroupState(props);
   const { radioGroupProps, labelProps } = useRadioGroup(props, state);
+  const radioGroupId = id || uuidV4();
+  const descriptionId = description ? `radio-group-description-${radioGroupId}` : undefined;
 
   return (
-    <div {...radioGroupProps} className={classnames(className, STYLE.group)} id={id} style={style}>
+    <div
+      {...radioGroupProps}
+      className={classnames(className, STYLE.group)}
+      id={radioGroupId}
+      style={style}
+      aria-describedby={description ? descriptionId : undefined}
+    >
       <span {...labelProps}>{label}</span>
-      {description && <Text type={TEXT_CONSTANTS.TYPES.BODY_SECONDARY}>{description}</Text>}
+      {description &&
+        <Text id={descriptionId} type={TEXT_CONSTANTS.TYPES.BODY_SECONDARY}>
+          {description}
+        </Text>
+      }
       <RadioContext.Provider value={state}>
         {options &&
           options.map((option: string | RadioProps) => {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by johnsoter13 — APPROVED** [review]

File: src/components/RadioGroup/RadioGroup.tsx [hunk 1]
```diff
@@ -29,9 +29,19 @@ const RadioGroup: FC<RadioGroupProps> = (props: RadioGroupProps) => {
   const { radioGroupProps, labelProps } = useRadioGroup(props, state);
 
   return (
-    <div {...radioGroupProps} className={classnames(className, STYLE.group)} id={id} style={style}>
+    <div
+      {...radioGroupProps}
+      className={classnames(className, STYLE.group)}
+      id={id}
```
nataliadelmar, line 35: id is optional, so if it's not provided then the aria-describedby would point to `radio-group-description-id-undefined (?) or `radio-group-description-id-`. We should generate a uuid like popovers component does in case we do not receive an id from props
image

src/components/RadioSimpleGroup/RadioSimpleGroup.tsx
```diff
@@ -25,9 +25,10 @@ const RadioSimpleGroup: FC<RadioSimpleGroupProps> = (props: RadioSimpleGroupProp
       data-disabled={state.isDisabled}
       id={id}
       style={style}
+      aria-describedby={description ? `radio-simple-group-description-id-${id}` : undefined}
     >
       {label && <span {...labelProps}>{label}</span>}
-      {description}
+      {description && <span id={`radio-simple-group-description-id-${id}`}>{description}</span>}
```
nataliadelmar, line 31: does adding a span wrapper around description makes the sr read the described by properly? in my head, description would have always been a node that has layers, multiple children, etc, and would come with its own aria-describedby on top:

description: `<div><Icon name-"something"><Text id="my-custom-id">This is the actual content of my description</Text></div>`

So in the component, given useRadioGroup is not passing aria-describedby down, then we could simply do:

```
const RadioSimpleGroup: FC<RadioSimpleGroupProps> = (props: RadioSimpleGroupProps) => {
  const { className, description, id, label, children, style } = props;
  const state = useRadioGroupState(props);
  const { radioGroupProps, labelProps } = useRadioGroup(props, state);
  return (
    <div
      {...radioGroupProps}
      data-disabled={state.isDisabled}
      id={id}
      style={style}
      aria-describedby={props['aria-describedby']}
    >
      {label && <span {...labelProps}>{label}</span>}
      {description}
...
```
So the consumer would call this like this:
```
<RadioGroup 
      aria-described-by="layout-radio-group-id"
      description=(
           <div><Icon name-"something"><Text id="layout-radio-group-id">Select only 1 layout please</Text></div>
      )
      ...
/>
```

src/components/RadioGroup/RadioGroup.tsx
stanjiawang: Code added to make sure id is not undefined.

```
  const radioSimpleGroupId = id || uuidV4();
  const descriptionId = description ? `radio-simple-group-description-${radioSimpleGroupId}` : undefined;
  ```

src/components/RadioSimpleGroup/RadioSimpleGroup.tsx
stanjiawang: Won't change this as discussion with @nataliadelmar

nataliadelmar (approved):

PR #507, opened 2024-06-18
feat(SelectNext): Escape key handling for trigger button

johnsoter13:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description
There is an issue in web client where escape won't close the menu that a select box is in because the select trigger button has focus and the onkeydown of the selectbox triggers instead of the menu's onkeydown
- Allowed the ability to pass escape functionality to the onkeydown of the trigger button
# Links

[https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506187](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506187)

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -38,6 +38,7 @@ function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement
     listboxMaxHeight,
     isInForm = DEFAULTS.IS_IN_FORM,
     listboxWidth,
+    escapeOnTriggerCallback,
   } = props;
   const [popoverInstance, setPopoverInstance] = useState<PopoverInstance>();
   const hasBeenOpened = useRef<boolean>(false);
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.tsx
```diff
@@ -88,6 +89,11 @@ function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement
    */
   const onKeyDown = useCallback((e) => {
     switch (e.key) {
+      case 'Escape':
+        if (escapeOnTriggerCallback) {
+          escapeOnTriggerCallback(e);
```
jor-row, line 94: This isn't ideal as when it is inside a popover of some sort you will have to have access to the popover instance to be able to have a callback that will close the popover.

This issue in Cantina is happening when it is inside a MenuTrigger and there is currently no way to control the popover from the outside (more momentum changes would be needed to make this fix work).

Looking at this issue more closely, I made a storybook with a Select inside a MenuTrigger and it works just fine. Looking at the code in Cantina, the issue is actually that the Select is wrapped inside a List component and it is the List that seems to be stopping propagation of the escape key press. Removing the List makes it work as expected which makes me think either something should be fixed in List, or, we should look at whether it needs to be inside of a List at all(maybe a menu or something works?)

jor-row (changes requested):

src/components/Select/Select.tsx
jor-row: If you've already explored these options just let me know, this solution will work it will just require a little more work

PR #508, opened 2024-06-19
feat(LinkNext): add Link component

jiwang7:
# Description
Snipaste_2024-06-20_15-32-27

- New Link component
- implement by react-aria/link useLink
- add storybook and test file 
 
# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-533449
https://react-spectrum.adobe.com/react-aria/useLink.html


src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * title to use for this component.
+   */
+  title?: string;
+
+  /**
+   * whether or not this component has icon【name = pop-out】 in the right
+   */
+  isWithIcon?: boolean;
```
nataliadelmar, line 33: I suggest this name:
opensNewTab?: boolean

src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * icon props
+   * When use iconProps isWithIcon is necessary
+   */
+  iconProps?: Partial<IconPrpos>;
```
nataliadelmar, line 39: typo in IconPrpos

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,47 @@
...
+          data-disabled={props.disabled || DEFAULTS.DISABLED}
+          data-inverted={props.inverted || DEFAULTS.INVERTED}
+          title={title}
+        >
+          {props.children}
+        </a>
+        {isWithIcon && <Icon className={STYLE.icon} scale={16} name="pop-out" {...iconProps} />}
+      </div>
+    </FocusRing>
+  );
```
nataliadelmar, line 42: missing: aria-label for the link and the link+icon. Let's research what is the best and conventional way of doing this? should the aria-label be on the div container wrapper? or in each individual element?

I suggest:
1. Add a property called 'opensNewTabIndicatorAriaLabel'?: string
2. const linkId = id passed from props || uuiid generated like popover component does
3. In the a tag, add an id prop = {linkId}
4. In the a tag, add an aria-labelledby. if !opensNewTab, then `aria-labelledby = {linkId}`; otherwise, if opensNewTabIndicatorAriaLabel, then `aria-labelledby  = ${linkId} ${opensNewTabIndicatorAriaLabel}`

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,47 @@
...
+          className={classnames(STYLE.wrapper, className)}
+          {...linkProps}
+          ref={ref}
+          data-disabled={props.disabled || DEFAULTS.DISABLED}
+          data-inverted={props.inverted || DEFAULTS.INVERTED}
+          title={title}
+        >
+          {props.children}
+        </a>
+        {isWithIcon && <Icon className={STYLE.icon} scale={16} name="pop-out" {...iconProps} />}
```
nataliadelmar, line 39: This icon is an informative icon (it conveys important meaning that is not explicitly visible in the screen) and non-sr users wouldn't know what it means. We would need to add a Tooltip it that with the content: {opensNewTabIndicatorLabel}

Wrap the icon on a ghost button circle to allow the tooltip to trigger on hover.

Add an uuid as id to the button circle. 
const iconId = uuid.v4()

nataliadelmar: tooltip should only be there if opensNewTabIndicatorLabel is defined, otherwise we risk an empty tooltip appearing on screen

nataliadelmar: image
check this tooltip for example from https://equalizedigital.com/accessibility-checker/link-opens-new-window-or-tab/#:~:text=Links%20that%20open%20new%20windows,link%20in%20the%20current%20tab.

src/components/Link/Link.types.ts
JudyZhuHz: I think isWithIcon is good, because this parameter is to distinguish whether to add an icon after the text.
For example, if the link not need open a new tab, don't need add pop-out icon after the text.

src/components/Link/Link.tsx
jiwang7: fixed but wait tooltip component update

jiwang7: Tested, it works already

src/components/Link/Link.style.scss
```diff
@@ -0,0 +1,67 @@
...
+    cursor: auto;
+  }
+}
+
+.md-link-container {
+  display: flex;
+  align-items: center;
+
+  .md-link-icon {
+    margin-left: 5px;
```
cscheffauer, line 65: all our spacing etc should be in `rem`, not px

cscheffauer (changes requested):
Tested locally - multiple issues:
- When opening it in storybook, i am trying to focus to it: i don't see the FocusRing appearing at all 
- The inverted variant should be removed completely, since the colors there are definitely not meeting the contrast ratio specs needed for accessibility.

src/components/Link/Link.types.ts
jor-row: But the reason we are adding this is because accessibility says if it opens in a new tab it should always have an icon, no? So ```opensNewTab``` makes more sense, or at least make it clear what icon it is. ```isWithIcon``` is too generic when the icon is hard coded and not changeable.

src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * icon props
+   * When use iconProps isWithIcon is necessary
+   */
+  iconProps?: Partial<IconProps>;
+
+  /**
+   * Used for tooltip content
+   */
+  opensNewTabIndicatorLabel?: string;
```
nataliadelmar, line 45: After discussing with Christoph, this Link should also accept the case in which the element is semantically a Link, but visibly a Button or MenuItem (we would send our own styling through props to make it look like a button)

For this reason, the tooltip could say: a) "Opens a new tab" only or b) the Button label + "Opens a new tab". 

So, we should change the name of this prop to 'tooltipContent?: string'

nataliadelmar: Additionally, we would need to add another prop called 'tooltipType' which is of type tooltipType (you can get that from Tooltip component's types), and that way the consumer can specify if this tooltip is the entire label of the link or only the description of the link.
type: 'label' | 'description' | 'none'

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  href: 'https://www.webex.com/',
+  target: '_blank',
+  hasIcon: true,
+  iconProps: { id: 'icon' },
+  opensNewTabIndicatorLabel: 'opens a new window'
+};
+
+LinkWithIcon.argTypes = { ...argTypes };
+
+export { Example, States, LinkWithIcon };
```
nataliadelmar, line 54: Following my other comment, we should add a story where the link look like a button pill and a button circle

nataliadelmar: Note: when the link looks like a button circle, there is no need to add an icon of 'new tab', so hasIcon: true in this case

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const ref = providedRef || internalRef;
+
+  const mutatedProps = {
+    ...otherProps,
+    title,
+    isDisabled: props.disabled,
+    isInverted: props.inverted,
+  };
+
+  delete mutatedProps.disabled;
```
maxinteger, line 23: you can "extract" the `disabled` value in line 12, so you do not need to delete the property here

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,79 @@
...
+            type="description"
+            placement='bottom'
+            triggerComponent={
+              <a
+                className={classnames(STYLE.wrapper, className)}
+                {...linkProps}
+                ref={ref}
+                data-disabled={props.disabled || DEFAULTS.DISABLED}
+                data-inverted={props.inverted || DEFAULTS.INVERTED}
+                title={title}
```
maxinteger, line 41: you can make this code more DRY if you put all the common props of `<a>` into an an object and use here and for the other `<a>` in line 61

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,79 @@
...
+              >
```
maxinteger, line 42: Please add `rel="noopener noreferrer` to external links.

More info here: 
https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

Maybe we can add an option to turn it off since most of our link pointing to Cisco sites, but either way this must be the default

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  delete mutatedProps.disabled;
+
+  const { linkProps } = useLink({ ...mutatedProps, elementType: 'a' }, ref);
+
+  return (
+    <FocusRing disabled={props.disabled}>
+      <>
+        {opensNewTabIndicatorLabel && (
```
maxinteger, line 30: Please use the `target` prop value for check is it an external link or not

for example:
```ts
const isExternalLinkl = props.target === '_blank';
```

probably you can drop the `Tooltip` when there is no `opensNewTabIndicatorLabel`, but the icon and the `rel` must be set for all external links

maxinteger (changes requested):

src/components/Link/Link.style.scss
```diff
@@ -0,0 +1,67 @@
...
+  cursor: pointer;
+  font-family: var(--md-globals-font-default);
+  outline: none !important;
+  padding: 0;
+  text-decoration: underline;
+  transition: background-color 0.2s, color 0.2s, border-color 0.2s;
+  display: inline-block;
+
+  &:focus,
+  &.focus {
```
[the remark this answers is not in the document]
jiwang7: image
I have tested, when I remove &:focus. It will display another color, which is from a:focus. So, I think we still need &:focus style.

src/components/Link/Link.types.ts
jiwang7: ok, I will fix

src/components/Link/Link.tsx
jiwang7: Do I just need to add rel = 'noopener noreferrer' for <a> with target = '_blank'? or I should add rel = 'noopener noreferrer' as a common props for a tag.

maxinteger: Let's keep it simple, the way you implemented is good for now.

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  target: '_blank',
+  hasIcon: true,
+  iconProps: { id: 'icon' },
+  tooltipContent: 'opens a new window',
+  type: 'description',
+};
+
+LinkHasIcon.argTypes = { ...argTypes };
+
+const ButtonCircleLink = Template<LinkProps>(Link).bind({});
```
maxinteger, line 62: This is incorrect. `<a>` can not contain interactive element 

see [A tag HTML Spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element:~:text=The%20a%20element%20can%20be%20wrapped%20around%20entire%20paragraphs%2C%20lists%2C%20tables%2C%20and%20so%20forth%2C%20even%20entire%20sections%2C%20so%20long%20as%20there%20is%20no%20interactive%20content%20within%20(e.g.%2C%20buttons%20or%20other%20links).%20This%20example%20shows%20how%20this%20can%20be%20used%20to%20make%20an%20entire%20advertising%20block%20into%20a%20link%3A)

maxinteger (changes requested):

src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * title to use for this component.
+   */
+  title?: string;
+
+  /**
+   * whether or not this component has icon【name = pop-out】 in the right
+   * icon display needed tooltipContent 
+   */
+  hasIcon?: boolean;
```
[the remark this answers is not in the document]
jiwang7: We already have a props named hasIcon, which is like `hideExternalLinkIcon`, we can hide the icon by setting `hasIcon = false`. And We can set IconProps = { name: 'xxx'} to set another icon. I think it is flexible enough.

src/components/Link/Link.stories.args.ts
```diff
@@ -0,0 +1,97 @@
...
+    table: {
+      type: {
+        summary: 'string',
+      },
+      defaultValue: {
+        summary: undefined,
+      },
+    },
+  },
+  type: {
```
nataliadelmar, line 84: 'tooltipType' should be the name of it

src/components/Link/Link.stories.tsx
nataliadelmar: correct: we have a button circle and pill requirement, but that means we will have to recreate it from scratch and not use ButtonCircle or ButtonPill as children of this Link. This way:
```
const ButtonCircleLink = Template<LinkProps>(Link).bind({});

ButtonCircleLink.args = {
  href: 'https://www.webex.com/',
  target: '_blank',
  tooltipContent: 'opens a new tab',
  type: 'description',
  children: (
    <div style={{// all the styling this div needs to become a circle, with fixed dimensions, with fixed var colors, display: flex, center-allign}}>
      <Icon name="chat-group" autoScale={150} />
    </div>
  ),
};

```

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const ButtonPillLink = Template<LinkProps>(Link).bind({});
+
+ButtonPillLink.args = {
+  href: 'https://www.webex.com/',
+  target: '_blank',
+  iconProps: { id: 'icon' },
+  tooltipContent: 'opens a new tab',
+  type: 'description',
+  children: <ButtonPill>example</ButtonPill>,
```
nataliadelmar, line 87: Same here: w have to recreate the buttonPill from scratch and not use ButtonPill as children of this Link. This way:
```
const ButtonPillLink = Template<LinkProps>(Link).bind({});

ButtonPillLink.args = {
  href: 'https://www.webex.com/',
  target: '_blank',
  tooltipContent: 'opens a new tab',
  type: 'description',
  children: (
    <div style={{// all the styling this div needs to become a pill, with fixed dimensions, with fixed var colors, display: flex, center-allign}}>
      <Text>Go to Download page</Text>
      <Icon name="chat-group" autoScale={150} />
    </div>
  ),
};

```

src/components/Link/Link.style.scss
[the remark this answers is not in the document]
nataliadelmar: for some reason, that mdl token is winning the race, please debug why; we shouldn't be falling into mdl (momentum-design-legacy) tokens anymore

src/components/Link/Link.types.ts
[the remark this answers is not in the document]
nataliadelmar: Do we always use target="_blank" when we are opening a new tab in Cantina?

src/components/Link/Link.stories.tsx
jiwang7: I misunderstood the requirement. I will fix it.

src/components/Link/Link.style.scss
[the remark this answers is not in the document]
jiwang7: About two issues @cscheffauer  mentioned before https://github.com/momentum-design/momentum-react-v2/pull/508#pullrequestreview-2156164854.
1. In Link component. We should set `href` props, if we don't set `href`, it can't be focused. I have added `href` in every storybook example. And it can be focused already.
2.we have conversation in acc marmot space. The inverted style is needed.
So, could you please take a book at these changes. Thank you!

src/components/Link/Link.types.ts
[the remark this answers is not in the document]
maxinteger: Ok, we can keep the prop, but:
- when it is not set and `target=_blank` => DO show the icon
- when it is not set and it is a normal link =>  DON'T show the icon
- otherwise show the icon based on the `hasIcon` values

BTW I think we should use a more explicit name, ex. `hasExternalLinkIcon` or something similar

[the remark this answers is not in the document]
jiwang7: I accept we should use `hasExternalLinkIcon`. But for target to effect displaying icon, I think it is out of bounds of target. Target should just control open link ways. It is not its responsibility. I think we should make `hasExternalLinkIcon` single-purpose.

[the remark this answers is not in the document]
maxinteger: I think the only reasonable use case for the "pop-out" icon is when the link is opening on a new tab which only happens with the `target=_blank". Also it is a requirement for A11y.

If the default visibility of the icon is not depend on the `target` then the consumer must remember to set both `target` and `hasExternalLinkIcon` which is error prone. 

The only reason why we need the `hasExternalLinkIcon` is because in some rare case consumer might want to hide the icon even when the link is external.

@nataliadelmar What do you think?

src/components/Link/Link.style.scss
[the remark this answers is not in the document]
nataliadelmar: acc style is needed to to a legacy style that sets color to --mdl-textColorHyperling-hover (legacy color). We don't want to remove that line because we are afraid of breaking existing code that relies on that line to get the blue color. We also cannot put !important in our css rule because we need states to change depending on if they are on hover, normal or pressed

nataliadelmar (changes requested):

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,79 @@
...
+        {tooltipContent && (
+          <Tooltip
+            type={tooltipType}
+            placement="bottom"
+            triggerComponent={
+              <a {...commonProps}>
+                <div className={STYLE.container}>
+                  {props.children}
+                  {hasExternalLinkIcon && (
+                    <Icon className={STYLE.icon} scale={16} name="pop-out" {...iconProps} />
```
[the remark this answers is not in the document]
jiwang7: In this change, when `hasExternalLinkIcon = false` and `target === 'blank'`, the Icon still exist. According to your description. Maybe we should implement like this？

`
let isShowIcon = false;

  if(hasExternalLinkIcon === undefined) {
    props.target === '_blank';
    isShowIcon = true;
  } else if (hasExternalLinkIcon) {
    isShowIcon = true;
  } else {
    isShowIcon = false;
  }
  
  isShowIcon && <Icon className={STYLE.icon} scale={16} name="pop-out" {...externalLinkIconProps} />
  `

src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,62 @@
...
+   * whether or not this component has icon【default icon name = pop-out】 in the right
+   * icon display needed tooltipContent 
+   */
+  hasExternalLinkIcon?: boolean;
+
+  /**
+   * icon props
+   * When use iconProps hasExternalLinkIcon is necessary
+   */
+  iconProps?: Partial<IconProps>;
```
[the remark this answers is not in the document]
jiwang7: ok

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+delete States.argTypes.disabled;
+delete States.argTypes.inverted;
+
+const LinkHasIcon = Template<LinkProps>(Link).bind({});
+
+LinkHasIcon.args = {
+  href: 'https://www.webex.com/',
+  target: '_blank',
+  hasExternalLinkIcon: true,
+  iconProps: { id: 'icon' },
```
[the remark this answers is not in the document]
jiwang7: Yes, i can remove it. By the way if we don't have href in Link, we can't focus Link.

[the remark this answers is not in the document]
jiwang7: In LinkHasIcon storybook example, If we don't have `target`, It's no different from a tag. So basically I think we need `target: '_blank' `

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const ButtonCircleLink = Template<LinkProps>(Link).bind({});
+
+ButtonCircleLink.args = {
+  href: 'https://www.webex.com/',
+  target: '_blank',
+  iconProps: { id: 'icon' },
+  tooltipContent: 'opens a new tab',
+  tooltipType: 'description',
+  children: (
```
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.style.scss
```diff
@@ -0,0 +1,87 @@
...
+  .md-link-icon {
+    margin-left: 0.3125rem;
+  }
+}
+
+.md-link-ButtonPillLink {
+  text-decoration: none;
+}
+
+.md-link-button-wrapper {
```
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.style.scss
```diff
@@ -0,0 +1,87 @@
...
+.md-link-container {
+  display: flex;
+  align-items: center;
+
+  .md-link-icon {
+    margin-left: 0.3125rem;
+  }
+}
+
+.md-link-ButtonPillLink {
```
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.stories.args.ts
```diff
@@ -0,0 +1,97 @@
...
+      defaultValue: {
+        summary: false,
+      },
+    },
+  },
+  iconProps: {
+    description: 'icon props',
+    table: {
+      type: {
+        summary: 'IconNextProps',
```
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Used for tooltip content
+   */
+  tooltipContent?: string;
```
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.tsx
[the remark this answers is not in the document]
nataliadelmar: I see your point; the last else is redundant because of the first line is already setting it to false:
```
let isShowIcon = false;

if(hasExternalLinkIcon === undefined && props.target === '_blank') {
    isShowIcon = true;
} else if (hasExternalLinkIcon) {
    isShowIcon = true;
}

isShowIcon && <Icon className={STYLE.icon} scale={16} name="pop-out" {...externalLinkIconProps} />
```

src/components/Link/Link.types.ts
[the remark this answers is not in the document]
nataliadelmar: Please also update the description of all properties on stories.args.ts with these latest description. For example, the description of tooltipContent is still 'Used for tooltip content'

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+    >
+      <Icon name="chat-group" autoScale={150} />
+    </div>
+  ),
+};
+
+ButtonCircleLink.argTypes = { ...argTypes };
+delete ButtonCircleLink.argTypes.children;
+delete ButtonCircleLink.argTypes.disabled;
+delete ButtonCircleLink.argTypes.disabled;
```
nataliadelmar, line 102: double disabled delete; the last delete should be of the inverted prop, which is not relevant for the button circle:
```
delete ButtonCircleLink.argTypes.children;
delete ButtonCircleLink.argTypes.disabled;
delete ButtonCircleLink.argTypes.inverted;
```

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <Text>Go to Download page</Text>
+      <Icon name="chat-group" autoScale={150} />
+    </div>
+  ),
+};
+
+ButtonPillLink.argTypes = { ...argTypes };
+delete ButtonPillLink.argTypes.children;
+delete ButtonPillLink.argTypes.disabled;
+delete ButtonPillLink.argTypes.disabled;
```
nataliadelmar, line 140: double disabled delete; the last delete should be of the inverted prop, which is not relevant for the button circle:
```
delete ButtonPillLink.argTypes.children;
delete ButtonPillLink.argTypes.disabled;
delete ButtonPillLink.argTypes.inverted;
```

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+};
+
+States.argTypes = { ...argTypes };
+delete States.argTypes.disabled;
+delete States.argTypes.inverted;
+
+const LinkHasIcon = Template<LinkProps>(Link).bind({});
+
+LinkHasIcon.args = {
+  target: '_blank',
```
nataliadelmar, line 62: missing an href otherwise the link is not focusable

nataliadelmar (approved):

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+  tooltipType: 'description',
+};
+
+LinkHasIcon.argTypes = { ...argTypes };
+
+const ButtonCircleLink = Template<LinkProps>(Link).bind({});
+
+ButtonCircleLink.args = {
+  tooltipContent: 'opens a new tab',
+  tooltipType: 'description',
```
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+ButtonCircleLink.argTypes = { ...argTypes };
+delete ButtonCircleLink.argTypes.children;
+delete ButtonCircleLink.argTypes.disabled;
+delete ButtonCircleLink.argTypes.disabled;
+
+const ButtonPillLink = Template<LinkProps>(Link).bind({});
+
+ButtonPillLink.args = {
+  tooltipContent: 'opens a new tab',
+  tooltipType: 'description',
```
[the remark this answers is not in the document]
jiwang7: added

src/components/Link/Link.stories.tsx
jiwang7: added

src/components/Link/Link.stories.tsx
jiwang7: fixed

src/components/Link/Link.stories.tsx
jiwang7: fixed

src/components/Link/Link.types.ts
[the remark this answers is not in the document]
jiwang7: fixed

src/components/Link/Link.tsx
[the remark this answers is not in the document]
jiwang7: fixed

cscheffauer (approved):

src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+States.parameters = {
+  variants: [
+    { label: 'Default link', href: 'https://www.webex.com/' },
+    { label: 'Inverted', inverted: true, href: 'https://www.webex.com/' },
+  ],
+};
+
+States.argTypes = { ...argTypes };
+delete States.argTypes.disabled;
+delete States.argTypes.inverted;
```
maxinteger, line 57: nitpicking: instead of `delete` please use `pick` or `omit` from lodash

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const { linkProps } = useLink({ ...mutatedProps, elementType: 'a' }, ref);
+
+  let isShowIcon = false;
+
+  if (hasExternalLinkIcon === undefined && props.target === '_blank') {
+    isShowIcon = true;
+  } else if (hasExternalLinkIcon) {
+    isShowIcon = true;
+  }
```
maxinteger, line 45: nitpicking: this could be simplified like this:

```suggestion
  let isShowIcon = hasExternalLinkIcon === true || isExternalLink;
```

src/components/Link/Link.unit.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+      expect(element.props()).toEqual({
+        className: STYLE.icon,
+        scale: 16,
+        name: 'pop-out',
+        autoScale: true,
+      });
+    });
+
+    it('should have rel when target is provided', async () => {
```
maxinteger, line 190: issue: please check the external icon as well, it should be there when the target is _blank

Also, you need one more test for the `target=_blank` and `hasExternalLinkIcon=false`

maxinteger (changes requested):

src/components/Link/Link.tsx
jiwang7: In this simplified change, `hasExternalLinkIcon` can't force hide it.
If `hasExternalLinkIcon  = false` but `props.target === '_blank'`, it will display icon.
So, I think we can't simple like this. I will replace `props.target === '_blank'` to `isExternalLink `

src/components/Link/Link.stories.tsx
jiwang7: fixed

src/components/Link/Link.unit.test.tsx
jiwang7: fixed

src/components/Link/Link.tsx
maxinteger: you are right, I missed that case

maxinteger (approved):
Looks great!

PR #512, opened 2024-06-20
fix(TooltipPopoverCombo): fix the aria-labelledby of the popover when the button has a tooltip

lismith2-cisco:
# Description

This fixes the aria-labelledby chaining issue to provide the correct label for the popover in TooltipPopoverCombo, by allowing Popover to have aria-labelledby overridden and allowing Tooltip to have labelId overridden. This fix removes some duplicate and now-unnecessary ids.

This also removes the extra aria-haspopup which was incorrectly being added to the tooltip itself in TooltipPopoverCombo.

This also fixes Tooltip treating type none the same as type description.

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,4 +1,4 @@
-import React, { ForwardedRef, forwardRef, useCallback, useEffect } from 'react';
+import React, { ForwardedRef, forwardRef, useCallback, useEffect, useRef } from 'react';
 import './Popover.style.scss';
 import ModalContainer from '../ModalContainer';
 import ButtonCircle from '../ButtonCircle';
@@ -65,6 +65,7 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
     autoFocus = DEFAULTS.AUTO_FOCUS,
     appendTo = DEFAULTS.APPEND_TO,
     continuePropagationOnTrigger,
+    'aria-labelledby': providedAriaLabelledby,
     ...rest
   } = props;
 
@@ -76,11 +77,14 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
 
   const popoverInstance = React.useRef<PopoverInstance>(undefined);
 
-  const triggerComponentId = triggerComponent.props?.id || uuidV4();
+  const generatedTriggerIdRef = useRef(uuidV4());
+  const generatedTriggerId = generatedTriggerIdRef.current;
+
+  const ariaLabelledby = providedAriaLabelledby || triggerComponent.props?.id || generatedTriggerId;
 
   const modalConditionalProps = {
     ...(interactive && {
-      'aria-labelledby': triggerComponentId,
+      'aria-labelledby': ariaLabelledby,
       focusLockProps: { restoreFocus: focusBackOnTrigger, autoFocus },
     }),
   };
@@ -114,13 +118,14 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
     firstFocusElement?.focus();
   }, [firstFocusElement]);
 
-  const triggerComponentCommonProps = {
-    id: interactive ? triggerComponentId : triggerComponent.props?.id || id,
-  };
+  const triggerComponentCommonProps = {};
 
   if (interactive) {
     triggerComponentCommonProps['aria-haspopup'] =
       triggerComponent?.props?.['aria-haspopup'] || 'dialog';
+    if (!providedAriaLabelledby) {
+      triggerComponentCommonProps['id'] = ariaLabelledby;
+    }
   }
 
   const mrv2Props = isMRv2Button(triggerComponent) ? { useNativeKeyDown: true } : {};
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -27,11 +27,14 @@ const Tooltip = forwardRef(
       variant = DEFAULTS.VARIANT,
       triggerComponent,
       children,
+      labelOrDescriptionId: providedLabelId,
+      'aria-haspopup': ariaHaspopup,
       ...otherProps
     }: Props,
     ref: ForwardedRef<HTMLElement>
   ) => {
-    const id = useId();
+    const generatedLabelId = useId();
+    const labelId = providedLabelId || generatedLabelId;
     const isLabelTooltip = type === 'label';
     const isDescription = type === 'description';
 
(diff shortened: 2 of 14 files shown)


cscheffauer (changes requested):
had a quick look, not tested locally yet

src/components/Popover/Popover.tsx
```diff
@@ -114,7 +115,7 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
   }, [firstFocusElement]);
 
   const triggerComponentCommonProps = {
-    id: interactive ? triggerComponentId : triggerComponent.props?.id || id,
```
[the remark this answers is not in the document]
jor-row: That was only added last week here https://github.com/momentum-design/momentum-react-v2/pull/496/files to fix something that is now using TooltipPopoverCombo

Given that I don't think this is will break things in Cantina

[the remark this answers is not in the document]
lismith2-cisco: Popover component still allows to pass in id prop, and it will be used as id for the popover (ModalContainer). The change made the popover and the button have the same id, which should never happen.

src/components/Popover/Popover.tsx
```diff
@@ -75,11 +76,13 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
 
   const popoverInstance = React.useRef<PopoverInstance>(undefined);
 
-  const triggerComponentId = triggerComponent.props?.id || uuidV4();
+  const [ariaLabelledby] = useState(
```
cscheffauer, line 79: useState doesn't have any effect here, since there is no setter `setAriaLabelledBy` used. You can just write it without the useState.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -44,17 +46,18 @@ const Tooltip = forwardRef(
           aria: {
             expanded: false,
             // we add `aria-labelledby` manually, see below
```
cscheffauer, line 48: this comment is not applicable anymore

src/components/Tooltip/Tooltip.types.ts
```diff
@@ -39,4 +39,6 @@ export interface Props
     | 'setInstance'
   > {
   type: TooltipTypes;
+  labelId?: string;
```
cscheffauer, line 42: nitpick: prop types should ideally have proper TSDocs, otherwise its a bit hard for consumers in Cantina to understand the meaning of the props, especially the prop `type`

cscheffauer (changes requested):

src/components/Popover/Popover.tsx
lismith2-cisco: This is to only call uuidv4 once and store the value between renders.

src/components/Tooltip/Tooltip.tsx
lismith2-cisco: this comment is atill applicable

src/components/Tooltip/Tooltip.types.ts
lismith2-cisco: this is true, I will add it (it wasn't me that added type though!)

src/components/Popover/Popover.tsx
nataliadelmar: you can use useRef to persist the value across renders; I think TextInput component is doing that

src/components/Tooltip/Tooltip.types.ts
```diff
@@ -38,5 +38,20 @@ export interface Props
...
+  /**
+   * When type is 'label', you can optionally use labelId specify the id to use for labelling.
+   * This enables you to use the same id to label e.g. a dialog, as is used by TooltipPopoverCombo.
+   */
+  labelId?: string;
+  /**
+   * aria-haspopup will be passed through to the trigger component if provided.
+   * This is only really useful for a nested component structure as in TooltipPopoverCombo.
+   */
+  'aria-haspopup'?: string;
```
nataliadelmar, line 56: is the type of this prop something we can import from aria?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -51,11 +54,11 @@ const Tooltip = forwardRef(
...
     );
 
-    const newTriggerComponent = isLabelTooltip
-      ? React.cloneElement(triggerComponent, { 'aria-labelledby': id })
-      : isDescription
-      ? React.cloneElement(triggerComponent, { 'aria-describedby': id })
-      : triggerComponent;
+    const newTriggerComponent = React.cloneElement(triggerComponent, {
+      ...(isLabelTooltip && { 'aria-labelledby': labelId }),
+      ...(isDescription && { 'aria-describedby': labelId }),
```
nataliadelmar, line 59: what about changing the name of labelId to 'tooltipId' to not make it label-exclusive?

src/components/Popover/Popover.tsx
lismith2-cisco: I am aware that we are, but I don't think we should be, the useRef docs say "Do not write or read ref.current during rendering... If you have to read or write something during rendering, use state instead." see 'Pitfall', https://react.dev/reference/react/useRef

src/components/Tooltip/Tooltip.tsx
lismith2-cisco: changed to labelOrDescriptionId since it isn't applied to the actual tooltip

src/components/Tooltip/Tooltip.types.ts
lismith2-cisco: Yes! Done.

src/components/Tooltip/Tooltip.types.ts
lismith2-cisco: added docs

src/components/Popover/Popover.tsx
nataliadelmar: "Do not write or read ref.current during rendering, except for [initialization.]"(https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents)

nataliadelmar: this isn't inside the render, it is in intialization

lismith2-cisco: back to the controversial useState

cscheffauer: @lismith2-cisco what u mentioned from the docs i think is not applicable - the sentence "Do not write or read ref.current during rendering" to me is a note that ref.current doesn't automatically re-render your dom, which means you shouldn't use useRef if your UI depends on it. But since in our case we only want to store the initial value of the uuidv4 and persist it between re-renders, i think we should use useRef.
Also from the same docs:
"If your component needs to store some value, but it doesn’t impact the rendering logic, choose refs."
https://react.dev/learn/referencing-values-with-refs#when-to-use-refs

For the sake of consistency across our code base my suggestion would be to use useRef, same as in all the other occurrences. I would be happy to discuss this further and would recommend you providing more information on how we should do it (with pros and cons) and have a convention for it moving forward - since there is no convention in place, my vote is to stick to consistency.

src/components/TooltipPopoverCombo/TooltipPopoverCombo.unit.test.tsx.snap
```diff
@@ -13,30 +13,25 @@ exports[`<TooltipPopoverCombo /> snapshot should match snapshot 1`] = `
     </Text>
   }
   triggerComponent={
-    <button
-      aria-haspopup="dialog"
-      id="test-id"
-    >
+    <button>
```
cscheffauer, line 16: why has the button the aria-haspopup removed here? i can see in the unit tests that aria-haspopup is passed in and in my opinion the button should still have aria-haspopup - is this a bug?

cscheffauer (changes requested):

src/components/TooltipPopoverCombo/TooltipPopoverCombo.unit.test.tsx.snap
lismith2-cisco: The component that it renders still has aria-haspopup, it's just set by Popover and passed through Tooltip instead of being set in TooltipPopoverCombo. You can see it in line 258/436

cscheffauer (approved):

PR #515, opened 2024-06-24
feat(tooltip-description): tooltip type description

nataliadelmar:
# Description

Whe tooltip type was "description", we were relying on tippy's default logic to link trigger and tooltip with aria-describedby. But that was adding the id to the external tooltip layer and not to the modal itself. So SR was not reading the content when focus on the button.

# Links

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -43,8 +43,7 @@ const Tooltip = forwardRef(
           // see https://atomiks.github.io/tippyjs/v6/all-props/#aria
           aria: {
             expanded: false,
-            // we add `aria-labelledby` manually, see below
-            content: isLabelTooltip || isDescription ? null : 'describedby',
+            content: null,
           },
         });
         otherProps?.setInstance?.(popoverInstance);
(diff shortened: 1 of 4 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -44,7 +44,7 @@ const Tooltip = forwardRef(
           aria: {
             expanded: false,
             // we add `aria-labelledby` manually, see below
-            content: isLabelTooltip ? null : 'describedby',
+            content: isLabelTooltip || isDescription ? null : 'describedby',
```
lismith2-cisco, line 47: line should just be:
`content: null,`
there was a bug previously

src/components/Tooltip/Tooltip.tsx
```diff
@@ -63,11 +65,12 @@ const Tooltip = forwardRef(
     // With SR the user can Read the page content without changing the focus so we need to provide a
     // always accessible label for the button.
```
lismith2-cisco, line 66: this comment should be updated to refer to label or description (and the line below)

src/components/Tooltip/Tooltip.unit.test.tsx
```diff
@@ -373,7 +373,7 @@ describe('<Tooltip type"description />', () => {
       );
```
lismith2-cisco, line 373: You should check Button has the correct description before tooltip opened as in the previous test for label

lismith2-cisco (changes requested):

lismith2-cisco (approved):

PR #517, opened 2024-06-24
feat(lightbox-legacy): add aria-labels

nataliadelmar:
# Description

As part of SPARK-503206, the buttons on the lightbox are missing sr-labels.

Cantina is already sending them using the tooltips array. In this PR, I just reuse these tooltips for the aria-labels of these buttons. 
image


src/legacy/Lightbox/index.js
```diff
@@ -354,6 +354,7 @@ class Lightbox extends React.Component {
           className="md-lightbox__page-control md-lightbox__page-control-icon md-lightbox__page-controls--left"
           role="button"
           tabIndex="0"
+          aria-label={tooltips.previous}
```
maxinteger, line 357: Something wrong with the label of the previous/next buttons on the side of the image.

Voice Over announce the image name first and then label of the button twice and then one more time the other button, for example on next VO announce "image.png Next Next Previous"


https://github.com/momentum-design/momentum-react-v2/assets/983050/ec705b33-d05d-4e85-894b-627b359d11f2

maxinteger (changes requested):
The previous/next announced incorrectly, otherwise looks good

maxinteger (approved):

PR #523, opened 2024-06-28
Fix: fix submenu customized menuitem click crash issue

junhao3268:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description
1.add title prop in ListItemBase for link button extra tooltip indicator.
2.fix submenu customized menuitem click crash issue.

image
image

# Links


src/components/ListItemBase/ListItemBase.constants.ts
```diff
@@ -20,6 +20,7 @@ const DEFAULTS = {
   ROLE: 'listitem',
   SHOULD_ITEM_FOCUS_BE_INSET: false,
   INTERACTIVE: true,
+  TITLE: 'title',
```
maxinteger, line 23: List item should not have default title and it never user anyway, please remove this line.

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -288,6 +289,7 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
         data-interactive={interactive}
         className={classnames(className, STYLE.wrapper, { active: isPressed || isSelected })}
         role={role}
+        title={title}
```
maxinteger, line 292: I do not understand why do you use `title` attribute here. And looks like it does not related to the title of this PR. Could you please share the Jira issue link?

maxinteger (changes requested):

src/components/ListItemBase/ListItemBase.tsx
junhao3268: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505149
we discussed with @nataliadelmar about the visual indicator in last catchups, decide to add Icon after the list item (link like a button), and add a extra title for hover tooltip to further display the visible indicator, so we need to add the titlt prop to solve the type error.
image

junhao3268: > I do not understand why do you use `title` attribute here. And looks like it does not related to the title of this PR. Could you please share the Jira issue link?

We thought this part was also related to link bucket, so we put it in this pr, we should move this part to another PR.

maxinteger (approved):

src/components/ListItemBase/ListItemBase.tsx
maxinteger: @junhao3268 yes, even if it is the same bucket it should be separated PR if it is not related to the original problem which is "fix submenu customized menuitem click crash issue" in this case.

I see you already moved your the title changes. Thank you

maxinteger (changes requested):
sorry I have to add the `validated` label first :)

maxinteger (approved):
Looks good!

PR #525, opened 2024-07-01
feat(ListItemBase): add title attribute to show tooltip in listitem

junhao3268:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description
Add title prop in ListItemBase for link button extra tooltip indicator.
image

# Links

### DIFF
--- a/src/components/ListItemBase/ListItemBase.stories.args.ts
+++ b/src/components/ListItemBase/ListItemBase.stories.args.ts
@@ -1,9 +1,10 @@
-import { commonStyles } from '../../storybook/helper.stories.argtypes';
+import { commonStyles, commonHTMLAttributes } from '../../storybook/helper.stories.argtypes';
 
 import { LIST_ITEM_BASE_CONSTANTS as CONSTANTS } from '.';
 
 export default {
   ...commonStyles,
+  ...commonHTMLAttributes,
   isDisabled: {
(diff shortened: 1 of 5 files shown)


maxinteger (changes requested):
@junhao3268 based on the attached image the whole list is wrong. The list items forced to be links instead of having proper link component inside the list items. So even if I approve this PR the root problem would be not fixed and the Cantina PR would be pushed back.

Also SR use title as a fallback when the component does not have text or label at all. In this case adding title will not make the component more accessible for SR users.

junhao3268:
> @junhao3268 based on the attached image the whole list is wrong. The list items forced to be links instead of having proper link component inside the list items. So even if I approve this PR the root problem would be not fixed and the Cantina PR would be pushed back.
> 
> Also SR use title as a fallback when the component does not have text or label at all. In this case adding title will not make the component more accessible for SR users.



> @junhao3268 based on the attached image the whole list is wrong. The list items forced to be links instead of having proper link component inside the list items. So even if I approve this PR the root problem would be not fixed and the Cantina PR would be pushed back.
> 
> Also SR use title as a fallback when the component does not have text or label at all. In this case adding title will not make the component more accessible for SR users.

maybe we need to investigate more to resolve the root problem, and will discuss with @nataliadelmar later about the title.

PR #530, opened 2024-07-05
feat(reaction-aria-toolbar): changes

torgeadelin:
# Description

> [!CAUTION]
> This PR is introducing **Breaking Changes**
PR With Cantina migration will be raised soon. (to accompany the release of this new changes).

- refactor AriaToolbar
  - introduce new AriaToolbarItem to be able to explicitly describe which elements are the toolbar items (this is useful in situations where we might have Popovers, or Tooltips as immediate children to the AriaToolbar component. Previously it assumes that the immediate children is the actual button, but that's not the case all the time.
  - add direct support for automatic wrap aria toolbar in button group, to facilitate easier styling (out of the box)

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505327

### DIFF
--- a/src/components/AriaToolbar/AriaToolbar.types.ts
+++ b/src/components/AriaToolbar/AriaToolbar.types.ts
@@ -1,4 +1,5 @@
 import { CSSProperties, ReactNode } from 'react';
+import { ButtonGroupProps } from '../ButtonGroup';
 
 export interface Props {
   /**
@@ -43,5 +44,15 @@ export interface Props {
    * This handler is called when tab is pressed by one of the elements
    * within the toolbar.
    */
-  onTabPress?: () => void;
+  onTabPress?: (e: React.KeyboardEvent) => void;
+
+  /**
+   * This prop is used to determine if the toolbar should render as a button group.
+   */
+  shouldRenderAsButtonGroup?: boolean;
+
+  /**
+   * Props to pass to the ButtonGroup component
+   */
+  buttonGroupProps?: ButtonGroupProps;
 }
--- a/src/components/AriaToolbar/AriaToolbar.types.ts
+++ b/src/components/AriaToolbar/AriaToolbar.types.ts
@@ -1,11 +1,15 @@
-import { CSSProperties, ReactNode } from 'react';
+import { CSSProperties, MutableRefObject, ReactElement } from 'react';
 import { ButtonGroupProps } from '../ButtonGroup';
+import { SupportedComponents } from '../ButtonGroup/ButtonGroup.types';
 
 export interface Props {
   /**
    * Child components of this AriaToolbar.
    */
-  children?: ReactNode;
+  children?:
+    | ReactElement<SupportedComponents>
+    | Array<ReactElement<SupportedComponents>>
+    | Array<HTMLElement>;
 
   /**
    * Custom class for overriding this component's CSS.
@@ -55,4 +59,19 @@ export interface Props {
    * Props to pass to the ButtonGroup component
    */
   buttonGroupProps?: ButtonGroupProps;
+
+  /**
+   * Provides context on how many toolbar items are in the aria toolbar. This information is used
+   * to calculate the correct item focus.
+   */
+  ariaToolbarItemsSize: number;
+}
+
+export interface AriaToolbarContextValue {
+  currentFocus?: number;
+  setCurrentFocus?: (newFocus: number) => void;
+  buttonRefs?: MutableRefObject<Record<string, HTMLButtonElement>>;
+  orientation: Props['orientation'];
+  onTabPress: Props['onTabPress'];
+  ariaToolbarItemsSize: number;
 }
--- a/src/components/AriaToolbarItem/AriaToolbarItem.tsx
+++ b/src/components/AriaToolbarItem/AriaToolbarItem.tsx
@@ -1,0 +1,84 @@
+import { Props } from './AriaToolbarItem.types';
+
+import React, { forwardRef, useCallback } from 'react';
+
+import { useKeyboard } from '@react-aria/interactions';
+import { useAriaToolbarContext } from '../AriaToolbar/AriaToolbar.utils';
+
+const AriaToolbarItem = forwardRef<HTMLButtonElement, Props>((props, providedRef) => {
+  const { children, itemIndex } = props;
+
+  const {
+    currentFocus,
+    setCurrentFocus,
+    buttonRefs,
+    orientation,
+    onTabPress,
+    ariaToolbarItemsSize,
+  } = useAriaToolbarContext();
+
+  const { keyboardProps } = useKeyboard({
+    onKeyDown: (e) => {
+      // for the escape key (and other key presses), continue propagation to let Popovers / Modals know that
+      // they should close
+      e.continuePropagation();
+
+      switch (e.key) {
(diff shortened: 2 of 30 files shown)


src/components/Avatar/Avatar.unit.test.tsx
```diff
@@ -5,9 +5,9 @@ import { MAX_INITIALS_SPACE, SIZES, STYLE, AVATAR_COLORS } from './Avatar.consta
 import { AvatarColor, AvatarSize, PresenceType } from './Avatar.types';
 import { mountAndWait } from '../../../test/utils';
 import Icon from '../Icon';
+import { action } from '@storybook/addon-actions';
```
torgeadelin, line 8: This file has been changed due to husky pre commit issue that was merged before....

src/components/AriaToolbarItem/AriaToolbarItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  const getPropsForChildren = useCallback(
+    (child, index) => {
+      return {
+        tabIndex: index === (ariaToolbarContext?.currentFocus || 0) ? 0 : -1,
+        ref: (e: HTMLButtonElement) => {
+          if (ariaToolbarContext.buttonRefs.current) {
+            ariaToolbarContext.buttonRefs.current[index] = e;
+            if (providedRef) {
```
Coread, line 60: question: Have you considered what happens if you were to pass a non-function ref into this component?

torgeadelin: I was initially unsure about merging two non-functional refs but discovered it can be done using `useImperativeHandle` by explicitly specifying which methods to merge. I could consider overriding the ref by default for non func refs or providing a custom merge function with useImperativeHandle, but I feel like is too much work for now. 

After testing this approach in our current use cases, it was working ok.

src/components/AriaToolbarItem/AriaToolbarItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const ariaToolbarContext = useAriaToolbarContext();
+
+  const { keyboardProps } = useKeyboard({
+    onKeyDown: (e) => {
+      // for the escape key (and other key presses), continue propagation to let Popovers / Modals know that
+      // they should close
+      e.continuePropagation();
+
+      switch (e.key) {
+        case ariaToolbarContext?.orientation === 'horizontal' ? 'ArrowLeft' : 'ArrowUp':
```
maxinteger, line 20: We could make AriaToolbar more accessible if we use left/right and up/down for both orientation

[WCAG Tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) also suggest that.

maxinteger (changes requested):

src/components/AriaToolbarItem/AriaToolbarItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+              (ariaToolbarContext?.currentFocus || 0) +
+              1) %
+              ariaToolbarContext?.ariaToolbarItemsSize
+          );
+          break;
+
+        case 'Tab': {
+          if (ariaToolbarContext?.onTabPress) {
+            ariaToolbarContext?.onTabPress(e);
+          }
```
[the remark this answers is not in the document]
torgeadelin: fixed

src/components/AriaToolbar/AriaToolbar.types.ts
```diff
@@ -43,5 +48,30 @@ export interface Props {
...
+  /**
+   * Props to pass to the ButtonGroup component
+   */
+  buttonGroupProps?: ButtonGroupProps;
+
+  /**
+   * Provides context on how many toolbar items are in the aria toolbar. This information is used
+   * to calculate the correct item focus.
+   */
+  ariaToolbarItemsSize: number;
```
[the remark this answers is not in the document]
torgeadelin: I would prefer to keep it as is, because it follows the same pattern we have for List.
If we really want to refactor this, we should do it in both components at the same time, but it feels a little bit too out of scope for my changes.

src/components/AriaToolbarItem/AriaToolbarItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+            ariaToolbarContext.buttonRefs.current[index] = e;
+            if (providedRef) {
+              if (typeof providedRef === 'function') {
+                providedRef(e);
+              }
+            }
+          }
+        },
+        onFocus: (e) => {
+          ariaToolbarContext?.setCurrentFocus(index);
```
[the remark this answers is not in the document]
torgeadelin: addressed

src/components/AriaToolbarItem/AriaToolbarItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+          if (child.props.onFocus) {
+            child.props.onFocus(e);
+          }
```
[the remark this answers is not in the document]
torgeadelin: addressed

src/components/AriaToolbarItem/AriaToolbarItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const getPropsForChildren = useCallback(
+    (child, index) => {
+      return {
+        tabIndex: index === (ariaToolbarContext?.currentFocus || 0) ? 0 : -1,
+        ref: (e: HTMLButtonElement) => {
+          if (ariaToolbarContext.buttonRefs.current) {
+            ariaToolbarContext.buttonRefs.current[index] = e;
+            if (providedRef) {
+              if (typeof providedRef === 'function') {
```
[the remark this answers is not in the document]
torgeadelin: removed

src/components/AriaToolbarItem/AriaToolbarItem.tsx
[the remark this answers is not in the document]
torgeadelin: refactored

src/components/AriaToolbarItem/AriaToolbarItem.tsx
torgeadelin: I have added `useProvidedRef` util which handles both cases.

src/utils/useProvidedRef.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param initialValue
+ * @returns
+ */
+export const useProvidedRef = <T>(
+  providedRef: React.ForwardedRef<T>,
+  initialValue: T = null
+): React.MutableRefObject<T> => {
+  let ref = useRef<T>(initialValue);
+
+  if (providedRef && typeof providedRef !== 'function') {
```
maxinteger, line 16: nitpicking: `providedRef &&` is not necessary because of `typeof`

src/utils/useProvidedRef.ts
```diff
@@ -0,0 +1,29 @@
...
+    ref = providedRef;
+  }
+
+  useLayoutEffect(() => {
+    if (providedRef) {
```
maxinteger, line 21: nitpicking: `if (providedRef) {` is not needed because of `typeof` in the next line

maxinteger (approved):

PR #531, opened 2024-07-06
feat(text): tagName prop to override to tag

Coread:
# Description

Add the prop tagName to allow the overriding of the tag used for the text component. The tagName prop is optional and if not supplied, the default tagName for each type will be used.

# Links

### DIFF
--- a/src/components/Text/Text.stories.tsx
+++ b/src/components/Text/Text.stories.tsx
@@ -52,7 +52,7 @@ TypesWithTagName.parameters = {
       return {
         type,
         children: type,
-        tagName: 'div',
+        tagName: 'small',
       };
     }),
   ],
--- a/src/components/Text/Text.stories.tsx
+++ b/src/components/Text/Text.stories.tsx
@@ -41,10 +41,27 @@ Types.parameters = {
   ],
 };
 
+const TypesWithTagName = MultiTemplate<TextProps>(Text).bind({});
+
+TypesWithTagName.argTypes = { ...argTypes };
+delete TypesWithTagName.argTypes.type;
+
+TypesWithTagName.parameters = {
+  variants: [
+    ...Object.values(TYPES).map((type) => {
+      return {
+        type,
+        children: type,
+        tagName: 'div',
+      };
+    }),
+  ],
+};
+
 const Fonts = MultiTemplate<TextProps>(Text).bind({});
 
 Fonts.argTypes = { ...argTypes };
-delete Types.argTypes.type;
+delete Fonts.argTypes.type;
 
 Fonts.parameters = {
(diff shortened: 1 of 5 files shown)


src/components/Text/Text.types.ts
```diff
@@ -20,6 +20,10 @@ export interface Props {
    * Custom style for overriding this component's CSS.
    */
   style?: CSSProperties;
+  /**
+   * Override the tag used to surround the text
+   */
+  tagName?: keyof JSX.IntrinsicElements;
```
cscheffauer, line 26: for better typing and avoiding mistakes by the consumer, i think u could type it more precise like:
`tagName?: keyof JSX.IntrinsicElements['h1' | 'h2' | 'h3' | ...];`

src/components/Text/Text.unit.test.tsx
```diff
@@ -60,6 +60,21 @@ describe('Text', () => {
     expect(container).toMatchSnapshot();
   });
 
+  it('should match snapshot with type and tagName override', async () => {
+    expect.assertions(1);
+
+    const texts = Object.values(TYPES).map((type, index) => {
+      return (
+        <Text key={index} type={type as FontStyle} tagName="div">
```
cscheffauer, line 68: div should not be allowed as a tagName - i think we should avoid unit tests which are actually breaking the component. instead would be nice to use different "allowed tagNames"

src/components/Text/Text.stories.tsx
```diff
@@ -41,10 +41,27 @@ Types.parameters = {
...
+TypesWithTagName.argTypes = { ...argTypes };
+delete TypesWithTagName.argTypes.type;
+
+TypesWithTagName.parameters = {
+  variants: [
+    ...Object.values(TYPES).map((type) => {
+      return {
+        type,
+        children: type,
+        tagName: 'div',
```
cscheffauer, line 55: same as below - div should not be allowed as a tagName for text, therefore having it here for the story is not recommended.

cscheffauer (changes requested):
some comments on improving it, otherwise looking good

src/components/Text/Text.types.ts
Coread: changed

src/components/Text/Text.unit.test.tsx
Coread: changed, now testing all allowed tag names

src/components/Text/Text.stories.tsx
Coread: updated to use small here

cscheffauer (approved):

PR #541, opened 2024-07-17
fix(eslint): don't read ref.current on render

lismith2-cisco:
# Description

add rule to warn some cases of ref.current being read during render

# Links

https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/Use+useState+not+useRef+for+persistent+generated+ids

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-543350

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,6 +63,13 @@ module.exports = {
     },
   ],
   rules: {
+    'no-restricted-syntax': [
+      'warn',
+      {
+        'selector': "JSXAttribute > JSXExpressionContainer > MemberExpression[property.name='current']",
+        'message': 'Do not read ref.current on render. Either use useState instead of useRef to store the value, or read ref.current in a useEffect or event handler.'
+      }
+    ],
     quotes: ['error', 'single', { allowTemplateLiterals: true, avoidEscape: true }],
     'no-console': ['error', { allow: ['warn'] }],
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -63,6 +63,13 @@ module.exports = {
     },
   ],
   rules: {
+    'no-restricted-syntax': [
+      'warn',
+      {
+        'selector': "JSXAttribute > JSXExpressionContainer > MemberExpression[property.name='current']",
```
maxinteger, line 69: This selector does not work in every case. For example in the `src/components/TextInput/TextInput.tsx` there are 3 `useRef`s:

```ts

  const messageId = useRef(uuidV4());
  const labelId = useRef(uuidV4());
  const clearButtonId = useRef(uuidV4());
```

When the `clearButtonId.current` used directly ESLint give me the warning, but when `messageId.cirrent` is in an expression then there is no warning

image

maxinteger (changes requested):

.eslintrc.js
lismith2-cisco: I know but I don't think it is possible to get eslint to validate this. If you can work out a way, please let me know :) It looks like you can do it in flow ( https://flow.org/en/docs/react/hook-syntax/#preventing-unsafe-mutation ) but I've not seen anything for TypeScript.

PR #542, opened 2024-07-18
fix(a11y): added section to notificationsystem

cscheffauer:
# Description

- Replaced div with section for NotificationSystem and added required attribute ariaLabel to make sure a11y is always met
- Updated stories and unit tests

# Links

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#technical_summary


jor-row (approved):
LGTM, tested with screen reader

PR #551, opened 2024-07-23
Alert only the string

Coread:
# Description

Only the text part of the notification should be read out as an alert by screen readers.

In this PR I've

1. removed the role=alert from all the toastify container
2. added role=alert to the Text in ToastNotification when the content is a string

This requires content other than strings to provide the alert separately.

# Links

### DIFF
--- a/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
+++ b/src/components/NotificationSystem/NotificationSystem.unit.test.tsx
@@ -1,5 +1,5 @@
 import React, { CSSProperties, ReactNode, isValidElement } from 'react';
-import { act, render, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react';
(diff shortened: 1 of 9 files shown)


src/components/NotificationSystem/NotificationSystem.unit.test.tsx
```diff
@@ -354,22 +375,25 @@ describe('<NotificationSystem />', () => {
...
-        NotificationSystem.dismiss(toastId);
-      });
+       NotificationSystem.dismiss(toastId);
+      }); 
+      jest.advanceTimersByTime(1000);
+      fireEvent.animationEnd(screen.getByText(textContent));
 
       // check if toast got removed and the toast is not active anymore
-      await waitForElementToBeRemoved(() => screen.queryByRole('alert'));
+      await waitForElementToBeRemoved(screen.getByText(textContent));
```
cscheffauer, line 386: this always need to be queryByRole, otherwise its not working correctly, see here:
https://testing-library.com/docs/guide-disappearance/#waiting-for-disappearance

PR #552, opened 2024-07-26
feat(aria-toolbar): pass down props + forward ref

torgeadelin:
# Description
- pass rest props on button group
- add forward ref to add reaction button
# Links


src/components/ButtonGroup/ButtonGroup.tsx
gabrielchl, line 1: should we add forwardRef to this as well while we're at it?

torgeadelin: button group usually wraps around multiple buttons, and forwarding the ref doesn't really make sense in this case. (1 ref for multiple buttons / 1 ref -> many buttons)

PR #553, opened 2024-07-29
feat(aria-toolbar-item): pass props to children

gabrielchl:
# Description

pass props to children

# Links


src/components/AriaToolbarItem/AriaToolbarItem.unit.test.tsx
```diff
@@ -35,5 +35,23 @@ describe('<AriaToolbarItem />', () => {
     });
   });
 
+  describe('behaviour', () => {
```
torgeadelin, line 38: not sure this is the appropriate suite for this test, and we should use valid props

torgeadelin (changes requested):
looking good, but 1 small comment

torgeadelin (approved):

PR #556, opened 2024-07-30
fix(a11y): checkbox id duplicate

cscheffauer:
# Description

- According to a11y duplicate id's should be avoided. At the moment the id is passed to the label and input. This fix will make sure to pass through the id only to the label, not the input itself to avoid duplication. This is also tested by the a11y tests in Storybook (not failing anymore with this fix).


jor-row (approved):
LGTM

PR #559, opened 2024-07-31
feat(accordion): pass props down to header button

gabrielchl:
# Description

# Links

### DIFF
--- a/src/components/Accordion/Accordion.style.scss
+++ b/src/components/Accordion/Accordion.style.scss
@@ -1,21 +1,21 @@
 .md-accordion-wrapper {
-    .md-accordion-header-row {
-        display: grid;
-        grid-template-columns: 1fr max-content;
-        align-items: center;
-        gap: 8px;
-    }
+  .md-accordion-header-row {
+    display: grid;
+    grid-template-columns: 1fr max-content;
+    align-items: center;
+    gap: 0.5rem;
+  }
 
-    .md-accordion-header-heading {
-        width: 100%;
-    }
+  .md-accordion-header-heading {
+    width: 100%;
+  }
 
-    .md-accordion-header-button {
-        display: flex;
-        gap: 8px;
-        align-items: center;
-        background: none;
-        border: none;
-        justify-content: flex-start;
-    }
+  .md-accordion-header-button {
+    display: flex;
+    gap: 0.5rem;
+    align-items: center;
+    background: none;
+    border: none;
+    justify-content: flex-start;
+  }
 }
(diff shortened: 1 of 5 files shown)


src/components/Accordion/Accordion.types.ts
```diff
@@ -40,4 +41,9 @@ export interface Props {
    * Hierarchical level of the accordion within a structure.
    */
   ariaLevel: number;
+
+  /**
+   * Extra props to be added to the header button.
+   */
+  buttonProps?: ButtonSimpleProps;
```
maxinteger, line 48: probably you want to `Omit` some of the button props which are already used, for example `aria-expanded` or `aria-controls`

src/components/Accordion/Accordion.tsx
```diff
@@ -48,6 +49,7 @@ const Accordion: FC<Props> = (props: Props) => {
             aria-expanded={expanded}
             aria-controls={panelId}
             onPress={onClick}
+            {...buttonProps}
```
maxinteger, line 52: this could overwrite the `onClick` from the previous line, either you exclude the `onClick` or merge yours with the user specified one. In other components we used the later solution and merged the 2 handler

maxinteger (changes requested):

src/components/Accordion/Accordion.types.ts
```diff
@@ -1,4 +1,23 @@
...
+
+type ButtonSimpleCustomPropKeys =
+  | 'className'
+  | 'id'
+  | 'aria-expanded'
+  | 'aria-controls'
+  | 'onPress';
+
+// props applied on the button, defined by the accordion component
+export type ButtonCustomProps = Required<Pick<ButtonSimpleProps, ButtonSimpleCustomPropKeys>> & {
```
maxinteger, line 12: Now it is a bit overcomplicated because inside the component we have full control on the values and we know what we want to pass to the button which has a type safe interface anyway. 

we need `ButtonAllowedProps` to make it clear to the consumer what props are valid

You can keep this change.

src/components/Accordion/Accordion.style.scss
```diff
@@ -1,21 +1,21 @@
...
-        display: grid;
-        grid-template-columns: 1fr max-content;
-        align-items: center;
-        gap: 8px;
-    }
+  .md-accordion-header-row {
+    display: grid;
+    grid-template-columns: 1fr max-content;
+    align-items: center;
+    gap: 0.5rem;
```
maxinteger, line 6: 🙇 thanks for changing it to `ram`

maxinteger (approved):

src/components/Accordion/Accordion.style.scss
gabrielchl: Untitled

torgeadelin (approved):

PR #561, opened 2024-07-31
fix(toggle buttons): add ariaStateKey prop to determine whether aria-pressed or aria-expanded is used for selected state

lismith2-cisco:
# Description

In ButtonCircleToggle and ButtonPillToggle, add ariaStateKey prop to determine whether aria-pressed or aria-expanded is used for selected state

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505240 (see comment)

### DIFF
--- a/src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
+++ b/src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
@@ -12,6 +12,7 @@ import { chain } from '@react-aria/utils';
 
 const ButtonCircleToggle = forwardRef((props: Props, providedRef: RefObject<HTMLButtonElement>) => {
   const {
+    ariaStateKey = DEFAULTS.ARIA_STATE_KEY,
     children,
     className,
     ghost = DEFAULTS.GHOST,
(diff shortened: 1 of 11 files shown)


src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -40,7 +41,7 @@ const ButtonCircleToggle = forwardRef((props: Props, providedRef: RefObject<HTML
       size={size as ButtonCircleSize}
       data-selected={state.isSelected || DEFAULTS.SELECTED}
       onPress={chain(state.toggle, onPress)}
-      aria-pressed={state.isSelected}
+      {...{ [ariaStateKey]: state.isSelected }}
```
torgeadelin, line 44: issue: you should be able to pass in `aria-pressed` or `aria-expanded` manually as a prop, the button should already have that prop type (it inherits button DOM props)

torgeadelin: i take it back, i see that we can't have both aria-pressed and aria-expanded, and we're already explicitly using aria-pressed... This feels a bit odd, because the component title is clearly a toggle..?? which means for sure it needs an aria-pressed state, but also controls something that is expanded/collapsed.

But since the A11y team wants us to remove it, then I guess this implementation is good.

torgeadelin (approved):

src/storybook/helper.stories.argtypes.ts
```diff
@@ -708,6 +708,19 @@ const commonAriaButtonToggle = {
       },
     },
   },
+  ariaStateKey: {
```
torgeadelin, line 711: nitpick: can you add the same for the button circle toggle

PR #566, opened 2024-08-05
feat: tree component

maxinteger:
# Description

Add new Tree and TreeBaseNode components

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-548345


src/components/TreeNodeBase/TreeNodeBase.types.ts
```diff
@@ -0,0 +1,100 @@
...
+  shape?: 'rectangle' | 'isPilled';
+
+  /**
+   * Determines whether the focus for this item should be inset or outset
+   * @default false
+   */
+  shouldItemFocusBeInset?: boolean;
+
+  /**
+   * Determines if the tree node is interactive (usually used as a header when false)
```
nataliadelmar, line 83: What do you mean with '(usually used as a header when false)' ? If the node is not interactive, it is probably a leaf node right?

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,224 @@
...
+    isSelected,
+    style,
+    interactive = DEFAULTS.INTERACTIVE,
+    onPress,
+    lang,
+    externalToggleFn,
+    ...rest
+  } = props;
+  if (!nodeId) {
+    throw new Error('TreeNodeBase: nodeId prop is required.');
```
nataliadelmar, line 47: I think we do not throw error in Tree components. We normally console.warn(message), like we do here 
image

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,224 @@
...
+    ref = providedRef;
+  }
+
+  // When used in a popover, the ref will be a callback.
+  // We need to update this callback ref, so the popover
+  // knows about the dom element, but we can't use the callback
+  // ref directly because we want it to be a useRef style ref
+  // We useLayoutEffect so that it happens in time for tippy
+  // to use the ref when adding event handlers
+  useLayoutEffect(() => {
```
nataliadelmar, line 69: Is this for the case of having a tree component inside a popover? Do we have a use case for that? I can only think of calendar item popover, but curious if there is any other

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,224 @@
...
+            {start}
+            {middle}
+            {end}
+          </>
+        );
+      } else if (verifyTypes(children, Text)) {
+        content = children;
+      } else {
+        console.warn(
+          'TreeNodeBase: this component can only receive ListItemBaseSection as children.'
```
nataliadelmar, line 98: can only receive ListItemBaseSection or Text as children

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,224 @@
...
+        );
```
nataliadelmar, line 99: So, if children is an array, the only permitted cases are:
`[<ListItemBaseSection />, <ListItemBaseSection/ >, <ListItemBaseSection>]`
`[<ListItemBaseSection />, <ListItemBaseSection/ >]`
`[<Text />, <Text/ >, <Text>]`
`[<Text />, <Text/ >]`
 
Why do we want to make this that restrictive? why not simply rendering the children inside the node?

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,224 @@
...
+      }
+    },
+  };
+
+  /**
+   * Focus management
+   */
+  const shouldFocusOnPress = treeContext?.shouldFocusOnPress || false;
+  const shouldItemFocusBeInset =
+    treeContext?.shouldItemFocusBeInset || DEFAULTS.SHOULD_ITEM_FOCUS_BE_INSET;
```
nataliadelmar, line 139: On TreeNodeBase.tsx I see there is a prop called shouldItemFocusBeInset. Why are we not using this for this line too?

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  const treeNodeTabIndex = getTreeNodeBaseTabIndex(nodeId, treeContext.activeNodeId, interactive);
```
nataliadelmar, line 141: shouldnt this const be called treeNodeBaseTabIndex?

src/components/Tree/test.utils.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Create a tree node with the given id, isOpen and children.
+ *
+ * @param id The id of the node
+ * @param isOpenByDefault
+ * @param children List of child nodes
+ */
+export const createTreeNode = (
+  id: TreeNodeId,
+  isOpenByDefault: boolean = undefined,
```
nataliadelmar, line 12: isn't this the same as to say `isOpenByDefault?: boolean`

nataliadelmar: do we need the default to undefined?

src/components/TreeNodeBase/TreeNodeBase.types.ts
maxinteger: good point. This was a leftover from the ListItemBase, but we do not really need this

I removed the `interactive` and `isDisabled` props for now because if we can disable nodes then it will effect on the navigation. 

Let's postpone this functionality until we really need it

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,335 @@
...
+
+export const useTreeContext = (): TreeContextValue => {
+  const value = useContext(TreeContext);
+  if (!value) {
+    throw new Error('useTreeContext hook used without TreeContext!');
+  }
+  return value;
+};
+
+const TREE_NAVIGATION_ITERATION_LIMIT = 1_024;
```
nataliadelmar, line 22: this should be in constants file

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,335 @@
...
+  return Array.from(tree.values()).find((node) => !node.parent)?.id;
+};
+
+/**
+ * Toggle the open state of a node
+ *
+ * @param tree
+ * @param nodeId
+ */
+export const toggleActiveNode = (tree: FlatTree, nodeId: TreeNodeId): TreeNodeId => {
```
nataliadelmar, line 38: the name of this util is confusing, it doesn't get the active node, it uses the passed nodeid and toggles it. So maybe a name like `toggleNode` would be better

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,335 @@
...
+  const node = tree.get(nodeId);
+  if (node) {
```
nataliadelmar, line 40: should we log any warning if the node doesn't exist?

src/components/TreeNodeBase/TreeNodeBase.tsx
maxinteger: Since this solution works in list and you already mentioned a possible use case, I keep this code

src/components/TreeNodeBase/TreeNodeBase.tsx
maxinteger: I re-phrased the warning.

Valid children are:
- Single, eg.: `<div/>`, `</>` (React.Frament)
- Multiple eg: `[<Text />, <Text/ >]

The message is misleading, `[<ListItemBaseSection />, <ListItemBaseSection/ >]` is not allowed, instead you have to use: `[<ListItemBaseSection />]` and put the multiple child into `ListItemBaseSection`

src/components/TreeNodeBase/TreeNodeBase.tsx
maxinteger: Since it is a private variable of the component, the prefix is not needed. I reduced the name to `tabIndex` only

src/utils/navigation.ts
```diff
@@ -0,0 +1,38 @@
+const FOCUSABLE_ELEMENT_SELECTORS =
+  'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex=""]';
+const FOCUSABLE_AND_TABBABLE_ONLY_ELEMENT_SELECTORS =
+  'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"]';
```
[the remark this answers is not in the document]
maxinteger: Nice catch!

src/components/TreeNodeBase/TreeNodeBase.tsx
maxinteger: Good point, we do not need the `shouldItemFocusBeInset` on the TreeNodeBase, removed
(also renamed it in the Tree component's porps)

src/components/Tree/test.utils.ts
maxinteger: it is the same as `isOpenByDefault?: boolean` but it is more explicit this way, and following the patter in the next line `children: Array<TreeNode> = []`

src/components/Tree/Tree.utils.ts
maxinteger: Actually we do not event need this function

src/components/Tree/Tree.utils.ts
maxinteger: I added a check to `getNextActiveNode`, not sure about the rest of the case. Some time is valid to have an undefined result for example when we reach the root which has not parent

src/components/TreeNodeBase/TreeNodeBase.stories.docs.mdx
```diff
@@ -0,0 +1,60 @@
+[INTERNAL] `<TreeNodeBase />` component should be the base of any Tree Item component that the
+Momentum Library requires. This component acts like a skeleton that helps you develop more complex TreeNodes.
+It manages many features of the tree, such as focus, keyboard navigation, open/close nodes, etc.
+
+`<TreeNodeBase />` uses the TreeContext from the `<Tree />` component to manage the state of the tree.
+If the `<TreeNodeBase />` can not access ot the context, it will throw an error.
+
+`<TreeNodeBase />` allows to navigate into the node itself when there are interactive elements inside it.
+After the node selected the user can use the `Tab` key to select the first, second, etc. interactive element.
```
nataliadelmar, line 9: 'After the node is **focused** (not selected), the user can ...'

src/components/TreeNodeBase/TreeNodeBase.stories.docs.mdx
```diff
@@ -0,0 +1,60 @@
...
+When the last interactable element is selected, the next tab will move the focus to the next interactable
```
nataliadelmar, line 10: 'When the last interactable element is **focused**, the next tab will move the focus to the next interactable...'

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,169 @@
...
+      const isRoot = parent === undefined;
+      const isSelected = activeNodeId === node.id;
+
+      // tabindex depends on the treeNodeBase params as well
+      const props = {
+        id: node.id.toString(),
+        'aria-setsize': isRoot ? 1 : parent.children.length,
+        'aria-level': node.level + 1,
+        'aria-posinset': node.index + 1,
+        'aria-selected': isSelected,
```
nataliadelmar, line 62: We shouldn't set aria-selected if activeNode === node.id (that is, if the node is simply focused).

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,188 @@
...
+  treeStructure: TreeRoot;
+
+  /**
+   * Tree structure can be rendered 2 ways in the DOM:
+   * 1) Nested: The tree is rendered as a nested list where the structure reflect the semantic structure of the tree
+   * 2) Flat: The tree is rendered as a single level list where the DOM does not reflect the semantic structure of the tree, and
+   *    we need to provide additional aria attributes to re-build it for the accessibility tree.
+   *    Virtualized trees are usually rendered flat.
+   *
+   * @see {@link https://codepen.io/maxinteger/pen/zYVNabV Example of a nested and flat tree}
```
torgeadelin, line 128: question: not sure if we're allowed to do this? (license issue)

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,188 @@
...
+   * @default true
+   */
+  isRenderedFlat?: boolean;
+
+  /**
+   * Determines if the tree root should be excluded from the tree navigation.
+   *
+   * In many case we want to hide the root of the tree, for example when the tree used for grouping some list items.
+   *
+   * Note: it does not change the visibility of the root node.
```
torgeadelin, line 138: suggestion: add another comment saying this only excludes the tree root from keyboard navigation

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * Get additional attributes of the tree node group.
+   * @param id unique identifier of the tree node
+   */
+  getNodeGroupProps: (id: TreeNodeId) => Partial<HTMLAttributes<HTMLElement>>;
+  /**
+   * Get the details of the tree node.
+   * @param id unique identifier of the tree node
+   */
+  getNodeDetails: (id: TreeNodeId) => FlatTreeNode;
```
torgeadelin, line 166: question: does this work for both flat and non-flat trees? do we always return flat tree node?

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,188 @@
...
+   * Add an external toggle function to the tree node.
+   * @param id unique identifier of the tree node
+   * @param toggleFunction Toggle function from other tree rendered
+   */
+  addExternalToggleFunction: (id: TreeNodeId, toggleFunction: ToggleTreeNode) => void;
+  /**
+   * Remove an external toggle function from the tree node.
+   * @param id unique identifier of the tree node
+   */
+  removeExternalToggleFunction: (id: TreeNodeId) => void;
```
torgeadelin, line 187: question: I don't understand the purpose of these 2. Are these just callbacks when we remove and add nodes to the tree or?

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+const findNextTreeNode = (tree: FlatTree, activeNodeId: TreeNodeId): TreeNodeId => {
+  let current = tree.get(activeNodeId);
+
+  // Step into an open node
+  if (!current.isLeaf && current.isOpen) {
+    return current.children[0];
+  }
+
+  // Otherwise, find the next sibling
+  for (let counter = 0; ; counter++) {
```
torgeadelin, line 45: issue: there's an extra `;`.

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+  if (current.index === 0 && excludeRootNode && current.parent === getFlatTreeRootId(tree))
+    return activeNodeId;
+
+  // Move one level up
+  if (current.index === 0) return current.parent;
+
+  // Find the previous sibling
+  let next = tree.get(tree.get(current.parent).children[current.index - 1]);
+
+  for (let counter = 0; next; counter++) {
```
torgeadelin, line 92: question: is `next` in the middle the condition for the loop, meaning that as long as next is defined?

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+  let current = tree.get(activeNodeId);
+
+  // Step into an open node
+  if (!current.isLeaf && current.isOpen) {
+    return current.children[0];
+  }
+
+  // Otherwise, find the next sibling
+  for (let counter = 0; ; counter++) {
+    if (counter > TREE_NAVIGATION_ITERATION_LIMIT) {
```
torgeadelin, line 46: question: we do support 1K participants, and if we're gonna use this in the participants list, we might reach the limit.. but not sure how realistic that is.

Any particular reason you added it or?

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+  return Array.from(tree.values()).find((node) => !node.parent)?.id;
+};
+
+/**
+ * Find the next active tree node based on the current active node
+ * @param tree
+ * @param activeNodeId
+ */
+const findNextTreeNode = (tree: FlatTree, activeNodeId: TreeNodeId): TreeNodeId => {
+  let current = tree.get(activeNodeId);
```
torgeadelin, line 37: question: what is activeNodeId? is it the currently selected node in the tree?

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+ * @see {@link https://www.w3.org/WAI/ARIA/apg/patterns/treeview/ WCAG Tree Pattern}
+ * @see {@link https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/ WCAG Directory Tree example}
+ *
+ * @param tree
+ * @param activeNodeId Current active tree node descriptor
+ * @param keyCode Arrow key code
+ * @param excludeRoot
+ * @param toggleTreeNode
+ */
+export const getNextActiveNode = (
```
torgeadelin, line 232: suggestion: maybe rename this func, there's another function above `findNextTreeNode `, and in the comment it says ` * Find the next active tree node based on the current active node`.  I can see that the imlpementation of this one is actually different, it actually handles the keyboard nav.

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+  // Move up one level if it's closed
+  if (current.parent) {
+    return current.parent;
+  }
+  // Otherwise, do nothing
+  return activeNodeId;
+};
+
+/**
+ * Travers the tree and convert it to a map between the node id and the node
```
gabrielchl, line 167:
```suggestion
 * Traverse the tree and convert it to a map between the node id and the node
```

src/components/ListItemBase/ListItemBase.types.ts
```diff
@@ -76,12 +76,6 @@ export interface Props extends PressEvents, ContextMenu {
    */
   role?: string;
 
-  /**
-   * Determines wether the focus for this item should be inset or outset
-   * @default false
-   */
-  shouldItemFocusBeInset?: boolean;
```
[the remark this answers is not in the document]
maxinteger: the prop never used in the ListItemBase we always get the value from the list context. I just checked in Cantina this prop appear only once on the List component, so I can not see how is this break the code

src/components/Tree/Tree.stories.args.ts
```diff
@@ -0,0 +1,30 @@
+import { commonStyles } from '../../storybook/helper.stories.argtypes';
+
+export default {
+  ...commonStyles,
+  shouldNodeFocusBeInset: {
```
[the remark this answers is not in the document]
maxinteger: It is used in the TreeNodeBase same way as we use it in the List/ListItemBase

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,169 @@
...
+import { useKeyboard } from '@react-aria/interactions';
+import { isFocusableNode } from '../../utils/navigation';
+
+const Tree: FC<Props> = (props: Props) => {
+  const {
+    className,
+    id,
+    style,
+    children,
+    shouldNodeFocusBeInset,
```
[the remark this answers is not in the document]
maxinteger: It is passed to the tree context and then used in the TreeNodeBase

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,169 @@
...
+    treeStructure,
+    isRenderedFlat = DEFAULTS.IS_RENDERED_FLAT,
+    excludeTreeRoot = DEFAULTS.EXCLUDE_TREE_ROOT,
+    ...rest
+  } = props;
+
+  if (excludeTreeRoot && !treeStructure.children.length) {
+    throw new Error('Tree must have at least one child when excludeTreeRoot is true');
```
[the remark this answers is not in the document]
maxinteger: I can change that, but then it will fail a few lines later with a more generic error, because it tries to read an `id` of an undefined child.

I would prefer the more descriptive error then the generic one

src/components/Tree/Tree.types.ts
maxinteger: So, what do you recommend? I can create a GitHub gist, or I can just copy the HTML and add it to the story docs, or I can add here as well

BTW the link is pointing to my CodePan :)

src/components/Tree/Tree.types.ts
maxinteger: It is already in the first line of the comment, but I added the "keyboard" word to make it more accurate

src/components/Tree/Tree.types.ts
maxinteger: Yes, this can be misleading. There are 2 things
1) The tree can be rendered in the DOM flat or nested (up to the consumer)
2) Our Tree component represents the tree data structure as a "FlatTree", it is more efficient to navigate in it when you need to find the next active node. (consumer has no control over this)

These are un-related to each other. Most of the time the internal representation is completely hidden for the consumer, except here. 

Let me know if you have better name for this type to make it less confusing.

src/components/Tree/Tree.utils.ts
maxinteger: `findNextTreeNode` is just a utility function which is not exported and used only in the `getNextActiveNode` function. I can update the docs to make it more descriptive, but I am also open for name suggestions

src/components/Tree/Tree.utils.ts
maxinteger: Yes, the `activeNodeIs` is the id of the currently focused node

src/components/Tree/Tree.utils.ts
maxinteger: `for` loop must have two `;`. This is an unconditional loop, where we decided about the exit condition in the body. I could use the `while(true)`, but then I have to define the counter in a separated line and the update as well, that mean 2 additional lines!!! :D

src/components/Tree/Tree.utils.ts
maxinteger: yes

src/components/ListItemBase/ListItemBase.types.ts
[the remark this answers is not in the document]
torgeadelin: Aaa gotcha. I guess it's safe to remove in that case. I haven't realised it was propagating through context.

src/components/Tree/Tree.tsx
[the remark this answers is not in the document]
torgeadelin: Ok, we can leave the throw I then.

src/components/Tree/Tree.types.ts
torgeadelin: Maybe gist? I know it points to your own codepad, but ai think the issue still applies regardless.

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,183 @@
...
+   * This is needed for virtualized tree
+   */
+  shouldNodeFocusBeInset?: boolean;
+
+  /**
+   * The initial tree structure
+   *
+   * It is used to build an internal tree for navigation and follow open/close states.
+   */
+  treeStructure: TreeRoot;
```
nataliadelmar, line 120: TreeRoot is of type TreeNode, which does not have a flat structure. What is we want a flat structure?

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * Note: it does not change the visibility of the root node.
+   * @default true
+   */
+  excludeTreeRoot?: boolean;
+
+  /**
+   * Toggle open/close state of the tree node.
+   */
+  setVirtualTreeNodeOpenState?: (id: TreeNodeId, isOpen: boolean) => void;
```
nataliadelmar, line 147: you have a type set for this called ToggleTreeNode

src/components/Tree/Tree.types.ts
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * Set the active node id in the tree.
+   * @param id unique identifier of the tree node
+   */
+  setActiveNodeId: (id: TreeNodeId) => void;
+  /**
+   * Toggle the isOpen state of the tree node.
+   * @param id unique identifier of the tree node
+   */
+  toggleTreeNode: (id: TreeNodeId) => void;
```
nataliadelmar, line 182: we have a type called ToggleTreeNode that is a function with id and isOpen as params. It is confusing that one toggleTreeNode only needs id and the other needs both id and isOpen stated, don't you think?

nataliadelmar (approved):

src/components/Tree/Tree.types.ts
maxinteger: flat tree is the internal representation of the tree inside the Tree component, it is more efficient to use the flattened tree for navigation then the normal nested tree.

Please note, it is independent from the `isRenderedFlat` because it is about how the consumer built the tree in the DOM

The consumer should not know about this internal representation.

src/components/Tree/Tree.types.ts
maxinteger: Unfortunately there are different. The `isOpen` parameter is option in the `ToggleTreeNode` and required in `setVirtualTreeNodeOpenState`

src/components/Tree/Tree.types.ts
torgeadelin: We're still using codepen, but I'm not going to block it for that.

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,355 @@
...
+
+  // Step into an open node
+  if (!current.isLeaf && current.isOpen) {
+    return current.children[0];
+  }
+
+  const loopCheck = new Set<TreeNodeId>();
+
+  // Otherwise, find the next sibling
+  for (;;) {
```
torgeadelin, line 51: suggestion: `while (true) {}`

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,355 @@
...
+  tree: TreeIdNodeMap,
+  activeNodeId: TreeNodeId,
+  toggleTreeNode: ToggleTreeNode
+): TreeNodeId => {
+  const current = tree.get(activeNodeId);
+
+  if (!current.isLeaf) {
+    if (!current.isOpen) {
+      // Open it if it's closed
+
```
torgeadelin, line 133: nitpick: unnecessary new line

src/components/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,355 @@
...
+ * Toggle the open/close state of the tree node.
+ *
+ * It also updates the hidden state of all children based on the parent's open state.
+ * And it returns the updated tree without changing the original tree.
+ *
+ * @param id
+ * @param prevTree
+ * @param isOpen
+ */
+export const toggleTreeNodeRecord = (
```
torgeadelin, line 284: question: what is the difference between this and `toggleTreeNode` from https://github.com/momentum-design/momentum-react-v2/blob/44edce188dfbcb694ef449ed8675454a75ef7fb7/src/components/Tree/Tree.tsx#L38

src/components/TreeNodeBase/TreeNodeBase.style.scss
```diff
@@ -0,0 +1,109 @@
...
+  justify-content: space-between;
+  align-items: center;
+  font-size: 1rem;
+  line-height: 1.5rem;
+  color: var(--mds-color-theme-text-primary-normal);
+  transition: background-color 0.2s, color 0.2s, border-color 0.2s;
+  box-sizing: border-box;
+  outline: none;
+
+  &[data-interactive='true'] {
```
torgeadelin, line 13: issue: you're missing interactive prop on the tree node base (you need to follow similar approach as we do in list item base), OR just remove it? Not sure we need `interactive` for tree node... but you never know..

torgeadelin, line 13: issue: I don't see any disabled state

torgeadelin: I've identified both of this issue from the storybook page of this component. I noticed you've used the template for pseudo states (praise!), and they're broken. The reason they are broken are because of the 2 issues abvove.

src/components/TreeNodeBase/TreeNodeBase.tsx
```diff
@@ -0,0 +1,204 @@
...
+        style={style}
+        ref={ref}
+        aria-selected={isSelected}
+        data-nodeid={nodeId}
+        data-size={size}
+        data-padded={isPadded}
+        data-shape={shape}
+        className={classnames(className, STYLE.wrapper, { active: isPressed || isSelected })}
+        lang={lang}
+        {...treeNodePressProps}
```
torgeadelin, line 186: suggestion: we should allow the pass in the `..rest` props to this as long as they're valid tags. Not sure what the best approach for this is.. We normally just pass in ...rest BUT I've noticed a lot in Cantina and mrv2 that a lot of times we pass in invalid DOM attributes to the HMTL elements, which result in warnings in unit test output and console.logs in browser.

torgeadelin: This is helpful if we need to pass any aria attributes from the consumer side.

torgeadelin (commented):
some small issues still, but looking great otherwise

torgeadelin (changes requested):
some small issues still, but looking great otherwise

src/components/Tree/Tree.types.ts
maxinteger: Removed the link and I added the examples to the storybook docs

src/components/Tree/Tree.types.ts
maxinteger: FYI, I renamed this and some other types to make it less confusing for devs and consumers

src/components/Tree/Tree.utils.ts
maxinteger: ESLint does not like it, `no-constant-condition`, I do it anyway, `while(true)` is more readable then `for(;;)`

src/components/Tree/Tree.utils.ts
maxinteger: in `toggleTreeNodeRecord` the `isOpen` parameter required while in the `toggleTreeNode` is optional. Also, `toggleTreeNodeRecord` is used internally meantime `toggleTreeNode` is part of the Tree context API

src/components/TreeNodeBase/TreeNodeBase.style.scss
maxinteger: Nice catch, Tree node is interactive by default, at least for now. I removed the `&[data-interactive='true'] {`, but kept the rest.

maxinteger: yes, there is no disabled state fro the tree node at the moment. We can add it when it is needed.

src/components/TreeNodeBase/TreeNodeBase.tsx
maxinteger: fixed

torgeadelin (approved):

PR #567, opened 2024-08-06
fix(a11y): aria changes in searchinput

peter7cole:
# Description

- adds aria-controls, aria-expanded (isExpanded), isCombobox, and 'combobox' role to SearchInput
- three warnings for usage of isCombobox, controls, and isExpanded
- types and unit testing
- updated snapshots accordingly

### DIFF
--- a/src/components/SearchInput/SearchInput.tsx
+++ b/src/components/SearchInput/SearchInput.tsx
@@ -34,18 +34,18 @@ const SearchInput = (props: Props, ref: RefObject<HTMLInputElement>): ReactEleme
     label,
     isDisabled,
     height = DEFAULTS.HEIGHT,
-    controls,
+    ariaControls,
     isCombobox = DEFAULTS.ISCOMBOBOX,
-    isExpanded,
+    isComboboxExpanded,
   } = props;
 
-  if (isCombobox && isExpanded === undefined) {
+  if (isCombobox && isComboboxExpanded === undefined) {
     console.warn(WARNINGS.ISCOMBOBOX_1_ISEXPANDED_0);
   }
-  if (!isCombobox && !!controls) {
+  if (!isCombobox && !!ariaControls) {
     console.warn(WARNINGS.ISCOMBOBOX_0_CONTROLS_1);
   }
-  if (!isCombobox && typeof isExpanded === 'boolean') {
+  if (!isCombobox && typeof isComboboxExpanded === 'boolean') {
     console.warn(WARNINGS.ISCOMBOBOX_0_ISEXPANDED_1);
   }
 
@@ -87,8 +87,8 @@ const SearchInput = (props: Props, ref: RefObject<HTMLInputElement>): ReactEleme
 
   const inputElement = isCombobox ? (
     <input
-      aria-controls={controls}
-      aria-expanded={isExpanded}
+      aria-controls={ariaControls}
+      aria-expanded={isComboboxExpanded}
       role={ARIA_ROLES.COMBOBOX}
       {...inputProps}
       {...focusProps}
--- a/src/components/SearchInput/SearchInput.tsx
+++ b/src/components/SearchInput/SearchInput.tsx
@@ -4,7 +4,13 @@ import React, { ReactElement, useRef, RefObject, forwardRef } from 'react';
 import classnames from 'classnames';
 
 import ButtonSimple from '../ButtonSimple';
-import { STYLE, DEFAULTS, ICON_HEIGHT_MAPPING, ARIA_ROLES } from './SearchInput.constants';
+import {
+  STYLE,
+  DEFAULTS,
+  ICON_HEIGHT_MAPPING,
+  ARIA_ROLES,
+  WARNINGS,
+} from './SearchInput.constants';
 import { Props } from './SearchInput.types';
 import './SearchInput.style.scss';
 import { useSearchField } from '@react-aria/searchfield';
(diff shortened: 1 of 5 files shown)


peter7cole:
I'm inclined to leave the formatting fixes in the test file, but am open to reverting them.

src/components/SearchInput/SearchInput.unit.test.tsx
```diff
@@ -186,17 +214,49 @@ describe('<SearchInput />', () => {
...
+      const element = (
+        await mountAndWait(
+          <SearchInput aria-expanded={false} searching={true} clearButtonAriaLabel="Clear" />
+        )
+      )
+        .find(SearchInput)
+        .getDOMNode();
+
+      expect(element.getAttribute('aria-expanded')).toBeFalsy();
+    });
```
peter7cole, line 252: this is the only real test change/addition

src/components/SearchInput/SearchInput.unit.test.tsx
```diff
@@ -186,17 +214,107 @@ describe('<SearchInput />', () => {
...
+      expect.assertions(1);
+
+      const element = (
+        await mountAndWait(<SearchInput searching={true} clearButtonAriaLabel="Clear" />)
+      )
+        .find(SearchInput)
+        .getDOMNode();
+
+      expect(element.getAttribute('role')).toBe('searchbox');
+    });
```
peter7cole, line 310: the only new tests

src/components/SearchInput/SearchInput.unit.test.tsx
```diff
@@ -186,17 +214,107 @@ describe('<SearchInput />', () => {
...
```
peter7cole, line 310: the only actual test additions

src/components/SearchInput/SearchInput.types.ts
```diff
@@ -90,4 +90,20 @@ export interface Props extends AriaSearchFieldProps {
    * @default 32
    */
   height?: SearchFieldHeight;
+
+  /**
+   * aria-controls attribute, the HTML ID of the element that this SearchInput relates to
+   */
+  ariaControls?: string;
```
chburket, line 97: I know that is some cases, we allow all html attrs to be specified (...rest  - see List.tsx). Here it looks like we are allowing aria-controls specifically and we are camel-casing it.  Is there a standard way of doing this?

chburket (commented):
just a comment about convention

src/components/SearchInput/SearchInput.tsx
```diff
@@ -75,6 +88,9 @@ const SearchInput = (props: Props, ref: RefObject<HTMLInputElement>): ReactEleme
       data-focus={isFocused}
       data-height={height}
       ref={containerRef}
+      aria-controls={controls}
```
robstax, line 91: couple things:
1. my understanding is that all of these attributes need to be on the `<input/>` itself on line 113.
2. we *only* want to add `aria-controls` and `aria-expanded` IF AND ONLY IF `isCombobox` is true. if it's not a combobox (`isCombobox` is false) we must not set the `aria-expanded` attribute at all. make sense?

robstax (changes requested):

src/components/SearchInput/SearchInput.unit.test.tsx.snap
```diff
@@ -11,6 +11,7 @@ exports[`<SearchInput /> snapshot should match snapshot 1`] = `
       data-focus={false}
       data-height={32}
       onClick={[Function]}
+      role="searchbox"
```
chrisadubois, line 14: Shouldn't there be a snapshot with combobox=true?

robstax (approved):

src/components/SearchInput/SearchInput.tsx
```diff
@@ -65,6 +85,19 @@ const SearchInput = (props: Props, ref: RefObject<HTMLInputElement>): ReactEleme
     }
   };
 
+  const inputElement = isCombobox ? (
+    <input
+      aria-controls={controls}
+      aria-expanded={isExpanded}
+      role={ARIA_ROLES.COMBOBOX}
+      {...inputProps}
```
nataliadelmar, line 93: do inputProps return any role? if yes, it might overwrite our `role={ARIA_ROLES.COMBOBOX}`

src/components/SearchInput/SearchInput.types.ts
```diff
@@ -90,4 +90,21 @@ export interface Props extends AriaSearchFieldProps {
    * @default 32
    */
   height?: SearchFieldHeight;
+
+  /**
+   * While isCombobox=true, the HTML ID of the element that this SearchInput relates to,
+   * used for the aria-controls attribute
+   */
+  controls?: string;
```
nataliadelmar, line 98: when passing props that will be using in aria-props, we are normally converting the aria-prop to camelCase.
`ariaControls?: string`
That way it is even more obvious that what are we intending to do with this value

src/components/SearchInput/SearchInput.types.ts
```diff
@@ -90,4 +90,21 @@ export interface Props extends AriaSearchFieldProps {
...
+  /**
+   * Whether SearchInput is being used with a role of combobox or not
+   */
+  isCombobox?: boolean;
+
+  /**
+   * While isCombobox=true, whether the combobox is expanded/collapsed,
+   * used for the aria-expanded attribute
+   */
+  isExpanded?: boolean;
```
nataliadelmar, line 109: In the expanded case, I am happy to leave it as isExpanded. However, isExpanded only works for the Combobox case, not any other case like a Tooltip or simple explanation popover triggering from the Input. Because of this, it might be a good idea to change this prop to isComboboxExpanded?: boolean

src/components/SearchInput/SearchInput.tsx
peter7cole: inputProps doesn't contain a role value and this works as is, or we could cover the future possibility of it with
```javascript
{...{ ...inputProps, role: ARIA_ROLES.COMBOBOX }}
```

PR #569, opened 2024-08-06
feat: add tabIndex for Link component

jiwang7:
# Description
When a consumer uses a Link without an href attribute, only using onPress to open a new tab, the user cannot focus on the Link.So, add tabIndex for Link component.

# Links


chburket (approved):
looks good!

nataliadelmar (approved):

PR #571, opened 2024-08-07
feat(popover): add disable focus lock

torgeadelin:
# Description
- add prop to disable react-aria focus scoping inside the popover. We need this because at the moment it doesn't have support for shadow-dom, and some of our 3rd party components are built using web-components.
- this will allow us to add our custom focus lock for web-components without interfering with react-aria's focus scope implementation.

# Links
Issue on react-aria: https://github.com/adobe/react-spectrum/issues/1472
PR with potential future solution: https://github.com/adobe/react-spectrum/pull/6046 (it's recent but not sure how long it will take until this gets released).

### DIFF
--- a/src/components/Popover/Popover.types.ts
+++ b/src/components/Popover/Popover.types.ts
@@ -219,4 +219,10 @@ export interface Props extends PopoverCommonStyleProps, Partial<LifecycleHooks> 
    * Only required in the unusual circumstance where the popover label cannot match the trigger.
    */
   'aria-label'?: string;
+
+  /**
+   * If true, the focus lock will be disabled for the Popover content.
+   * @default `false`
+   */
+  disableFocusLock?: boolean;
 }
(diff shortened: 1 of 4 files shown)


src/components/Popover/Popover.types.ts
```diff
@@ -219,4 +219,10 @@ export interface Props extends PopoverCommonStyleProps, Partial<LifecycleHooks>
    * Only required in the unusual circumstance where the popover label cannot match the trigger.
    */
   'aria-label'?: string;
+
+  /**
+   * If true, the focus lock will be disabled for the Popover content.
+   * @default `false`
+   */
+  disableFocusLock?: boolean;
```
maxinteger, line 227: suggestion: please describe in the JSDoc the use case when this prop is useful.

maxinteger (approved):
Added 1 suggestion, otherwise looks good

PR #572, opened 2024-08-08
feat(Ally): Added new props to Menu and MenuTrigger components

Ashwin7mak:
# Description
- Added new prop to Menu component isGroupRole for conditionally adding role="group"
- Added new prop to MenuTrigger component so that role is not defaulted to dialog.

# Links

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -33,6 +33,7 @@ const Menu = <T extends object>(props: Props<T>, providedRef: RefObject<HTMLULis
     isTickOnLeftSide = DEFAULTS.IS_TICK_ON_LEFT_SIDE,
     itemShape = DEFAULTS.ITEM_SHAPE,
     itemSize = DEFAULTS.ITEM_SIZE,
+    isGroupRole,
   } = props;
   const contextProps = useMenuContext();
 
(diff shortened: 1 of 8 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -88,6 +89,7 @@ const Menu = <T extends object>(props: Props<T>, providedRef: RefObject<HTMLULis
           style={style}
           ref={ref}
           {...menuProps}
+          role={!!isGroupRole ? 'group' : menuProps.role}
```
nataliadelmar, line 92: let's make this string 'group' a constant

src/components/Menu/Menu.types.ts
```diff
@@ -36,6 +36,12 @@ export interface Props<T> extends AriaMenuProps<T> {
    * @default false
    */
   isTickOnLeftSide?: boolean;
+
+  /**
+   * Wether it is part of nested menu items.
+   * @default false
+   */
+  isGroupRole?: boolean;
```
nataliadelmar, line 44: let's add this to storybook and let's get the Sections template adjusted to make use of this prop

src/components/MenuTrigger/MenuTrigger.tsx
```diff
@@ -107,6 +108,7 @@ const MenuTrigger = forwardRef((props: Props, ref: ForwardedRef<{triggerComponen
       className={classnames(className, STYLE.wrapper)}
       trigger="click"
       id={id}
+      role={role}
```
nataliadelmar, line 111: it is ok that the menu trigger sets role="dialog" to its triggered element. What we need to make sure is that the first direct child under that dialog is a role="menu"

nataliadelmar (approved):

PR #574, opened 2024-08-08
fix(ListItem): remove aria-current from legacy ListItem

chburket:
This PR is simply to change the legacy ListItem to not add the `aria-current` attribute when the item has focus.  Adding `aria-current` causes the screen reader to read each item with "current" added to the label, which is redundant.

Before:
image

After:
image

Testing notes:
* yarn start
* go to Legacy > ListItem
* open dev tools on (A Center Button)
* ensure the `aria-current` attr is not present on the anchor


chburket:
Since there is already a way to suppress/disable the aria-current attr, I will just use that in cantina

PR #579, opened 2024-08-14
feat(menu): added aria-labelledby prop to Menu component

Ashwin7mak:
# Description
Added new prop 'aria-labelledby' prop to Menu component as it is essential to link Menu header with its items

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-505129)

### DIFF
--- a/src/components/Menu/Menu.stories.args.ts
+++ b/src/components/Menu/Menu.stories.args.ts
@@ -116,7 +116,7 @@ export default {
   },
   'aria-labelledby': {
     description:
-      'The ariaLabelledBy is used to provide an aria-labelledby for the nested menu items inside menu',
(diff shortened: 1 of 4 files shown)


chburket (changes requested):
Looks good!  Can you add ariaLabelledby to the story args?  (see RadioSimple.stories.args.ts for an example)

src/components/Menu/Menu.stories.args.ts
```diff
@@ -114,4 +114,15 @@ export default {
       },
     },
   },
+  'aria-labelledby': {
```
chburket, line 117:
```suggestion
  'ariaLabelledby': {
```
This doesn't work in the story unless it's named the same as the property/attribute.

chburket (changes requested):
see comment

chburket (approved):
LGTM!

PR #581, opened 2024-08-16
feat(only-triggers-tooltip): add onlyTriggersTooltip prop to ButtonCircle and ButtonPill

nataliadelmar:
# Description
ButtonCircle:  For accessibility reasons, informative icons (icons that do not trigger any action on click
        but convey a meaning to the user) should have a visible label next to them or in a tooltip.
        If a tooltip is used, the Icon needs to wrapper in a ButtonCircle onlyTriggersTooltip=true to get no change in color on hover and pressed

image

ButtonPill:   For accessibility reasons, additional explanation to text in the UI should be placed
        next/below to the text or using a Tooltip. If a tooltip is used, the Icon needs to wrapper
        in a ButtonCircle onlyTriggersTooltip=true  to get no change in color on hover and
        pressed

image


src/components/ButtonCircle/ButtonCircle.tsx
```diff
@@ -17,27 +17,35 @@ const ButtonCircle = forwardRef((props: Props, providedRef: RefObject<HTMLButton
...
     ...otherProps
   } = props;
 
   if (ghost && inverted) {
     console.warn('MRV2: Momentum does not support a ghost inverted ButtonCircle.');
   }
 
+  if (onlyTriggersTooltip && (outline || inverted || color || disabled || shallowDisabled)) {
+    console.warn(
+      'MRV2: Momentum does not support a button that onlyTriggersTooltip and is outline, inverted, has color, or is disabled'
```
torgeadelin, line 30: nice!

torgeadelin (approved):
lgtm

PR #585, opened 2024-08-21
fix(modal-header): close button missing aria label

gabrielchl:
# Description
add aria-label to the close btn

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506101

### DIFF
--- a/src/legacy/ModalHeader/index.js
+++ b/src/legacy/ModalHeader/index.js
@@ -27,7 +27,7 @@ class ModalHeader extends React.PureComponent {
       <div className={'md-modal__header' + `${(className && ` ${className}`) || ''}`} {...props}>
         {showCloseButton && (
           <div className="md-modal__close-container">
-            <CloseIcon className="md-modal__close" onClick={handleClose} {...closeBtnProps} />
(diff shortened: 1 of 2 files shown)


src/legacy/ModalHeader/index.js
```diff
@@ -27,7 +27,7 @@ class ModalHeader extends React.PureComponent {
       <div className={'md-modal__header' + `${(className && ` ${className}`) || ''}`} {...props}>
         {showCloseButton && (
           <div className="md-modal__close-container">
-            <CloseIcon className="md-modal__close" onClick={handleClose} {...closeBtnProps} />
+            <CloseIcon className="md-modal__close" onClick={handleClose} aria-label={`Close ${headerLabel || ariaLabel || 'modal'}`} {...closeBtnProps} />
```
torgeadelin, line 30: We can't do this, as it breaks translations. We need a new prop probably that is translated that will be used here.

torgeadelin (changes requested):
1 small issue

gabrielchl:
closing PR, doing this on the cantina side using the existing closeBtnProps prop

PR #588, opened 2024-08-23
feat(sr-announcer): move sr announcer from cantina to mrv2

nataliadelmar:
# Description
Moving SR Announcer from Cantina to MRV2 for other components in Cantina to reuse it. This component causes a screen reader to announce a message when required.

### DIFF
--- a/src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.tsx
+++ b/src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.tsx
@@ -1,0 +1,139 @@
+import React, { useEffect, useState, useRef, useCallback, ReactElement } from 'react';
+
+import { STYLE, DEFAULTS } from './ScreenReaderAnnouncer.constants';
+import {
+  Level,
+  AnnounceOptions,
+  Announce,
+  Clear,
+  Message,
+  AnnouncementProps,
+  AnnouncerProps,
+} from './ScreenReaderAnnouncer.types';
+
+const registry: Record<string, { announce: Announce }> = {};
+
+// This hides the announcer, since it should not be visible on the page (see https://www.a11yproject.com/posts/how-to-hide-content/)
+const containerStyle = {
+  clip: 'rect(0 0 0 0)',
+  clipPath: 'inset(50%)',
+  height: '1px',
+  overflow: 'hidden',
+  position: 'absolute',
+  width: '1px',
+  whiteSpace: 'nowrap',
+} as const;
+
+const register = (identity: string, announce: Announce) => {
+  if (registry[identity]) {
+    throw Error(`ScreenReaderAnnouncer with identity ${identity} is already registered`);
+  }
+  registry[identity] = { announce };
+};
+
+const deregister = (identity: string) => {
+  delete registry[identity];
+};
+
+/**
+ * Announce a message via a screen reader.
+ * To allow for multiple announcers to exist concurrently, an announcer identity can be provided.
+ * This is useful for scenarios where another announcer may be hidden from the screen reader.
+ * An example of this is when an element with role="dialog" and aria-modal="true" is on the page.
+ * In this case, everything outside of the modal is hidden from the accessibility tree.
+ * If an unregistered announcer identity is used, this will error.
+ */
+export const announce = (options: AnnounceOptions, announcerIdentity = DEFAULTS.IDENTITY): void => {
+  if (registry[announcerIdentity]) {
+    return registry[announcerIdentity].announce(options);
+  }
+  throw Error(`ScreenReaderAnnouncer with identity ${announcerIdentity} is not registered`);
+};
+
+const ScreenReaderAnnouncement = ({
+  clear,
+  delay,
+  identity,
+  level,
+  body,
+  timeout,
+}: AnnouncementProps) => {
+  const [message, setMessage] = useState<AnnouncementProps['body']>();
+
+  useEffect(() => {
+    let timer = setTimeout(() => {
+      setMessage(body);
+
+      timer = setTimeout(() => {
+        clear({ messageIdentity: identity });
+      }, timeout);
+    }, delay);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [clear, delay, identity, body, timeout]);
+
+  return (
+    <div aria-live={level} data-testid={`screen-reader-announcement-${identity}`}>
+      {message}
+    </div>
+  );
+};
+
+/**
+ * Component which will cause a screen reader to announce a message when required.
+ * To allow for multiple announcers to exist concurrently, each announcer must have a unique identity.
+ * If no identity is provided, a default one is used (useful for a top level announcer).
+ * If an announcer with a duplicate identity is mounted, an error will occur.
+ */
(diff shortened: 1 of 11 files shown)


src/index.js
```diff
@@ -185,6 +185,8 @@ export {
   AriaGroup,
   RadioSimple,
   RadioSimpleGroup,
+  ScreenReaderAnnouncer,
+  announce,
```
martroos, line 189: suggestion: It might be cleaner to attach the `announce` function to the `ScreenReaderAnnouncer` component, as is done with the `NotificationSystem` - this means that it needs no extra description to make it clear that it is associated with the announcer, and keeps everything more modular.

src/components/index.ts
```diff
@@ -72,6 +72,7 @@ export { default as ComboBox } from './ComboBox';
 export { default as AriaGroup } from './AriaGroup';
 export { default as RadioSimple } from './RadioSimple';
 export { default as RadioSimpleGroup } from './RadioSimpleGroup';
+export { default as ScreenReaderAnnouncer, announce } from './ScreenReaderAnnouncer';
```
martroos, line 75: suggestion: Assuming you go with my suggestion of attaching `announce` to the `ScreenReaderAnnouncer`, the `announce` export won't be needed here.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Announce a message via a screen reader.
+ * To allow for multiple announcers to exist concurrently, an announcer identity can be provided.
+ * This is useful for scenarios where another announcer may be hidden from the screen reader.
+ * An example of this is when an element with role="dialog" and aria-modal="true" is on the page.
+ * In this case, everything outside of the modal is hidden from the accessibility tree.
+ * If an unregistered announcer identity is used, this will error.
+ */
+export const announce = (options: AnnounceOptions, announcerIdentity = DEFAULTS.IDENTITY): void => {
```
martroos, line 46: thought: Previously this was defined _after_ the `ScreenReaderAnnouncer` component - this was to avoid any chance of confusion with the `announce` defined on line 94. You get away with it due to function (block) scoping, so it's not a big deal, but something to keep in mind in future.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.unit.test.tsx
```diff
@@ -0,0 +1,391 @@
...
+  'aria-live': string;
+  children: AnnounceOptions['body'];
+  'data-testid': string | ReturnType<typeof expect.any>;
+};
+type ExpectedAnnouncementNodes = Record<string, Element>;
+
+describe('<ScreenReaderAnnouncer />', () => {
+  let container;
+  beforeAll(() => {
+    jest.clearAllTimers();
```
martroos, line 18: nitpick: I _think_ calling `useFakeTimers` before each test implicitly clears all timers - not a big deal, but I suspect the clear isn't needed.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,139 @@
...
+          identity={messageIdentity}
+          clear={clear}
+          {...messageProps}
+        />
+      ))}
+    </div>
+  );
+};
+
+export default ScreenReaderAnnouncer;
```
martroos, line 139: suggestion: For modularity / namespacing, don't export `announce` above, instead do:

```ts
ScreenReaderAnnouncer.announce = announce;
```

And obviously update relevant places that call it. This means that if we want to expose other methods related to this in future (like the `clear` function`), we can.

martroos (changes requested):
I think it would be worth attaching the `announce` function to the default export, other than that everything looks fine.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.types.ts
```diff
@@ -0,0 +1,61 @@
+import { ReactNode } from 'react';
+
+type Level = 'assertive' | 'polite';
+type Body = string | ReactNode;
```
martroos, line 4: nitpick: I _think_ `ReactNode` includes strings, so probably doesn't need to be explicitly set here.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.types.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface CompoundProps {
+  announce: ScreenReaderAnnouncerAnnounce;
+}
+
+type AnnounceOptions = {
+  /**
+   * A message to announce with a screen reader
+   */
+  body: Body;
```
martroos, line 13: question: Why is this not just text? Since this (the screen reader) is an invisible component, what is the benefit of allowing non-text items here? Not suggesting it's wrong, just trying to understand the use case.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.unit.test.tsx
```diff
@@ -0,0 +1,390 @@
...
+type ExpectedAnnouncement = {
+  'aria-live': string;
+  children: AnnounceOptions['body'];
+  'data-testid': string | ReturnType<typeof expect.any>;
+};
+type ExpectedAnnouncementNodes = Record<string, Element>;
+
+describe('<ScreenReaderAnnouncer />', () => {
+  let container;
+  beforeAll(() => {
```
martroos, line 17: suggestion: I think this does want to be `beforeEach` still.

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.types.ts
nataliadelmar: that is true 
image

Fixed

src/components/ScreenReaderAnnouncer/ScreenReaderAnnouncer.types.ts
nataliadelmar: ohh because sometimes what we want to announce is already a constructed react node on the screen, which we can clone and pass that clone to the sr announcer instaed of passing a stringified version of it.

**Example:** Search input. 
When a search result is empty, we might display something like:
```
<>
<h2>Oh no!!</h2>
<p>We have not been able to find any results</p>
</>
```
We also want the SR to announce "Oh no We have not been able to find any results". So instaed of having to stringify that node, we just pass it directly to the SRAnnouncer

Same for a Toggletip content that does not have simple text inside but has sth like this:
image

martroos: Ah ok, so the content of the toggle tip is passed in as `children`, so it is more effort to stringify them (or demand that they are provided in string form as well). One thing to note is this makes it possible for people to nest a live section within a live section (if someone provided `<div aria-live-"assertive">Some text to announce</div>`), but this is completely avoidable by the developer.

PR #591, opened 2024-08-24
fix(select): optionally overwrite ariaLabelledBy

chrisadubois:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description

Optionally overwrrite aria-labelledby which is normally provided by @react-aria. 

This feels kind of hacky, so if there is a better way to do this with react-aria... i'm all ears. I dug around and could not find anything avilable besides explicitly setting it to null based on consumption params

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-555479
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-555478

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -25,6 +25,7 @@ import Text from '../Text';
 // eslint-disable-next-line @typescript-eslint/ban-types
 function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement>): ReactElement {
   const {
+    ariaLabelledBy,
     className,
     style,
     id,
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.tsx
```diff
@@ -48,7 +49,7 @@ function Select<T extends object>(props: Props<T>, ref: RefObject<HTMLDivElement
 
   const state = useSelectState(props);
   const { labelProps, triggerProps, valueProps, menuProps } = useSelect(props, state, selectRef);
-  const { buttonProps } = useButton({ ...triggerProps, isDisabled }, selectRef);
+  const { buttonProps } = useButton({ ...triggerProps, isDisabled, ...(ariaLabelledBy === '' ? {'aria-labelledby': null} : ariaLabelledBy ? { 'aria-labelledby': ariaLabelledBy } : {})}, selectRef);
```
cscheffauer, line 52: i think this can be simplified to `...(ariaLabelledBy ? {'aria-labelledby': ariaLabelledBy} : {})`

cscheffauer (approved):

src/components/Select/Select.tsx
chburket, line 52: non-blocking: Not a fan of nested ternaries, and this is a bit hard to read.
```suggestion
  const { buttonProps } = useButton(
    { ...triggerProps, isDisabled, 'aria-labelledby': ariaLabelledBy || null },
    selectRef
  );
 ```
 
 Is this close enough?  It looks like you specifically wanted to add nothing if `ariaLabelledBy` is nullish but does it really matter?

chburket (changes requested):
Can you add ariaLabelledBy to the story args so it can be tested in storybook?

src/components/Select/Select.tsx
chrisadubois: this ^ unfortunately does not work. I did in fact try this. react-aria is weird. this is a hack

chburket (approved):
Approved after offline discussion

PR #592, opened 2024-08-28
fix(overlay-alert): missing gap in the header when there are no controls

gabrielchl:
noticed when doing https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-555472


Coread (approved):
This does also change the focus ring for the close button, from a rounded top right corner to a completely square box. I think that is fine.

PR #593, opened 2024-08-28
feat(toggle-announcement): add sr announcement to togglegtip

nataliadelmar:
# Description

As per the toggletip pattern, toggletips should make sr announce its content when the open. 

image
Source: https://inclusive-components.design/tooltips-toggletips/

So Ive used SRAnnouncement component to do it. 

# Links

### DIFF
--- a/src/components/Toggletip/Toggletip.tsx
+++ b/src/components/Toggletip/Toggletip.tsx
@@ -34,7 +34,7 @@ const Toggletip = forwardRef(
   ) => {
     const tippyRef = useRef<PopoverInstance>(null);
     const triggerComponentRef = useRef<HTMLElement>(null);
-    const [announcerId] = useState<string>(uuidV4());
+    const [announcerId] = useState<string>(uuidV4);
     const [isTooltipOpen, setIsTooltipOpen] = useState<boolean>(false);
 
     // SR should announce the content of the toggletip everytime it displays tooltip content
@@ -50,7 +50,7 @@ const Toggletip = forwardRef(
       if (isTooltipOpen) {
         ScreenReaderAnnouncer.announce({ body: children }, announcerId);
       }
-    }, [isTooltipOpen]);
+    }, [isTooltipOpen, children]);
 
     // Update aria props manually, because "The `aria` attribute is reserved for future use in React."
     // see https://atomiks.github.io/tippyjs/v6/all-props/#aria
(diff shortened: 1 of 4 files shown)


src/components/Toggletip/Toggletip.tsx
```diff
@@ -33,6 +34,23 @@ const Toggletip = forwardRef(
   ) => {
     const tippyRef = useRef<PopoverInstance>(null);
     const triggerComponentRef = useRef<HTMLElement>(null);
+    const announcerId = useRef(uuidV4());
```
maxinteger, line 37: Please use `useState` instead of `useRef`

https://confluence-eng-gpk2.cisco.com/conf/display/WTWC/Use+useState+not+useRef+for+persistent+generated+ids

src/components/Toggletip/Toggletip.tsx
```diff
@@ -33,6 +34,23 @@ const Toggletip = forwardRef(
...
+    const [displayedTooltip, setDisplayedTooltip] = useState<React.ReactNode>(null);
```
maxinteger, line 38: suggestion: it would be better to use a boolean state instead

maxinteger (changes requested):

src/components/Toggletip/Toggletip.tsx
```diff
@@ -33,6 +34,23 @@ const Toggletip = forwardRef(
   ) => {
     const tippyRef = useRef<PopoverInstance>(null);
     const triggerComponentRef = useRef<HTMLElement>(null);
+    const [announcerId] = useState<string>(uuidV4());
```
maxinteger, line 37: The convention says you do not need the `()` after `uuidV4` because the state call call it:

```suggestion
    const [announcerId] = useState<string>(uuidV4);
```

src/components/Toggletip/Toggletip.tsx
```diff
@@ -33,6 +34,23 @@ const Toggletip = forwardRef(
...
+
+    const handleHide = () => {
+      setIsTooltipOpen(false);
+    };
+
+    useEffect(() => {
+      if (isTooltipOpen) {
+        ScreenReaderAnnouncer.announce({ body: children }, announcerId);
+      }
+    }, [isTooltipOpen]);
```
maxinteger, line 53: this effect should depends on the `children` as well

maxinteger (changes requested):

maxinteger (approved):
Looks Good!

PR #600, opened 2024-08-30
feat(listitembase): dynamic list support

Coread:
# Description

This PR contains a number of improvements to List/ListItemBase:
1) Improved handling of adding and removing items from the list
2) Exclusion of single tabbable elements from the focus handling logic in ListItemBase
3) onFocus/blur and onFocusWithin/blurWithin

# Links


src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -271,6 +271,24 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
     };
   }, []);
 
+  useEffect(() => {
```
martroos, line 274: issue: This breaks looping lists if one of the ends is considered empty - maybe not a huge issue, but it doesn't behave as expected.

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -271,6 +271,24 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
...
+        }
+      } else {
+        if (itemIndex >= listContext.listSize - 1) {
+          listContext.setCurrentFocus(itemIndex - 1);
+        } else {
+          listContext.setCurrentFocus(itemIndex + 1);
+        }
+      }
+    }
+  }, [focus, lastCurrentFocus, listContext.listSize]);
```
martroos, line 290: suggestion: I think this should include `indexItem` and probably `listContext.setCurrentFocus` (the latter is admittedly the result of a `useState` so not hugely important, but this module doesn't really know what it is based on the context).

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -271,6 +271,24 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
...
+          listContext.setCurrentFocus(itemIndex + 1);
+        } else {
+          listContext.setCurrentFocus(itemIndex - 1);
+        }
+      } else {
+        if (itemIndex >= listContext.listSize - 1) {
+          listContext.setCurrentFocus(itemIndex - 1);
+        } else {
+          listContext.setCurrentFocus(itemIndex + 1);
+        }
```
martroos, line 287: suggestion: Some comments to explain what is happening would be useful - with some analysis you can see that you are actively keeping items at the (direction dependent) previous item if this item is at one of the ends, and otherwise skipping it, but it's not necessarily simple at a glance (the comment would have also made it clear that looping wouldn't work any more).

src/components/List/List.stories.tsx
```diff
@@ -327,4 +327,39 @@ const ProgramaticFocus = Template<ListProps>((args) => {
   );
 }).bind({});
 
-export { Example, Common, HorizontalList, CalendarList, DynamicList, Search, ProgramaticFocus };
+const ListWithNullElementsWrapper = () => {
+  return (
+    <>
+      <List noLoop listSize={7}>
```
martroos, line 333: issue: This is a list of length 9 (looping will happen to work going down because this number is wrong!).

src/components/ListItemBase/ListItemBase.style.scss
```diff
@@ -10,6 +10,20 @@
...
+  }
+
+  &:empty {
+    clip: rect(0 0 0 0);
+    clip-path: inset(50%);
+    height: 1px;
+    overflow: hidden;
+    position: absolute;
+    width: 1px;
+    white-space: nowrap;
```
martroos, line 24: nitpick: This is another instance of the ["visually hidden"](https://www.a11yproject.com/posts/how-to-hide-content/) styling which is in various parts of the library, would be nice to mixin / include this in scss terms.

martroos (dismissed):

src/components/ListItemBase/ListItemBase.style.scss
Coread: as far as I know, we don't currently have shared scss between the new components?

src/components/List/List.stories.tsx
Coread: fixed

src/components/ListItemBase/ListItemBase.tsx
Coread: this is now fixed

src/components/ListItemBase/ListItemBase.tsx
Coread: I've added the full list of dependencies to the useEffect

src/components/ListItemBase/ListItemBase.tsx
Coread: this is now fixed

src/components/ListItemBase/ListItemBase.types.ts
```diff
@@ -92,8 +92,32 @@ export interface Props extends PressEvents, ContextMenu {
...
+  onBlurWithin?: (e: React.FocusEvent<Element>) => void;
+  /**
+   * event fired when any of the children or the element itself are focused
+   */
+  onFocus?: (e: React.FocusEvent<Element>) => void;
+
+  /**
+   * event fired when any of the children or the element itself lose focus
+   */
+  onBlur?: (e: React.FocusEvent<Element>) => void;
```
maxinteger, line 122: `Props` should extends `FocusProps` and `FocusWithinProps` instead of manually adding the props. JSDocs should be moved to the original definition.

src/components/List/List.stories.tsx
```diff
@@ -327,4 +327,251 @@ const ProgramaticFocus = Template<ListProps>((args) => {
...
+    </>
+  );
+};
+const ListWithButtons = Template<unknown>(ListWithButtonsWrapper).bind({});
+
+const ListWithNonFocusableChildrenWrapper = () => {
+  return (
+    <>
+      <List shouldFocusOnPress listSize={2}>
+        <ListItemBase data-exclude-focus-children size="auto" itemIndex={0} key={0}>
```
maxinteger, line 429: What is the expected behaviour for `data-exclude-focus` and `data-exclude-focus-children`? I can see they might be excluded from intractable element selection, but it is not prevent the browser focus the element with Tab.

I check this story, and I can focus on button 1 with Tab.

maxinteger (changes requested):

src/hooks/useFocusState.ts
```diff
@@ -27,4 +33,36 @@ const useFocusState = (
...
+const useFocusWithinState = (
+  props: FocusWithinProps
+): { isFocusedWithin: boolean; focusWithinProps: HTMLAttributes<HTMLElement> } => {
+  const [isFocusedWithin, setFocusWithin] = useState(false);
+  const { focusWithinProps } = useFocusWithin({
+    onFocusWithin: (...args) => {
+      setFocusWithin(true);
+      if (props.onFocusWithin) {
+        props.onFocusWithin.apply(this, args);
+      }
```
maxinteger, line 45: nitpicking: this can be simplified to:
```suggestion
    props.onFocusWithin?.(...args);
```
because the arrow functions `this` is `undefined` anyway

src/components/List/List.stories.tsx
Coread: I've removed the exclude-focus-children as this was covered by your recent changes. data-exclude-focus is used by the Aria-Toolbar. The toolbar only wants to expose one of its tabable buttons to preserve focus order when you tab forward/backward from within a ListItemBase. It sets data-exclude-focus on all but the first one so the first element is the only one to ever be in the tab order.

src/components/ListItemBase/ListItemBase.types.ts
Coread: I'm now extending this definition using the props from react-aria. Note that it isn't as simple as extending directly, because I haven't added support for all of FocusProps or all of FocusWithinProps

src/hooks/useFocusState.ts
Coread: I've changed this

.eslintrc.js
jor-row, line 1: On press of root nodes in tree the component is crashing
https://github.com/user-attachments/assets/6669c4dc-24d2-4d74-83b9-0c3edbd861c2

jor-row: Menu SelectionGroups are broken (and a unit test is failing because of it!) Keyboard nav doesn't work at all for them in this PR but is fine on master
https://github.com/user-attachments/assets/e378db7f-e767-436e-b929-4906a31d183a

jor-row (changes requested):

jor-row (approved):
Tested locally and changes look good to me. A few small comments

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -141,59 +172,120 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
...
    * we want to silently update the currentFocus back to the first element in
    * case the index of the element focused before the list shrink is now outside
    * the size of the new list size (shrinked size)
    */
-  useEffect(() => {
-    if (!!listContext?.listSize && listContext?.currentFocus >= listContext?.listSize) {
-      // set focus to first item
-      listContext.setCurrentFocus(0);
+  useLayoutEffect(() => {
+    // if (!!listSize && currentFocus >= listSize) {
```
maxinteger, line 275: please remove this comment

maxinteger (approved):

src/components/ListItemBase/ListItemBase.tsx
jor-row: Done

src/hooks/useOrientationBasedKeyboardNavigation.ts
```diff
@@ -1,23 +1,37 @@
...
+  focusWithinProps: HTMLAttributes<HTMLElement>;
   getContext: () => {
     listSize: number;
     currentFocus: number;
     setCurrentFocus: Dispatch<SetStateAction<number>>;
     shouldFocusOnPress?: boolean;
     shouldItemFocusBeInset?: boolean;
+    noLoop?: boolean;
+    updateFocusBlocked?: boolean;
+    supressFocus?: boolean;
```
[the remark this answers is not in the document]
jor-row: done

src/components/AriaToolbar/AriaToolbar.tsx
```diff
@@ -59,15 +61,31 @@ const AriaToolbar: FC<Props> = (props: Props) => {
...
+      return (
+        <div ref={ref} {...commonProps}>
+          {children}
+        </div>
+      );
     }
   };
 
+  useEffect(() => {
+    // When the toolbar is rendered inside a list, only item in the toolbar
```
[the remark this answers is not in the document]
jor-row: done

src/components/List/List.stories.tsx
```diff
@@ -327,4 +327,338 @@ const ProgramaticFocus = Template<ListProps>((args) => {
...
+        <ListItemBase itemIndex={1} key={1}>
+          <ButtonPill>2</ButtonPill>
+        </ListItemBase>
+      </List>
+    </>
+  );
+};
+const ListWithNonFocusableChildren = Template<unknown>(ListWithNonFocusableChildrenWrapper).bind(
+  {}
+);
```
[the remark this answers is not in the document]
jor-row: fixed

PR #604, opened 2024-09-02
feat(buttons): revert onlyTriggersTooltip

nataliadelmar:
# Description

Reverting this property as the a11y team prefers us not wrapping icons in buttons to make a toggletip trigger. They say wrapping it on a div with tabIndex 0 is enough.


torgeadelin (approved):
image

PR #606, opened 2024-09-03
feat: handle dynamic changes in tree component

maxinteger:
# Description

- re-calculate internal tree when tree structure change
- do not render TreeNodeBase content when the node details are not available


nataliadelmar (approved):
works well

PR #607, opened 2024-09-03
fix(tree): missing recalculation of mapped tree on tree structure change

gabrielchl:
it wasn't recalculating and updating the internal tree data, causing a page to crash if the tree data were to update (context tried to get an invalid node)


gabrielchl:
already being fixed in #606 oopsie doopsie

PR #608, opened 2024-09-03
fix(avatar): avatar add buttonclassname prop

cscheffauer:
# Description

- Allow passing in a buttonClassName to override the classname of the button of the avatar
- This is required to fix a a11y issue.


jor-row (approved):
LGTM

PR #615, opened 2024-09-04
fix(popover): remove event listener onDestroy, not onHidden

robstax:
# Description
* the focusOut event listener should be removed when the tippy instance is destroyed, not hidden. if done onHidden, blur handling only works once since the event listener gets removed onHidden.

# Links
* https://atomiks.github.io/tippyjs/v6/all-props/#ondestroy


chburket (approved):
lgtm

stanjiawang (approved):

PR #620, opened 2024-09-05
feat: tree node selection

maxinteger:
# Description

- added useItemSelected hook
- integrated useItemSelected into Tree and TreeNodeBase
- fixed active node calculation after tree structure change
- merged all node aria props context functions into one
- fixed scrollToNode in VTree only node is not in the DOM
- added imperative Tree component API
- added new stories

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-554523


src/components/Tree/Tree.test.tsx
```diff
@@ -596,4 +663,118 @@ describe('<Tree />', () => {
...
+                {() => node.id}
+              </TreeNodeBase>
+            ),
+            { excludeRootNode: false }
+          )}
+        </Tree>
+      );
+    };
+
+    describe('with keyboard navigation', () => {
```
torgeadelin, line 694: praise: great testing coverage 🎉

src/components/Tree/Tree.tsx
```diff
@@ -25,132 +37,169 @@ const Tree: FC<Props> = (props: Props) => {
...
+    selectedByDefault,
+    onSelectionChange,
+    isRequired = DEFAULTS.IS_REQUIRED,
     ...rest
   } = props;
 
-  const ref = useRef<HTMLDivElement>();
+  const treeRef = useRef<HTMLDivElement>();
 
+  const itemSelection = useItemSelected<TreeNodeId>({
```
torgeadelin, line 50: nitpick: react-aria has some nice selection hooks, maybe in the future we could look into reusing those

src/components/Tree/Tree.tsx
```diff
@@ -25,132 +37,169 @@ const Tree: FC<Props> = (props: Props) => {
...
+    selectionMode = DEFAULTS.SELECTION_MODE,
+    selectableNodes = DEFAULTS.SELECTABLE_NODES,
+    selectedByDefault,
+    onSelectionChange,
+    isRequired = DEFAULTS.IS_REQUIRED,
     ...rest
   } = props;
 
-  const ref = useRef<HTMLDivElement>();
+  const treeRef = useRef<HTMLDivElement>();
```
torgeadelin, line 48: suggestion: we should probably allow connecting `ref` that is provided via forwaradRef to the treeRef. There's a new hook that combines them (and also handles both func and non func refs).

https://github.com/momentum-design/momentum-react-v2/blob/9871677f681a1543f0b438c0b7a52b239fc16a66/src/utils/useProvidedRef.ts#L10

src/components/Tree/Tree.tsx
```diff
@@ -25,132 +37,169 @@ const Tree: FC<Props> = (props: Props) => {
...
+    let newActiveNodeId = activeNodeId;
+    while (newActiveNodeId) {
+      if (newTree.has(newActiveNodeId)) break;
+      newActiveNodeId = previousTree.get(newActiveNodeId)?.parent;
+    }
+    // Fallback to the first node
+    if (!newActiveNodeId) {
+      newActiveNodeId = excludeTreeRoot ? treeStructure?.children?.[0]?.id : treeStructure.id;
+    }
+    setActiveNodeId(newActiveNodeId);
```
torgeadelin, line 76: question: why are these changes needed?

src/components/Tree/Tree.tsx
torgeadelin: neverming, i saw you're using imperative handle.

src/components/Tree/Tree.types.ts
```diff
@@ -150,6 +153,16 @@ export interface Props extends DetailedHTMLProps<HTMLAttributes<HTMLDivElement>,
    */
   excludeTreeRoot?: boolean;
 
+  /**
+   * The selection mode of the tree nodes.
+   *
+   * Note: When user cli ck to select a node and `selectableNodes` is `any` and the node is not a leaf node, the node will be opened/closed.
```
torgeadelin, line 159: typo: user clicks

src/components/Tree/Tree.types.ts
```diff
@@ -210,4 +252,33 @@ export interface TreeContextValue extends Pick<Props, 'shouldNodeFocusBeInset' |
...
+   */
+  treeRef: MutableRefObject<HTMLDivElement>;
+  /**
+   * Toggle a single item selection
+   */
+  toggleSelection: ItemSelection<TreeNodeId>['toggle'];
+  /**
+   * Update selection with new list of selected items
+   */
+  updateSelection: ItemSelection<TreeNodeId>['update'];
```
torgeadelin, line 279: question: what is the difference between these two and why do we need both?

src/hooks/useItemSelected.test.ts
```diff
@@ -0,0 +1,498 @@
...
+          expect(onSelectionChange).not.toHaveBeenCalled();
+        }
+      );
+    });
+
+    describe('isSelected', () => {
+      it.each`
+        selectedByDefault | expected
+        ${undefined}      | ${false}
+        ${['1']}          | ${false}
```
torgeadelin, line 53: question: shouldn't this return true?

torgeadelin: oh, it's because selectionMode is `none`

torgeadelin (commented):
looks and works good, just some small questions

src/components/Tree/Tree.test.tsx
```diff
@@ -306,7 +332,7 @@ describe('<Tree />', () => {
 
         if (toggleNode) {
           expect(setNodeOpen).toHaveBeenCalledWith(nextFocusedNode, true);
-          expect(scrollToNode).toHaveBeenNthCalledWith(1, nextFocusedNode);
+          expect(scrollToNode).not.toHaveBeenNthCalledWith(1, nextFocusedNode);
```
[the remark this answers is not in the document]
maxinteger: because previously we called the scroll to every time when the the user clicked on it. It was unnecessary when the node already visible and it caused jump in the tree view.

From now on, we call `scrollToNode` only when the node is not in the DOM.

src/components/Tree/Tree.test.tsx
```diff
@@ -324,7 +350,7 @@ describe('<Tree />', () => {
 
         if (toggleNode) {
           expect(setNodeOpen).toHaveBeenCalledWith(nextFocusedNode, false);
-          expect(scrollToNode).toHaveBeenCalledWith(nextFocusedNode);
+          expect(scrollToNode).not.toHaveBeenCalledWith(nextFocusedNode);
```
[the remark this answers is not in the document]
maxinteger: same answer as above

src/components/Tree/Tree.test.tsx
maxinteger: almost 101%

src/components/Tree/Tree.tsx
maxinteger: good idea, although we usually add react-aria hooks and then try to workaround them...

src/components/Tree/Tree.tsx
maxinteger: The component re-calculate the tree when a new tree structure come in. But we do it even when the data in the tree structure did not changed, or it changed (node opened/closed) without effecting the active node visibility. In these cases we must keep the active node in focus. 

This code also handle the case when we remove node removed but one of its parent still there.

I added this change because I faced with the focus reset problem in the breakout session.

I just noticed we do not handle the case when the node is exists, bug hidden which might cause focus lost. I will fix it

src/components/Tree/Tree.types.ts
maxinteger: `toggleSelection` is for change a single item selection. It is great for selectable items because toggles the selection by default.

`updateSelection` is for programatic update of the selection, for example when you want to do "select all" or "invert selection". It will trigger a single change in React instead of hundreds if you have a long list

PR #621, opened 2024-09-05
feat(searchinput): allow searchinput to accept an onkeydown prop

robstax:
# Description
* allow SearchInput to accept an onKeyDown prop. the parent component may want the ability to act on keydown, for example, when pressing "down" from input may need to switch focus to a result


nataliadelmar (approved):
lgtm

PR #624, opened 2024-09-05
fix(Menu): Keyboard navigation fix for menus with multiple sections

johnsoter13:
# Description

1. Removing Multimenu story as we are no longer going to follow the approach of having more than 1 menu on a single popover.
2. For the previous reason, we are no longer needing the isGroupRole prop on Menu.
3. Google Chrome Settings Menu has horizontally displayed items that are navigated using up and down arrow key. For that reason, we are removing the orientation prop from Menu as we no longer need to enforce the kb navigation to left-right if the items are horizontally displayed.
4. We are removing any List/ul traces from Menu. We realized menus with sections had 3 layers of ul's without any purpose. 
5. We are letting menu navigation be dictated by react-aria useMenu hook, and not by list navigation hooks. For that reason, we are also removing itemIndex prop from MenuSection and MenuItem. 

# Links

### DIFF
--- a/src/components/MenuSection/MenuSection.tsx
+++ b/src/components/MenuSection/MenuSection.tsx
@@ -7,7 +7,6 @@ import { Props } from './MenuSection.types';
 import './MenuSection.style.scss';
 import MenuItem from '../MenuItem';
 import { useMenuSection } from '@react-aria/menu';
-import { ListContext } from '../List/List.utils';
 
 const MenuSection = <T extends object>(props: Props<T>): ReactElement => {
   const { item, state, onAction } = props;
@@ -24,7 +23,7 @@ const MenuSection = <T extends object>(props: Props<T>): ReactElement => {
   }, [state]);
 
   return (
-    <ul {...itemProps}>
+    <div {...itemProps}>
       {!React.isValidElement(item.rendered) && item.rendered ? (
         <span className={STYLE.header} {...headingProps}>
           {item.rendered}
@@ -32,12 +31,10 @@ const MenuSection = <T extends object>(props: Props<T>): ReactElement => {
       ) : (
         item.rendered && React.cloneElement(item.rendered as ReactElement, { ...headingProps })
       )}
-      <ListContext.Provider value={{}}>
-        <ul {...groupProps} className={STYLE.wrapper}>
-          {renderItems()}
-        </ul>
-      </ListContext.Provider>
-    </ul>
+      <ul {...groupProps} className={STYLE.wrapper}>
+        {renderItems()}
+      </ul>
+    </div>
   );
 };
 
--- a/src/components/MenuSection/MenuSection.tsx
+++ b/src/components/MenuSection/MenuSection.tsx
@@ -8,23 +8,18 @@ import './MenuSection.style.scss';
 import MenuItem from '../MenuItem';
 import { useMenuSection } from '@react-aria/menu';
 import { ListContext } from '../List/List.utils';
-import useOrientationBasedKeyboardNavigation from '../../hooks/useOrientationBasedKeyboardNavigation';
 
 const MenuSection = <T extends object>(props: Props<T>): ReactElement => {
-  const { item, state, onAction, orientation } = props;
+  const { item, state, onAction } = props;
 
   const { itemProps, headingProps, groupProps } = useMenuSection({
     heading: item.rendered,
     'aria-label': item['aria-label'],
   });
-  const childItems = Array.from(item.childNodes);
-  const listSize = childItems.length;
-
-  const {keyboardProps, getContext} = useOrientationBasedKeyboardNavigation({listSize, orientation});
 
   const renderItems = useCallback(() => {
-    return Array.from(item.childNodes).map((node, index) => (
-      <MenuItem itemIndex={index} key={node.key} item={node} state={state} onAction={onAction} />
+    return Array.from(item.childNodes).map((node) => (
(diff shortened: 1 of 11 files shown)


src/hooks/useOrientationBasedKeyboardNavigation.ts
```diff
@@ -15,8 +15,8 @@ type IUseOrientationBasedKeyboardNavigationReturn = {
 };
 
 export type IUseOrientationBasedKeyboardNavigationProps = {
-  listSize: number;
-  orientation: ListOrientation;
+  listSize?: number;
+  orientation?: ListOrientation;
```
shnaaz, line 19: Why this change was required, we are not using this hook in Menu or MenuSection

maxinteger, line 19: Please undo this change, `orientation` and `listSize` are critical to calculate the next focus item.

The consumer of this hook can set the default value for `orientation`, also define the `listSize`

src/components/MenuTrigger/MenuTrigger.unit.test.tsx.snap
maxinteger, line 1: Please re-generate the snapshot file for the MenuTrigger and commit it

src/components/MenuSection/MenuSection.types.ts
```diff
@@ -18,16 +18,4 @@ export interface Props<T> {
...
-   * Determines the orientation of the list
-   *
-   * The orientation of the list change the keyboard navigation in the list:
-   *
-   * - vertical: up and down arrow keys
-   * - horizontal: left and right arrow keys
-   *
-   * @default 'vertical'
-   */
-  orientation?: ListOrientation
```
maxinteger, line 32: As far as I know there is at least one example for horizontal menu in cantina. Please make sure this is not breaking Cantina

src/components/MenuSection/MenuSection.tsx
```diff
@@ -37,8 +32,8 @@ const MenuSection = <T extends object>(props: Props<T>): ReactElement => {
       ) : (
         item.rendered && React.cloneElement(item.rendered as ReactElement, { ...headingProps })
       )}
-      <ListContext.Provider value={getContext()}>
-        <ul {...groupProps} {...keyboardProps} className={STYLE.wrapper}>
+      <ListContext.Provider value={{}}>
```
maxinteger, line 35: Since Menus are a list like components from navigation point of view, so I do not understand why the ListContext has no value at this point

maxinteger (changes requested):
Sorry but I do not understand what issue you try to fix on the menu navigation.

Next time please add the issue link and the fill the description section of the PR.

src/hooks/useOrientationBasedKeyboardNavigation.ts
nataliadelmar, line 19: not needed
useOrientationBasedKeyboardNavigationProps is only used once from Lists now, and there you always will send listSize (as it is a required prop on List) and orientation is defaulted to VERTICAL, so this change to make them optional here is not needed :)

nataliadelmar (approved):

src/hooks/useOrientationBasedKeyboardNavigation.ts
nataliadelmar: I pushed a fix to fix this one bit

torgeadelin:
I have a feeling this is the revert PR??

torgeadelin:
Can we please use the revert feature and then create a new PR with the actual fix? @johnsoter13

nataliadelmar:
> Sorry but I do not understand what issue you try to fix on the menu navigation.
> 
> Next time please add the issue link and the fill the description section of the PR.

Before using the kb-navigation hook in menus, a menu was able to have sections inside that were navigated using arrow keys (across sections).

nataliadelmar:
> I have a feeling this is the revert PR??

it is the revert pr, but we want to keep things from that pr in master, which is why this new pr was made

src/hooks/useOrientationBasedKeyboardNavigation.ts
johnsoter13: Yeah I changed this back, this was some testing I was doing

src/components/MenuSection/MenuSection.types.ts
johnsoter13: I did, we no longer need to specify orientation on the MenuSection level. Can all be handled in Menu.

src/components/MenuSection/MenuSection.tsx
johnsoter13: This is just reverting the Menu back to it's original state. I thought we needed the context defined in order to do orientation based navigation but that's not the case

src/hooks/useOrientationBasedKeyboardNavigation.ts
johnsoter13: I thought I pushed a change to this but I left it in local accidentally. This was just from some testing I was doing and forgot to change it back

johnsoter13:
Sorry for the confusion @torgeadelin and @maxinteger . @nataliadelmar  and I discussed that reverting the old PR and pushing a new one was not the direction we wanted to go, but I didn't communicate that when I made this PR. 

But as Natalia said, we wanted to keep the old PR in master. This work isn't a full revert, just an adjustment to support kb navigation across sections while also supporting orientation prop for Menus

shnaaz (approved):
Also tested with Cantina, seems fine.

maxinteger (approved):

PR #626, opened 2024-09-07
tippy blur esc plugin

robstax:
# Description
* confirmed with a11y team, but when closing the popover through Tab, focus should go to the next focusable element. when pressing Esc, focus should go back to the trigger element.

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -16,8 +16,6 @@ import { addTippyPlugins } from './Popover.utils';
 import { v4 as uuidV4 } from 'uuid';
 import { PopoverInstance } from '.';
 
-type CustomInstance = PopoverInstance & { props: Props } & {hasRelatedTarget?: boolean;};
-
 /**
  * The Popover component allows adding a Popover to whatever provided
  * `triggerComponent`. It will show the Popover after a specific event, which is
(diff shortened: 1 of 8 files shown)


src/components/Popover/Popover.tsx
```diff
@@ -119,12 +124,15 @@ const Popover = forwardRef((props: Props, ref: ForwardedRef<HTMLElement>) => {
   // needs special handling since FocusScope doesn't work with the Popover from Tippy
   // needs to focus back to the reference item when the popover is completely hidden
   const handleOnPopoverHidden = useCallback(() => {
-    if (focusBackOnTrigger) {
+    // if focusBackOnTrigger props is true AND we have nothing else to focus on
+    // in the case of tab navigation, we should instead focus on the next focusable?
+    // element, so don't focus the instance reference
```
nataliadelmar, line 129: I'd change the comment to:
```    
When the popover hides, the focus goes to the next focusable element by default. Except if focusBackOnTrigger popover prop is true AND shouldFocusTrigger (determined by hideOnBlurPlugin) is true. 

shouldFocusTrigger is true when the focusout event has no relatedTarget, that is, focusOut was triggered by Esc or Click.

```

nataliadelmar (approved):
small minor comments but lgtm

PR #627, opened 2024-09-09
feat(Text): add title attribute

jiwang7:
# Description

- add title props 
# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506030
image

### DIFF
--- a/src/components/Text/Text.stories.args.ts
+++ b/src/components/Text/Text.stories.args.ts
@@ -42,4 +42,16 @@ export default {
       },
     },
   },
+  title: {
+    description: 'Show the content when hovering over the component.',
+    control: { type: 'text' },
+    table: {
+      type: {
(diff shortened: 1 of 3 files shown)


maxinteger (changes requested):
`title` attribute has a poor Screen reader support and it has other UX and Accessibility issues. There are many articles about this problem, for example: https://www.a11yproject.com/posts/title-attributes/

Sorry, but I can not accept this PR until you describe your use case

jiwang7:
> `title` attribute has a poor Screen reader support and it has other UX and Accessibility issues. There are many articles about this problem, for example: https://www.a11yproject.com/posts/title-attributes/
> 
> Sorry, but I can not accept this PR until you describe your use case

I haven't noticed the effect of Screen reader. I will close the PR. Thanks!

PR #628, opened 2024-09-09
style(high-contrast): add focus outline, checkbox styles and tab active styles in high contrast mode

dominiccarrington:
# Description

When Windows High Contrast mode is enabled, certain CSS properties are forced to have certain values. One of these is `box-shadow` which is forced to `none`.

`box-shadow` is used to display the focused element therefore we need a alternative. Since `outline` is already forced to none when using the `focus-styles` mixin, we can use it to display a focus ring.

# Examples
Unfocused:
image

Focused:
image

---

Unfocused:
image

Focused:
image

---

image

# Links
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors

Should fix a couple of issues raised here:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506028

### DIFF
--- a/scss/tools/mixins/focus.scss
+++ b/scss/tools/mixins/focus.scss
@@ -5,12 +5,8 @@
     border-color: $border-color;
   }
 
-  outline: none !important;
+  outline: var(--md-globals-focus-ring-outline) !important;
   box-shadow: var(--md-globals-focus-ring-box-shadow);
-
-  @media (forced-colors: active) {
-    outline: 0.125rem solid var(--mds-color-theme-outline-theme-normal) !important;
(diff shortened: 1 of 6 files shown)


maxinteger (changes requested):
The PR looks good, but seems like it does not take into account the `isInset` prop of the [`FocusRing`](src/components/FocusRing/FocusRing.tsx).

image

you can do [`outline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset) for inset focus ring in force colour mode.

maxinteger (approved):

PR #633, opened 2024-09-11
fix: tree component selection

maxinteger:
# Description

- added selectedItems prop to useItemSelected hook to make it controllable input like component.
- added active-node class to TreeNodeBase

### DIFF
--- a/src/components/Tree/Tree.test.tsx
+++ b/src/components/Tree/Tree.test.tsx
@@ -3,6 +3,7 @@ import { mount } from 'enzyme';
 import '@testing-library/jest-dom';
 import { render, waitFor, waitForElementToBeRemoved } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import { act } from 'react-test-renderer';
 
 import Tree, { TREE_CONSTANTS, TreeProps } from './index';
 import { createTreeNode as tNode } from './test.utils';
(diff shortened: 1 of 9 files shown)


src/components/Tree/Tree.test.tsx
```diff
@@ -809,5 +800,42 @@ describe('<Tree />', () => {
         }
       );
     });
+
+    it('should selectionMode change reflected in selection', async () => {
```
torgeadelin, line 804: nitpick: test title doesn't really make sense

src/hooks/useItemSelected.ts
```diff
@@ -92,58 +103,77 @@ export const useItemSelected = <TItemId extends string | number>({
...
   onSelectionChange,
+  selectedItems: selectedItemsFromProps,
   isRequired,
 }: UseItemSelectedProps<TItemId>): ItemSelection<TItemId> => {
-  const [selectedItems, setSelectedItems] = useState<Set<TItemId>>(
+  const isControlled = selectedItemsFromProps !== undefined;
+
+  const prevIsControlled = usePrevious(isControlled);
+
+  if (prevIsControlled === true && prevIsControlled !== isControlled) {
```
torgeadelin, line 113: suggestion: a comment on why this could happen would be useful. Also the warning message is confusing.

src/hooks/useItemSelected.ts
```diff
@@ -92,58 +103,77 @@ export const useItemSelected = <TItemId extends string | number>({
...
+  );
+
   return useMemo(
     () => ({
       selectionMode,
       selectedItems: Array.from(selectedItems),
       isSelected: (itemId: TItemId) => selectedItems.has(itemId),
       toggle(itemId: TItemId, value?: boolean) {
         if (selectionMode === 'none') return;
-        setSelectedItems((prev) => {
```
torgeadelin, line 108: question: would this change from using func to make the state update to the simple one `setInternalSelectedItems(result);` not result in inconsistencies? I'm thinking about the counter example, where the counter could end up in a mixed state because it reads old value. 

Also, our `prev` is still from selectedItems, but we're updating the internalSelectedItems with a new state based on prev which is based on selectedItems and not internalSelectedItems. Could this also cause inconsistencies?

torgeadelin: Like for example when the user toggles an item really quickly multiple times.

torgeadelin: I just tested and seems to work well, but just wondering...

torgeadelin (approved):
some small non-blocking questions, but otherwise looks great!

src/hooks/useItemSelected.ts
maxinteger: I refactored it, in controlled mode we completely ignore the internal state, otherwise update the internal state based on the previous value

PR #634, opened 2024-09-12
feat(space-list-item): split space list item component

gabrielchl:
in prep. for the team spaces list -> tree work, the `<SpaceListItem />` component needs to be split into 2:
- `<SpaceRowContent />`; and
- `<SpaceListItem />`, which uses `<SpaceRowContent />`

this PR simply moves the children part into a new component

a further PR will create a 3rd component: `<SpaceTreeNode />`, which also uses `<SpaceRowContent />`


maxinteger (approved):
Nice one!

PR #635, opened 2024-09-13
feat(text): Add more tag extensions

weiwli2:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description
Add more tag extensions, include "div" "span".

# Links


src/components/Text/Text.tsx
```diff
@@ -31,6 +31,12 @@ const Text: FC<Props> = (props: Props) => {
       case TYPES.HYPERLINK_PRIMARY:
         return 'p';
 
+      case TYPES.BUTTON_PRIMARY:
+        return 'span';
+
+      case TYPES.LIST_ITEM:
+        return 'div';
```
maxinteger, line 38: why do we use `div` for list item?

If it is really a list item then it should be `li` element. 

You should either add [`role="listitem"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role) or give a different name the type, for example `TYPES.DIV` or `TYPES.GENERIC`.

Actually the same happens with `BUTTON_PRIMARY`

maxinteger (changes requested):
Could you please help me to understand what is the use case for this change?

weiwli2:
> Could you please help me to understand what is the use case for this change?

We use the <Text> component for most of the text content in our project. It provides a unified text class name that allows us to solve scenarios such as responsiveness. However, the HTML elements it contains are only h, p, and small. From the perspective of tag semantics, the scenarios it covers are not comprehensive enough. For example, the text in a button is neither a header nor a paragraph, and of course, it is not small, and the same is true for select list items. More importantly, for accessibility, the plug-in will increase the paragraph spacing, and the text in the button is not a paragraph. Using the <p> element will increase its margin-bottom, causing the text to exceed the button border. We cannot forcibly control its margin-bottom, such as using "margin-bottom:0! improtant", which will make the accessibility software function invalid on our page. I think the more correct approach is that the text in the button should use the <span> element instead of the paragraph (<p>) element. For more specific issues, you can check these two bugs:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-549379
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-506036

maxinteger:
@weiwli2 
Thanks for the explanation

maxinteger (approved):

PR #643, opened 2024-09-18
fix: preserve tabindex for specified elements

maxinteger:
# Description

TreeNodeBase change tabindex of all interactable elements, but this can change elements which are under focus management,
eg. Menu, nested lists, etc.

### DIFF
--- a/src/components/Tree/Tree.stories.tsx
+++ b/src/components/Tree/Tree.stories.tsx
@@ -18,6 +18,7 @@ import { TreeNodeRecord, TreeRefObject } from './Tree.types';
 import ButtonPill from '../ButtonPill';
 import Flex from '../Flex';
 import { v4 as uuidV4 } from 'uuid';
+import { PRESERVE_TABINDEX_CLASSNAME } from '../../utils/navigation';
 
 // prettier-ignore
 const exampleTree =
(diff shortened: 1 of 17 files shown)


src/components/TreeNodeBase/TreeNodeBase.stories.docs.mdx
```diff
@@ -10,6 +10,10 @@ After the node selected the user can use the `Tab` key to select the first, seco
 When the last interactable element is selected, the next tab will move the focus to the next interactable
 element after the tree.
 
+`<TreeNodeBase />` might contains other components which has their own focus management, and they're updating tabindex
```
gabrielchl, line 13:
```suggestion
`<TreeNodeBase />` might contain other components which have their own focus management, and they're updating tabindex
```

src/components/Tree/Tree.stories.tsx
```diff
@@ -89,6 +89,7 @@ const ExampleTreeNode = ({ node }: ExampleTreeNodeProps) => {
           </div>
           {nodeDetails.isLeaf && (
             <MenuTrigger
+              className="data-preserve-tabindex"
```
gabrielchl, line 92: should this be just `data-preserve-tabindex` as an attr or `md-nav-preserve-tabindex`?

maxinteger: nice catch, it should be `md-nav-preserve-tabindex` just forgot to update

gabrielchl (approved):

PR #644, opened 2024-09-18
feat(meeting-row-content): add MeetingRowContent component and consume in MeetingListItem

jor-row:
# Description

This is to fix a bug in ActivityList refactor where the MeetingListItem is causing issues as it's not immediately wrapped in a ListNext component. To solve this we will just use MeetingRowContent in Cantina inside ActivityList and style it as if it were a MeetingListItem.


maxinteger (approved):
Looks good

PR #649, opened 2024-09-20
feat(icon): add brand visuals icons to icon component

gabrielchl:
# Description
- add brand-visuals library
- add prop to Icon component to load icons from the brand-visuals library

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-564427

### DIFF
--- a/src/hooks/useDynamicSVGImport.ts
+++ b/src/hooks/useDynamicSVGImport.ts
@@ -1,5 +1,6 @@
 import { useRef, useState, useEffect } from 'react';
 import { useIsMounted } from './useIsMounted';
+import { IconLibrary } from 'src/components/Icon/Icon.types';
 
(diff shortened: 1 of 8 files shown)


src/hooks/useDynamicSVGImport.ts
```diff
@@ -24,7 +25,8 @@ interface UseDynamicSVGImportReturn {
  */
 function useDynamicSVGImport(
   name: string,
-  options: UseDynamicSVGImportOptions = {}
+  options: UseDynamicSVGImportOptions = {},
+  library: IconLibrary
```
maxinteger, line 29: recommendation: the `options` parameter should be the last one, so `library` should come before `options` or even before `name` since we need to pick the library first and then the icon from it.

maxinteger (approved):
nice one!

PR #652, opened 2024-09-24
feat(content-separator): allow ContentSeparator in Menus

dominiccarrington:
# Description

Update \<ContentSeparator\> to be classed as a valid component in \<Menu\>, \<MenuSection\> and \<MenuSelectionGroup\>


src/components/Menu/Menu.utils.test.tsx
```diff
@@ -10,55 +10,100 @@ jest.mock('@react-stately/collections', () => ({
...
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore - We don't export this method in the types
+    const generator = MenuSeperator.getCollectionNode(props);
+    const results = Array.from(generator);
+
+    expect(results).toHaveLength(1);
+    expect(results[0]).toEqual({
+      type: 'seperator',
+      props: expect.any(Object),
```
nataliadelmar, line 103: Could we do: 
    expect(results[0]).toEqual({
      type: 'seperator',
      props: props,
    });

And avoid the next 2 lines below?

src/components/Menu/Menu.unit.test.tsx
```diff
@@ -344,6 +360,38 @@ describe('<Menu />', () => {
...
+      await user.keyboard('{ArrowDown}');
+
+      expect(menuItems[0]).toHaveFocus();
+
+      await user.keyboard('{ArrowRight}');
+
+      expect(menuItems[0]).toHaveFocus();
+    });
+
     it('should handle up/down arrow keys correctly - for vertical menu with section', async () => {
```
nataliadelmar, line 395: would be also great to add MenuSeparator in this test

nataliadelmar: as well as in the selectiongroup test below

src/components/Menu/Menu.stories.tsx
```diff
@@ -256,14 +225,10 @@ SelectionGroups.parameters = {
...
           title={
-            <>
-              <ListHeader outline={true} outlinePosition="top" outlineColor="secondary">
-                <ListItemBaseSection position="start">
-                  <Icon scale={16} name="accessibility" strokeColor="none" />
-                </ListItemBaseSection>
-                <ListItemBaseSection position="fill">Layout</ListItemBaseSection>
-              </ListHeader>
-            </>
+            <Flex direction="row" alignItems="center" xgap="0.25rem">
```
nataliadelmar, line 228: nothing to do with your PR, but could you remove the aria-labels from all selectionGroups on this story? They all have a visible title that labels the group, so the aria-label is lost anyways and it is confusing to have it there

nataliadelmar (commented):
lgtm! only some non-blocking comments :D

src/components/Menu/Menu.utils.test.tsx
dominiccarrington: Simplified the test

src/components/Menu/Menu.unit.test.tsx
dominiccarrington: Done!

nataliadelmar (approved):

src/components/Menu/Menu.utils.test.tsx
```diff
@@ -1,64 +1,101 @@
...
+    resultNodes.forEach((node, index) => {
+      expect(node).toEqual({
+        ...mockNodes[index],
+        props: { ...selectionGroupProps, selectionGroup: true },
       });
     });
+  });
+});
+
+describe('ContentSeparator', () => {
```
shnaaz, line 66: I think this test should live in ContentSeparator.unit.test.tsx

dominiccarrington: Moved

src/components/Menu/Menu.utils.test.tsx
```diff
@@ -1,64 +1,64 @@
 import React from 'react';
 import { SelectionGroup } from './Menu.utils';
-import { Item , Section, PartialNode } from '@react-stately/collections';
-import {  SelectionMode } from '@react-types/shared';
+import { Item, Section, PartialNode } from '@react-stately/collections';
+import { SelectionMode } from '@react-types/shared';
 import { SelectionGroupProps } from './Menu.types';
+import ContentSeparator from '../ContentSeparator';
```
nataliadelmar, line 6: do we need this import for something?

src/components/MenuSelectionGroup/MenuSelectionGroup.style.scss
```diff
@@ -10,4 +10,8 @@
   font-size: 0.875rem;
   line-height: 1.375rem;
   padding: 0.25rem 0.75rem;
+
+  .md-text-wrapper {
+    font-size: 0.875rem;
```
nataliadelmar, line 15: are we using the Text component anywhere in the header?
font-size should normally be set on the Text component as type="subheader/title/etc"
I'd prefer not setting it in css

src/components/Menu/Menu.utils.test.tsx
dominiccarrington: Not anymore

src/components/MenuSelectionGroup/MenuSelectionGroup.style.scss
dominiccarrington: We are using the Text component in the SelectionGroup stories. The same font-size can be achieved with type="body-secondary". I have reverted this change.

shnaaz:
Tested with Cantina, works fine

shnaaz (approved):

PR #654, opened 2024-09-25
feat(icon): add brand visuals icons to icon component

gabrielchl:
original PR (https://github.com/momentum-design/momentum-react-v2/pull/649) notes:

# Description
- add brand-visuals library
- add prop to Icon component to load icons from the brand-visuals library

note: the new, larger icons included in the new brand-visuals library will not show correctly with the icons component, but this change should be sufficient for our usage in cantina for now

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-564427

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "@momentum-design/animations": "^0.0.4",
+    "@momentum-design/brand-visuals": "^0.0.24",
     "@momentum-design/icons": "0.0.164",
     "@momentum-design/tokens": "0.0.77",
     "@momentum-ui/design-tokens": "^0.0.63",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -95,6 +95,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "@momentum-design/animations": "^0.0.4",
+    "@momentum-design/brand-visuals": "^0.0.24",
```
cscheffauer, line 98: brand visuals are different in terms of sizing and color as Icons are - we don't want to use the same component for both. I would recommend to create a separate component if there is a need for it.

cscheffauer (changes requested):

gabrielchl:
not needed for now

PR #657, opened 2024-09-30
fix: tree component focus loss issue

maxinteger:
# Description

In some case, for example on high CPU load the scroll to active node can not keep up with the user input and the active element removed because it is not in the view.

- check and scroll to active element when it is not in the DOM after each arrow key event.
- optimize MutationObserver creation


gabrielchl (approved):

gabrielchl:
tested with trees in cantina, seemed to work fine :)

gabrielchl (changes requested):
sry, just noticed a bug, steps to repro.:
1. go to a team with a lot of spaces, select one of them
2. scroll that item out of view
3. click any other item
4. notice that the newly selected item is correctly toggled, but the originally selected item is brought back into view

gabrielchl (approved):
thx for fixing :)

PR #659, opened 2024-10-02
feat(tree-node-base): add context menu support

gabrielchl:
# Description
the space list had custom context menus via list item base
tree node base does not
this PR extracts the context menu into a separate component and adds it to tree node base
also changed the positioning of the context menu to be relative to the item in question instead of fixed to the page

# Links
as part of the work for https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-503138


gabrielchl:
closing, we're going with another approach, replacing the context menu with a 3 dots button

PR #662, opened 2024-10-03
feat(space-row-content): replace context menu with 3 dots button

gabrielchl:
# Description
space list item had context menu, it's not accessible
as part of the work to convert space list to tree, we're replacing the context menu to a 3-dots button that opens a menu

non-compact
Screenshot 2024-10-03 at 16 07 42

compact
Screenshot 2024-10-03 at 16 07 51

# Links
as part of the work for https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-503138


src/components/SpaceRowContent/SpaceRowContent.unit.test.tsx
```diff
@@ -497,4 +499,25 @@ describe('<SpaceRowContent />', () => {
       expect(mentionIcon.props().fillColor).toBe('var(--mds-color-theme-text-primary-disabled)');
     });
   });
+
+  describe('actions', () => {
```
maxinteger, line 503: Please add a snapshot test as well

maxinteger (changes requested):
It looks good, but I found some minor problems.

Need some additional testing. And the SpaceListItem story does not work as expected.

image

When I press Tab it does not focus the "..." button. It is strange because it does work in the SpaceTreeNode storybook.

maxinteger (approved):
Nice job!

PR #664, opened 2024-10-04
chore(tooling): update node and typescript

maxinteger:
# Description

- updates nodejs to the next lts version, 22
- updated TS to the make it compatible with @types/node@22


cscheffauer (commented):
Just a comment (haven't reviewed the changes):
Can we make sure this thread is addressed?
https://stackoverflow.com/a/70020984
Like the "moduleResolution" is set correctly?
thanks!

maxinteger:
Please note,

- when you test this PR you need to remove the `node_modules` and install everything again
- if you see `Error: error:0308010C:digital envelope routines::unsupported` then you have to add the `NODE_OPTIONS=--openssl-legacy-provider` before run yarn, eg: `NODE_OPTIONS=--openssl-legacy-provider yarn start`

maxinteger:
> Just a comment (haven't reviewed the changes): Can we make sure this thread is addressed? https://stackoverflow.com/a/70020984 Like the "moduleResolution" is set correctly? thanks!

Thanks for the tip

I updated the `moduleResolution` to `node16`, but I did not changed the exports in the package.json. That will be another PR

chrisadubois:
did you test this with the webclient? else lgtm

cscheffauer (approved):

PR #665, opened 2024-10-04
Revert "feat(space-row-content): replace context menu with 3 dots button"

gabrielchl:
Reverts momentum-design/momentum-react-v2#662


gabrielchl:
seems like this won't be needed :)

PR #670, opened 2024-10-14
feat(Tabs): create TabList component to abstract a11y requirements for tabs

dominiccarrington:
# Description

This PR adds three components:
- `<TabList />`
  - Exported as `<TabListNext />` 
  - Wrapper around `<ButtonGroupNext />` that adds properties for tab lists and adds the state based properties on child tabs
- `<TabPanel />` 
  - Exported as `<TabPanelNext />` 
  - Wrapper around a div or section
- `<TabsProvider />`
  - Exported as `<TabsProviderNext />` 
  - A context provider

This PR also makes the `<Tab />` component pass the disabled prop to the Button it wraps.

# Links
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ - Tabs Pattern
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-367582 - Jira ticket it will help to address


gabrielchl (commented):
very nice! i think we should probably have a storybook page to showcase <Tabs /> as well :)

src/components/TabList/TabList.utils.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const useTabsContext: () => TabsContextValue = () => useContext(TabsContext);
+
+/**
+ * The Tabs Component
+ */
+export const Tabs: FC<TabsProps> = (props: TabsProps) => {
+  const { selectedTab, id: _id, children } = props;
+
+  const id = useRef(_id || uuidv4());
```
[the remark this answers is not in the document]
dominiccarrington: Thanks, changed to use useState instead

src/components/TabList/TabList.tsx
```diff
@@ -0,0 +1,127 @@
...
+      return element;
+    }
+
+    const isActiveTab = tabsContext.selectedTab === elementKey;
+
+    return React.cloneElement(element, {
+      ...commonProps,
+      active: isActiveTab,
+      'aria-selected': isActiveTab,
+      'aria-controls': isActiveTab ? tabsContext.activePanelId : undefined,
```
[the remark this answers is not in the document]
dominiccarrington: In the react-aria example ([here](https://react-spectrum.adobe.com/react-aria/examples/category-tabs.html)) they don't have an aria-controls property. I suspect this is because the element being referenced isn't in the DOM whereas in the WAI example all the tabs are in the DOM.

src/components/TabList/TabList.stories.docs.mdx
```diff
@@ -0,0 +1,76 @@
+The `<TabList />` component provides a wrapper around the `<ButtonGroupNext />` component that
+implements the accessibility requirements of a
+[tablist](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).
+
+This component has been created for use inside and outside the `<Tabs>` component.
```
nataliadelmar, line 5: can you add what determines if I should use it inside or outside a Tabs component?

nataliadelmar: oh I saw you added it after. Then we can change this phrase to:
This component has been created for use:
a) inside the `<Tabs>` component (Controlled)
b) outside the `<Tabs>` component (Uncontrolled)

nataliadelmar, line 5: do you refer to Tabs or TabsNext? I see that in some places your using the exported name of the component ***Next and in others you are using the internal name of the component. Let's stick to one approach only

src/components/TabPanel/TabPanel.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  component: TabPanel,
+  parameters: {
+    expanded: true,
+    docs: {
+      page: DocumentationPage(Documentation),
+    },
+  },
+};
+
+// NOTE: Primary story. This renders a single component with all external props.
```
nataliadelmar, line 19: Delete note

src/components/TabsProvider/TabsProvider.stories.args.ts
```diff
@@ -0,0 +1,41 @@
+const tabsArgTypes = {
+  children: {
+    description: 'Provides the child nodes for this element.', // NOTE: Description of this prop.
+    control: { type: 'text' }, // NOTE: Control type for this prop.
```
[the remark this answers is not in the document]
nataliadelmar: Oh I didn't mean to delete the arg, just to delete the control prop from it. The developer still should know the children of type ReactElement are expected

nataliadelmar (approved):
Approved with 1 minor unresolved comment

PR #680, opened 2024-10-24
feat(tab-list): add hasBackground prop

nataliadelmar:
# Description

Sometimes tab lists have a background color under them. We want to standarize the application of this background color so that it looks the same everywhere in the app.

### DIFF
--- a/src/components/TabList/TabList.style.scss
+++ b/src/components/TabList/TabList.style.scss
@@ -1,10 +1,12 @@
 .md-tab-list-wrapper {
   &[data-has-background='true'] {
     width: 100%;
-    height: 2.75rem;
-    padding-right: 0.5rem;
-    padding-left: 0.5rem;
+    padding-top: 0.25rem;
+    padding-bottom: 0.25rem;
+    padding-right: 0.25rem;
+    padding-left: 0.25rem;
     background-color: var(--mds-color-theme-background-primary-hover);
     border-radius: 8rem;
+    flex-shrink: 0;
   }
 }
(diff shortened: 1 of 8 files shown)


src/components/TabList/TabList.style.scss
```diff
@@ -1,2 +1,10 @@
 .md-tab-list-wrapper {
+  &[data-has-background='true'] {
+    width: 100%;
+    height: 2.75rem;
```
lismith2-cisco, line 4: This is 3rem in meetings tab for some reason, which I think looks a little better, but that might just be because I've been staring at it so long.

src/components/TabList/TabList.style.scss
```diff
@@ -1,2 +1,10 @@
...
+    padding-right: 0.5rem;
+    padding-left: 0.5rem;
+    background-color: var(--mds-color-theme-background-primary-hover);
+    border-radius: 8rem;
+  }
```
lismith2-cisco, line 9: could you add min-height: 2.75rem (or 3rem) because otherwise it flexes smaller in the calendar and becomes a weird grey line behind the tab buttons.

lismith2-cisco (approved):
I do request small changes! But I'm away tomorrow, and it'll be fine once you've fixed a couple of things.

src/components/TabList/TabList.style.scss
nataliadelmar: added padding-bottom and top instaed

src/components/TabList/TabList.style.scss
nataliadelmar: added flex-shrink instaed

PR #682, opened 2024-10-25
fix(tree): stop scrolling away when selecting a new node in virtualized tree

gabrielchl:
# Description

bug: (virtualized tree) after clicking on a node, then scrolling away to click on another node, the tree attempts to scroll back to the last selected node

cause: for keyboard users, it's important that the node is rendered in the virtualized tree, thus there's some code that makes the virtualized tree scroll to the active node if focus comes within the tree. however, this caused the bug above

fix: we don't do the scrolling thing IF the focus is on / within another tree node

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-573336

### DIFF
--- a/src/components/Tree/Tree.test.tsx
+++ b/src/components/Tree/Tree.test.tsx
@@ -548,6 +548,27 @@ describe('<Tree />', () => {
       expect(scrollToNode).toHaveBeenCalledTimes(0);
     });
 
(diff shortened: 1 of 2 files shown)


dominiccarrington (approved):
LGTM, tested in Cantina and fixes the bug.

Just a typo

PR #683, opened 2024-10-29
fix(SPARK-573927): add text color judgment in light theme

zheng-xiaoyang:
# IMPORTANT

# Guide

```
{type}/{scope?}/{message}
```

# Description
    Original behavior: Overwrite tooltip text color, no judgment on tooltip text color change, use the color inherited from the parent element
Now changed to: When the theme is lightWebex, add style={{color:""}} attribute to each tooltip, because the inline style has a higher weight than the style inherited from the parent, so the text color will not be overwritten
# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-573927?filter=-1

### DIFF
--- a/src/legacy/Lightbox/index.js
+++ b/src/legacy/Lightbox/index.js
@@ -501,7 +501,6 @@ class Lightbox extends React.Component {
                   className="md-lightbox__control"
                   data-test="zoom-out-button"
                   onPress={() => this.setZoom(-0.25)}
-                  color="red"
                 >
                   <IconNext
                     name="zoom-out"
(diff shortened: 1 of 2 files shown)


src/legacy/Lightbox/index.js
```diff
@@ -27,6 +27,7 @@ class Lightbox extends React.Component {
       height: 600,
     },
     zoom: 1,
+    theme: 'lightWebex',
```
JudyZhuHz, line 30: Why need add theme to state?

JudyZhuHz (approved):

PR #685, opened 2024-10-30
feat: themeprovider migration

cscheffauer:
# Description

- Using the new Momentum Design Themeprovider component, see consumption docs [here](https://momentum-design.github.io/momentum-design/en/components/)
- Removed Inter file and using new Inter from momentum design instead
- Upgrade of tokens
- Lightbox fixes due to issues discovered in Bug bash
- Few scss fixes had to be made


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,9 +1,18 @@
 import React, { FC } from 'react';
-
+import classNames from 'classnames';
+// TODO: legacy tokens - to be removed:
```
maxinteger, line 3: Did you created task for these TODOs?

maxinteger (approved):

PR #686, opened 2024-10-30
fix(TabsProvider): remove uuidv4 call every render

dominiccarrington:
# Description

- Fixes TabsProvider story so that it updates with props
- Remove uuidv4 call every rerender


src/components/TabsProvider/TabsProvider.tsx
```diff
@@ -10,7 +10,8 @@ import { TabsProviderContext } from './TabsProvider.utils';
 const TabsProvider: FC<Props> = (props: Props) => {
   const { selectedTab, id: _id, children } = props;
 
-  const [id] = useState(_id || uuidv4());
```
gabrielchl, line 13: i don't believe the original implementation here is calling the uuidv4 function on every render.
useState only evaluates the initial state once on first render, then keep tracks of the value internally

PR #690, opened 2024-11-04
fix(TabList): require aria-label (or equivalent)

dominiccarrington:
# Description

Introduce a AriaLabelRequired type and require aria-label (or equivalent) for TabList

# Links

> If the tab list has a visible label, the element with role tablist has aria-labelledby set to a value that refers to the labelling element. Otherwise, the tablist element has a label provided by aria-label.
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/


src/utils/a11y.ts
```diff
@@ -0,0 +1,24 @@
...
+      'aria-labelledby': AriaAttributes['aria-labelledby'];
+    }
+  | {
+      /**
+       * Defines text representing advisory information related to the element it belongs to.
+       *
+       * @deprecated - Use aria-label or aria-labelledby, using title has a11y concerns
+       * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns
+       */
+      title: string;
```
lismith2-cisco, line 23: why allow this?

dominiccarrington: Removed

src/components/TabList/TabList.tsx
```diff
@@ -27,6 +27,10 @@ const TabList: FC<Props> = (props: Props) => {
     ? (_children as ReactElement<TabProps>[])
     : [_children];
 
+  if (otherProps['aria-labelledby'] === undefined && otherProps['aria-label'] === undefined) {
+    console.warn('MRV2: TabList requires aria-labelledby or aria-label.');
+  }
+
```
gabrielchl, line 33: so empty string fails:
```suggestion
  if (!otherProps['aria-labelledby'] && !otherProps['aria-label']) {
    console.warn('MRV2: TabList requires aria-labelledby or aria-label.');
  }

```
also can we make this a util function? so other components can use this check too

dominiccarrington: Moved!

gabrielchl (approved):

PR #692, opened 2024-11-08
fix(ScreenReaderAnnouncer): fix storybook docs page

dominiccarrington:
# Description

This PR fixes the Screen Reader Announcer docs page. Currently it crashes when trying to load due to duplicate screen reader identities. This PR makes the doc use iframes instead of inline loading the components.


gabrielchl (approved):
noice-nice

PR #693, opened 2024-11-12
fix(types): abstract RequireEither type util 

dominiccarrington:
# Description
This PR does the following:
- Abstract out the type structure for AriaLabelRequired to allow any options.
- Bump Prettier version to last version in 2.x as the current version wasn't able to parse the type definition


src/utils/a11y.ts
```diff
@@ -1,18 +1,22 @@
...
+/**
+ * Utility type to require one or more of a set of attributes.
+ *
+ * For example, this is used to check that `aria-label` or `aria-labelledby` is defined on components when consumed.
+ * `RequireEither<AriaAttributes, ['aria-label', 'aria-labelledby']>`
+ *
+ * @typeParam TFrom - Used to get the type of the required attributes
+ * @typeParam TAttributes - One of these attributes satifies this requirement. Must be given as an array of attributes.
+ */
+export type RequireEither<TFrom, TAttributes extends (keyof TFrom)[]> = TAttributes extends [
```
lismith2-cisco, line 12: Might be better to call it RequireOneOf as you can give more than 2

src/utils/a11y.ts
```diff
@@ -1,18 +1,22 @@
...
+  infer TFirst extends keyof TFrom,
+  ...infer TLast extends (keyof TFrom)[]
+]
+  ? { [K in TFirst]-?: TFrom[TFirst] } | RequireEither<TFrom, TLast>
```
lismith2-cisco, line 16: suggest leaving comments on the weirder things you are doing - the { [K in TFirst]-?: TFrom[TFirst] } expanding out even though there's only one element, the -? is quite unsearchable

src/utils/a11y.ts
```diff
@@ -1,18 +1,22 @@
...
-      /**
-       * Identifies the element (or elements) that labels the current element.
-       */
-      'aria-labelledby': AriaAttributes['aria-labelledby'];
-    };
+/**
+ * Utility type to require one or more of a set of attributes.
+ *
+ * For example, this is used to check that `aria-label` or `aria-labelledby` is defined on components when consumed.
+ * `RequireEither<AriaAttributes, ['aria-label', 'aria-labelledby']>`
```
lismith2-cisco, line 7: Suggest adding the return value for this example to this comment

lismith2-cisco (changes requested):

lismith2-cisco (changes requested):

lismith2-cisco (approved):

PR #698, opened 2024-11-14
fix(legacy/list-separator): remove low font-weight

dominiccarrington:
# Description

Remove `font-weight: 250` from legacy list separator component.

Before:
Screenshot 2024-11-14 at 15 58 11

After:
Screenshot 2024-11-14 at 16 27 49

# Links
Jira Ticket: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-579344
Figma Design: https://www.figma.com/design/FmXv8xvun1xAZiOMfztwqg/%F0%9F%A7%A9-Webex-App---Web?m=auto&node-id=5585-13072&t=oauxKqnz3RqbQApS-1

### DIFF
--- a/scss/components/list-separator/list-separator.scss
+++ b/scss/components/list-separator/list-separator.scss
@@ -30,10 +30,6 @@
 
     &__text {
       padding: 0 8px;
-      font-family: $brand-font-regular;
-      font-size: 13px;
-      line-height: 1.5em;
-      letter-spacing: 0.015em;
       color: $md-gray-70;
     }
   }
--- a/scss/components/list-separator/list-separator.scss
+++ b/scss/components/list-separator/list-separator.scss
@@ -32,7 +32,6 @@
       padding: 0 8px;
       font-family: $brand-font-regular;
       font-size: 13px;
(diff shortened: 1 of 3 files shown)


scss/components/list-separator/list-separator.scss
```diff
@@ -32,7 +32,6 @@
       padding: 0 8px;
       font-family: $brand-font-regular;
       font-size: 13px;
-      font-weight: 250;
```
cscheffauer, line 35: just a note - the way this is styled is not really great. Ideally the Text component should be used and we just choose the right type. Then we don't need to set font-family, font-size, font-weight, line-height at all here.
Maybe if u want you could change this. Also letter-spacing seems to be weird.

cscheffauer (changes requested):

scss/components/list-separator/list-separator.scss
dominiccarrington: Updated to use Text component

cscheffauer (approved):

nataliadelmar (approved):

PR #699, opened 2024-11-15
feat: text-component-migration

cscheffauer:
# Description

- Using the new MdcText component from momentum-design
- Bumped momentum-design version to get latest version
- Fixed existing Unit tests
- Fixed ThemeProvider test cause found a better way to test web components in RTL
- Updated snapshots
- Its possible now to pass in both the old and new Text types to the component.


src/components/Text/Text.stories.tsx
```diff
@@ -41,92 +43,38 @@ Types.parameters = {
   ],
```
nataliadelmar, line 43: argTypes of all stories has type still pointing to only the current 'old' text types, and not to the new types
image

nataliadelmar: this causes the dropdown to only show the old types as alternatives:
image

src/components/Text/Text.stories.tsx
```diff
@@ -6,6 +6,8 @@ import Text, { TextProps } from './';
 import argTypes from './Text.stories.args';
```
nataliadelmar, line 6: this is not your fault, but args is missing tagName for the user to change the tagName

nataliadelmar (commented):
Question:

With these changes, when you render the component, now no type is wrapped in an html text tag like before. I remember that a11y was asking us to be able to navigate the page using the tags. Will that still be possible now?

Before:
image

 After:
image

nataliadelmar:
> Question:
> 
> With these changes, when you render the component, now no type is wrapped in an html text tag like before. I remember that a11y was asking us to be able to navigate the page using the tags. Will that still be possible now?
> 
> Before: image
> 
> After: image

Regarding this comment, the tags are in the dom, they were just inside the shadow dom. Example:
image

nataliadelmar (approved):
non-blocking comments but lgtm

PR #707, opened 2024-11-21
ci: add stylelint-rem-over-px rule

dominiccarrington:
# Description

Add the stylelint rule rem-over-px

Also:
- Defines the file extensions which should be checked in vscode by the stylelint extension
- Adds a lint:css script and add it to the lint script


.stylelintrc.json
```diff
@@ -30,6 +32,13 @@
         "ignore": ["after-comment", "first-nested"],
         "severity": "warning"
       }
+    ],
+    "rem-over-px/rem-over-px": [
+      true,
+      {
+        "severity": "warning",
+        "disableFix": true
```
nataliadelmar, line 40: do we want to disabled having the possibility to accept it?

nataliadelmar (approved):

.stylelintrc.json
dominiccarrington: `disableFix` stops the plugin from automatically fixing the issues on commit since we want to fix them manually (or case by case)

PR #708, opened 2024-11-22
fix(TextInput): make label and clearAriaLabel required and provide options for autoComplete prop

dominiccarrington:
# Description
This PR adds the following:
 - Require one of label, aria-label or aria-labelledby (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582964)
 - Makes TextInput.clearAriaLabel prop required (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582967)
 - Adds guide to the autoComplete prop (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582965)
 - Add aria-invalid if the messages provided are errors (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582957)


src/components/TextInput/TextInput.types.ts
```diff
@@ -14,7 +76,7 @@ export interface Props extends Omit<AriaTextFieldProps, 'errorMessage'> {
   /**
    * Optional aria label on the clear button
```
gabrielchl, line 77:
```suggestion
   * Aria label on the clear button
```

src/components/ComboBox/ComboBox.tsx
```diff
@@ -192,246 +196,278 @@ const ComboBox: React.FC<Props> = (props: Props) => {
...
     <>
-      {label ? (<div className={STYLE.label}>{label}</div>) : null}
+      {label ? <div className={STYLE.label}>{label}</div> : null}
       <div {...wrapperProps} ref={containerRef}>
         <div className={STYLE.inputSection}>
           <TextInput
+            ref={inputRef}
             aria-label={STYLE.input}
+            // ComboBox hides the clear button from the input therefore set the clear aria label to undefined
+            clearAriaLabel={undefined}
```
gabrielchl, line 465: i must be missing something but how is it not upset we're passing undefined in a string field

dominiccarrington: Ah! It's because we don't have `strictNullChecks` turned on in tsconfig. Therefore undefined and null are valid for any type.

src/components/TextInput/TextInput.types.ts
```diff
@@ -1,8 +1,70 @@
...
+  | 'tel-country-code'
+  | 'tel-national'
+  | 'tel-area-code'
+  | 'tel-local'
+  | 'tel-local-prefix'
+  | 'tel-local-suffix'
+  | 'tel-extension'
+  | 'email'
+  | 'impp'
+  | (string & Record<never, never>);
```
dominiccarrington, line 65: This is to prevent type overreach.

If it was just string, we wouldn't get any autocomplete help since all the above would be merged into just `string`.

src/components/ComboBox/ComboBox.tsx
gabrielchl: wtf, nonetheless isn't this wrong then? maybe we can't make this prop mandatory?

dominiccarrington: I'll make it an empty string. The prop can be mandatory since this is the only exception and the clear button is hidden from tab order and sr.

gabrielchl (approved):

PR #712, opened 2024-11-25
fix(AriaGroup): require aria label

dominiccarrington:
# Description
- Add AriaLabelRequired
- Rename props from ariaLabelledBy to aria-labelledby

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582970

### DIFF
--- a/src/components/AriaGroup/AriaGroup.tsx
+++ b/src/components/AriaGroup/AriaGroup.tsx
@@ -4,18 +4,14 @@ import { Props } from './AriaGroup.types';
 
 /**
  * The AriaGroup component.
+ *
+ * Requires aria-label or aria-labelledby
  */
 const AriaGroup: FC<Props> = (props: Props) => {
-  const { ariaLabel, ariaLabelledby, children, className, ...otherProps } = props;
+  const { children, className, ...otherProps } = props;
 
   return (
-    <div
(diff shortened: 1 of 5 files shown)


src/components/AriaGroup/AriaGroup.tsx
```diff
@@ -1,27 +1,17 @@
-import React, { FC} from 'react';
+import React, { FC } from 'react';
 
 import { Props } from './AriaGroup.types';
 
 /**
  * The AriaGroup component.
+ *
+ * Requires aria-label or aria-labelledby
```
gabrielchl, line 8: nitpick, the type should be sufficient in documenting what is / isn't required

src/components/AriaGroup/AriaGroup.types.ts
```diff
@@ -1,16 +1,8 @@
+import { AriaLabelingProps } from '@react-types/shared';
 import { ReactNode } from 'react';
+import { AriaLabelRequired } from 'src/utils/a11y';
```
gabrielchl, line 3:
```suggestion
import { AriaLabelRequired } from '../../utils/a11y';
```
unfortunately my VS code does the same thing (sometimes), havn't found a solution yet

gabrielchl (approved):

PR #720, opened 2024-11-27
chore: install husky automatically

dominiccarrington:
# Description

As per Husky 7 docs (https://github.com/typicode/husky/tree/v7.0.4/docs#yarn-2), yarn 2+ doesn't support the `prepare` script therefore there is a workaround.
- Run `husky install` when postinstall is run
- Since this package is public and published on npm, use pinst to disable the postinstall script when the package is being published.

# Testing

To test this PR:
1. Clone this branch to a new folder
2. Confirm that .git/config doesn't have core.hooksPath set
3. Run `yarn`
4. Confirm that .git/config now has core.hooksPath set to .husky


package.json
```diff
@@ -53,8 +53,9 @@
     "lint:src": "tsc --noEmit --skipLibCheck && esw \"./src/**/*.{js,jsx,ts,tsx}\"",
     "lint:css": "cross-env stylelint 'src/**/*.scss'",
     "lint:watch": "yarn lint:src --watch",
-    "prepublishOnly": "yarn build",
-    "prepare": "husky install",
+    "prepublishOnly": "yarn build && pinst --disable",
+    "postinstall": "husky install",
+    "postpublish": "pinst --enable",
```
gabrielchl, line 58: how come you're using `prepublishonly` and `postpublish` instead of `prepack` and `postpack` as shown in pinst's doc?

dominiccarrington: `prepublishonly` and `postpublish` are the scripts defined in the husky 7 docs.

I'll take a look and see if one is better than the other.

dominiccarrington: prepublish shouldn't have side effects according to the yarn docs, moved the scripts to pre/post pack instead like the pinst docs.

PR #721, opened 2024-11-27
feat(Radio, RadioGroup): add label requirements and add description prop

dominiccarrington:
# Description
This PR addresses the following:
- Requires RadioGroup to have a label, aria-label or aria-labelledby (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582955)
- Requires Radio to have a label (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582955)
- Adds a description prop to the basic Radio styled per the Figma (https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582968)

# Links
Figma Design: https://www.figma.com/design/FmXv8xvun1xAZiOMfztwqg/%F0%9F%A7%A9-Webex-App---Web?m=auto&node-id=40968-67045&t=JBkDsxJT6cWwwE8F-1


src/components/RadioGroup/Radio.tsx
```diff
@@ -1,32 +1,54 @@
...
   const ref = useRef(null);
-  const { inputProps } = useRadio({ 'aria-label': label, ...props }, state, ref);
+
+  const [radioUuid] = useState(uuidv4);
+  const radioId = id || radioUuid;
+
+  const radioProps = {
+    'aria-label': label,
+    ...(description ? { 'aria-describedby': `radio-description-${radioId}` } : {}),
+    ...props,
```
nataliadelmar, line 25: if we pass our own aria-describedby AND we pass description, aria-describedby prop would overwrite line 25. We should make it work so that both descriptions ids end up in the aria-describedby string

nataliadelmar (approved):

PR #726, opened 2024-12-02
feat(modal-container): add title on h2 tag

nataliadelmar:
# Description

If there is a visible title on a modal, it should be on an H2 tag and it should by default be the label of the modal unless aria-label or aria-labelledby are specified.

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582989
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582910


nataliadelmar:
closing for https://github.com/momentum-design/momentum-react-v2/pull/737

PR #732, opened 2024-12-12
feat(accessible-names): require accessible names in non-text wrapping components

nataliadelmar:
DRAFT

### DIFF
--- a/src/components/AddReactionButton/AddReactionButton.tsx
+++ b/src/components/AddReactionButton/AddReactionButton.tsx
@@ -6,10 +6,14 @@ import { DEFAULTS, STYLE } from './AddReactionButton.constants';
 import { Props } from './AddReactionButton.types';
 import './AddReactionButton.style.scss';
 import Icon, { IconScale } from '../Icon';
+import { useCheckAriaLabel } from '../../utils/a11y';
 
 const AddReactionButton = forwardRef((props: Props, ref: ForwardedRef<HTMLButtonElement>) => {
   const { className, id, style, ...otherProps } = props;
   delete otherProps.size;
+
+  useCheckAriaLabel('AddReactionButton', props);
+
   return (
     <ButtonCircle
       ref={ref}
--- a/src/components/AriaToolbar/AriaToolbar.tsx
+++ b/src/components/AriaToolbar/AriaToolbar.tsx
@@ -13,7 +13,6 @@ import { getKeyboardFocusableElements } from '../../utils/navigation';
  */
 const AriaToolbar: FC<Props> = (props: Props) => {
   const {
-    ariaLabel,
     className,
     id,
     style,
@@ -21,7 +20,6 @@ const AriaToolbar: FC<Props> = (props: Props) => {
     orientation = DEFAULTS.ORIENTATION,
     shouldRenderAsButtonGroup = DEFAULTS.SHOULD_RENDER_AS_BUTTON_GROUP,
     onTabPress,
-    ariaControls,
     buttonGroupProps,
     ariaToolbarItemsSize,
     ...rest
@@ -52,8 +50,6 @@ const AriaToolbar: FC<Props> = (props: Props) => {
     className: classnames(className, STYLE.wrapper),
     id: id,
     style: style,
-    'aria-label': ariaLabel,
-    'aria-controls': ariaControls,
     role: 'toolbar',
     'aria-orientation': orientation,
     ...rest,
--- a/src/components/AriaToolbar/AriaToolbar.types.ts
+++ b/src/components/AriaToolbar/AriaToolbar.types.ts
@@ -1,8 +1,10 @@
 import { CSSProperties, MutableRefObject, ReactElement } from 'react';
 import { ButtonGroupProps } from '../ButtonGroup';
 import { SupportedComponents } from '../ButtonGroup/ButtonGroup.types';
+import { AriaLabelingProps } from '@react-types/shared';
+import { AriaLabelRequired } from 'src/utils/a11y';
 
-export interface Props {
+interface AriaToolbarProps extends AriaLabelingProps {
   /**
    * Child components of this AriaToolbar.
    */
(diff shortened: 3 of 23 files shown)


src/components/AddReactionButton/AddReactionButton.tsx
```diff
@@ -6,10 +6,14 @@ import { DEFAULTS, STYLE } from './AddReactionButton.constants';
 import { Props } from './AddReactionButton.types';
 import './AddReactionButton.style.scss';
 import Icon, { IconScale } from '../Icon';
+import { useCheckAriaLabel } from '../../utils/a11y';
 
 const AddReactionButton = forwardRef((props: Props, ref: ForwardedRef<HTMLButtonElement>) => {
   const { className, id, style, ...otherProps } = props;
   delete otherProps.size;
+
+  useCheckAriaLabel('AddReactionButton', props);
```
nataliadelmar, line 15: we can't do type requirement because some (most) times, ButtonCircles should be labelled by its Tooltip. So ButtonCircle would not have aria-label or aria-labelledby directly and TS will complain

src/components/AriaToolbar/AriaToolbar.tsx
```diff
@@ -13,15 +13,13 @@ import { getKeyboardFocusableElements } from '../../utils/navigation';
...
   const {
-    ariaLabel,
     className,
     id,
     style,
     children,
     orientation = DEFAULTS.ORIENTATION,
     shouldRenderAsButtonGroup = DEFAULTS.SHOULD_RENDER_AS_BUTTON_GROUP,
     onTabPress,
-    ariaControls,
```
nataliadelmar, line 24: we have decided not to use camelCase anymore on aria properties, to keep them consistent. I know this would cause a breaking change, so Im preparing the Cantina PR for it.

src/components/AriaToolbar/AriaToolbar.types.ts
```diff
@@ -67,6 +59,8 @@ export interface Props {
   ariaToolbarItemsSize: number;
 }
 
+export type Props = AriaToolbarProps & AriaLabelRequired;
```
nataliadelmar, line 62: aria-toolbar gives the wrapper a role of 'toolbar'. The SR will read toolbar with an empty label if the developer doesn't pass any aria-label(lledby) to the Toolbar.

src/components/ButtonCircle/ButtonCircle.tsx
```diff
@@ -24,6 +25,8 @@ const ButtonCircle = forwardRef((props: Props, providedRef: RefObject<HTMLButton
     console.warn('MRV2: Momentum does not support a ghost inverted ButtonCircle.');
   }
 
+  useCheckAriaLabel('ButtonCircle', props);
```
nataliadelmar, line 28: ButtonCircle can be labelled by a tooltip, so we won't require aria-label(ledby) to it. I have added a story to Tooltip that we don't get a console warning only when we trigger a label tooltip from the ButtonCircle

src/components/AddReactionButton/AddReactionButton.tsx
nataliadelmar: I have added a story to Tooltip that we don't get a console warning only when we trigger a label tooltip from the ButtonCircle

src/components/ButtonCircleLink/ButtonCircleLink.tsx
```diff
@@ -26,6 +27,8 @@ const ButtonCircleLink = forwardRef((props: Props, providedRef: RefObject<HTMLAn
     ...otherProps
   } = props;
 
+  useCheckAriaLabel('ButtonCircleLink', props);
```
nataliadelmar, line 30: same as ButtonCircle. ButtonCircle can be labelled by a tooltip, so we won't require aria-label(ledby) to it. I have added a story to Tooltip that we don't get a console warning only when we trigger a label tooltip from the ButtonCircle

src/components/ButtonCircleToggle/ButtonCircleToggle.tsx
```diff
@@ -23,6 +24,8 @@ const ButtonCircleToggle = forwardRef((props: Props, providedRef: RefObject<HTML
     ...otherProps
   } = props;
 
+  useCheckAriaLabel('ButtonCircleToggle', props);
```
nataliadelmar, line 27: Same as ButtonCircle. ButtonCircle can be labelled by a tooltip, so we won't require aria-label(ledby) to it. I have added a story to Tooltip that we don't get a console warning only when we trigger a label tooltip from the ButtonCircle

src/components/ButtonControl/ButtonControl.tsx
```diff
@@ -8,13 +8,16 @@ import { ICONS, STYLE } from './ButtonControl.constants';
...
 import { InferredIconName } from '../Icon/Icon.types';
+import { useCheckAriaLabel } from '../../utils/a11y';
 
 /**
  * `<ControlButtons />` are used to control [close, maximize, minimize, etc] components [usually panels] they are assigned to.
  */
 const ButtonControl = forwardRef((props: Props, providedRef: RefObject<HTMLButtonElement>) => {
   const { className, control, isCircular, ...otherProps } = props;
 
+  useCheckAriaLabel('ButtonControl', props);
```
nataliadelmar, line 19: Same as ButtonCircle. ButtonCircle can be labelled by a tooltip, so we won't require aria-label(ledby) to it. I have added a story to Tooltip that we don't get a console warning only when we trigger a label tooltip from the ButtonCircle

src/components/List/List.types.ts
```diff
@@ -67,6 +69,8 @@ export interface Props {
   initialFocus?: number;
 }
 
+export type Props = ListProps & AriaLabelRequired;
```
nataliadelmar, line 72: All lists should have an aria-label(lledby) SPARK-582998

src/components/NavigationTab/NavigationTab.stories.tsx
```diff
@@ -15,6 +15,9 @@ export default {
       page: DocumentationPage(Documentation, StyleDocs),
     },
   },
+  args: {
+    icon: 'chat',
```
nataliadelmar, line 19: This is something I saw we missed and was causing the story to look empty

src/components/Reaction/Reaction.tsx
```diff
@@ -24,6 +24,9 @@ const Reaction: FC<Props> = (props: Props) => {
         autoplay: autoPlay,
         animationData: animationData,
         name,
+        rendererSettings: {
+          title: props['aria-label'], // adds aria-label to the svg
```
nataliadelmar, line 28: source: https://github.com/airbnb/lottie-web/?tab=readme-ov-file#other-loading-options

PR #735, opened 2024-12-13
feat(a11y): make aria-label mandatory for popover close button when present

lismith2-cisco:
# Description

Use types to make aria-label mandatory for popover close button when present, and modify types which derive from Popover.

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582991

### DIFF
--- a/src/components/Popover/Popover.typetest.ts
+++ b/src/components/Popover/Popover.typetest.ts
@@ -1,0 +1,36 @@
+import { PopoverCloseButtonProps } from './Popover.types';
+import { ExpectExtends, Expect, ExpectFalse } from '../../utils/typetest.util';
+
+// NOTE: STATIC TYPE TESTS RUN ON LINT
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type cases = [
+  // if there is a close button, there must be a closeButtonProps aria-label
+  Expect<ExpectExtends<PopoverCloseButtonProps, Record<string, never>>>,
+  Expect<ExpectExtends<PopoverCloseButtonProps, { closeButtonPlacement: 'none' }>>,
+  ExpectFalse<ExpectExtends<PopoverCloseButtonProps, { closeButtonPlacement: 'top-left' }>>,
+  ExpectFalse<ExpectExtends<PopoverCloseButtonProps, { closeButtonPlacement: 'top-right' }>>,
+  ExpectFalse<
+    ExpectExtends<
+      PopoverCloseButtonProps,
+      { closeButtonPlacement: 'top-left'; closeButtonProps: Record<string, never> }
+    >
+  >,
+  ExpectFalse<
+    ExpectExtends<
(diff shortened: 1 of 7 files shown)


src/components/Popover/Popover.typetest.ts
```diff
@@ -0,0 +1,36 @@
+import { PopoverCloseButtonProps } from './Popover.types';
+import { ExpectExtends, Expect, ExpectFalse } from '../../utils/typetest.util';
+
+// NOTE: STATIC TYPE TESTS RUN ON LINT
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type cases = [
+  // if there is a close button, there must be a closeButtonProps aria-label
+  Expect<ExpectExtends<PopoverCloseButtonProps, Record<string, never>>>,
+  Expect<ExpectExtends<PopoverCloseButtonProps, { closeButtonPlacement: 'none' }>>,
```
nataliadelmar, line 9: can we check that the following cases are also ok for ts:

- `Expect<ExpectExtends<PopoverCloseButtonProps, { closeButtonPlacement: 'none', closeButtonProps: {'aria-label': 'abc' }}>>,`
- `Expect<ExpectExtends<PopoverCloseButtonProps, { closeButtonPlacement: 'none', closeButtonProps: {'className': 'something' }}>>,`

src/components/TooltipPopoverCombo/TooltipPopoverCombo.types.ts
```diff
@@ -21,7 +22,7 @@ export interface Props {
   /**
    * An object of Popover props to be passed through to the Popover component.
    */
-  otherPopoverProps?: Partial<Omit<PopoverProps, 'id'>>;
+  otherPopoverProps?: Partial<Omit<PopoverProps, 'id'>> & PopoverCloseButtonProps;
```
nataliadelmar, line 25: are you adding `& PopoverCloseButtonProps` because omiting id from PopoverProps makes it forget about our complex type?

src/components/Popover/Popover.typetest.ts
lismith2-cisco: can do, I left it because it didn't seem to matter if it was valid or not, but I will pin it down

src/components/TooltipPopoverCombo/TooltipPopoverCombo.types.ts
lismith2-cisco: no, because making it partial makes the close button props always optional

lismith2-cisco: For some reason, I didn't have any issue with omit, that I noticed...

lismith2-cisco: maybe I misunderstood the issue

nataliadelmar (approved):

PR #737, opened 2024-12-16
feat(modal-container): add title on h2 tag 

nataliadelmar:
# Description

If there is a visible title on a modal, it should be on an H2 tag and it should by default be the label of the modal unless aria-label or aria-labelledby are specified.

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582989
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-582910

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -40,6 +40,7 @@ const Banner: FC<Props> = (props: Props) => {
       round={75}
       data-shape={shape}
       style={style}
+      aria-label={title}
       {...mutatedOtherProps}
     >
       {image}
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -44,7 +44,7 @@ const Banner: FC<Props> = (props: Props) => {
       {...mutatedOtherProps}
     >
       {image}
-      <Text className={STYLE.title} type={TEXT_CONSTANTS.TYPES.BANNER_PRIMARY} tagName="h1">
+      <Text className={STYLE.title} type={TEXT_CONSTANTS.TYPES.BANNER_PRIMARY} tagName="h2">
         {title}
       </Text>
       <Text
--- a/src/components/ModalContainer/ModalContainer.tsx
+++ b/src/components/ModalContainer/ModalContainer.tsx
@@ -3,6 +3,7 @@ import classnames from 'classnames';
 import { FocusScope } from '@react-aria/focus';
 
 import ModalArrow from '../ModalArrow';
+import Text from '../Text';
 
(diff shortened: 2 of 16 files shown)


src/components/ModalContainer/ModalContainer.tsx
```diff
@@ -41,6 +43,7 @@ const ModalContainer = (props: Props, ref: RefObject<HTMLDivElement>) => {
       className={classnames(className, STYLE.wrapper)}
       id={id}
       style={style}
+      aria-label={props['aria-label'] || title}
```
lismith2-cisco, line 46: Feel like it would be better to use aria-labelledby like OverlayAlert does, but this is ok.

src/components/ModalContainer/ModalContainer.unit.test.tsx
```diff
@@ -128,15 +129,25 @@ describe('<ModalContainer />', () => {
...
 
-      const container = mount(<ModalContainer focusLockProps={{ restoreFocus: true }} />);
+      const container = mount(
+        <ModalContainer aria-label="some modal" focusLockProps={{ restoreFocus: true }} />
+      );
+
+      expect(container).toMatchSnapshot();
+    });
+
+    it('should match snapshot with title', () => {
```
lismith2-cisco, line 147: Would be good to see one with title and children

src/components/ModalContainer/ModalContainer.unit.test.tsx
```diff
@@ -278,13 +289,26 @@ describe('<ModalContainer />', () => {
...
+      expect.assertions(1);
+
+      const title = 'Some visible title';
+
+      const wrapper = mount(<ModalContainer title={title} />)
+        .find(ModalContainer)
+        .find(Text)
+        .filter({ tagName: 'h2' });
+
+      expect(wrapper.exists()).toBe(true);
```
lismith2-cisco, line 310: Doesn't check for 'Some visible title'

src/components/Banner/Banner.tsx
```diff
@@ -40,10 +40,11 @@ const Banner: FC<Props> = (props: Props) => {
       round={75}
       data-shape={shape}
       style={style}
+      aria-label={title}
```
lismith2-cisco, line 43: wouldn't it be better to do it with aria-labelledby?

lismith2-cisco (changes requested):

lismith2-cisco:
I am concerned that you have added a 'title' parameter and not been able to use it anywhere in practice, because there are always children before the title. It would be better if we could rethink this to be actually useful.

PR #748, opened 2025-01-06
feat(nav-tab): require aria-label(ledby) or label

nataliadelmar:
It is important that, when the NavTab is not expanded, the circular button remains with an aria-label. 
To prevent that, we are making aria-label(ledby) or label required and using it to label the nav tab in case it is not expanded.

Non-breaking change in Cantina as all instances of NavigationTab have aria-labels 🎉

### DIFF
--- a/src/components/NavigationTab/NavigationTab.tsx
+++ b/src/components/NavigationTab/NavigationTab.tsx
@@ -14,12 +14,14 @@ import './NavigationTab.style.scss';
 const NavigationTab: FC<Props> = (props: Props) => {
   const { icon, label, count = 0, className, id, size, style, active, ...otherProps } = props;
 
+  const isExpanded = size == 200;
+
   const iconComponent = icon ? (
     <Icon className={STYLE.icon} name={icon} scale={24} weight={'filled'} strokeColor={'none'} />
   ) : null;
 
   const labelComponent =
-    size == 200 && label ? (
+    isExpanded && label ? (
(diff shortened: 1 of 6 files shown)


src/components/NavigationTab/NavigationTab.tsx
```diff
@@ -14,12 +14,14 @@ import './NavigationTab.style.scss';
 const NavigationTab: FC<Props> = (props: Props) => {
   const { icon, label, count = 0, className, id, size, style, active, ...otherProps } = props;
 
+  const isExpanded = size == 200;
```
lismith2-cisco, line 17: I know this is a copy-paste, but I think our convention is always to use triple equals

src/components/NavigationTab/NavigationTab.tsx
```diff
@@ -45,6 +47,7 @@ const NavigationTab: FC<Props> = (props: Props) => {
         data-size={size || DEFAULTS.SIZE}
         id={id}
         style={style}
+        aria-label={isExpanded && label ? props['aria-label'] : props['aria-label'] || label} // if expanded with label, the accessible name will default to the visible label inside.
```
lismith2-cisco, line 50: I don't get what you're doing here

lismith2-cisco, line 50: And you've not added any unit tests for it

lismith2-cisco (changes requested):

PR #754, opened 2025-01-30
feat(list): remove current focus from list item

Coread:
# Description

To improve the performance of the List, remove parts of the context that change often.

### DIFF
--- a/src/components/ListItemBase/ListItemBase.tsx
+++ b/src/components/ListItemBase/ListItemBase.tsx
@@ -25,24 +25,6 @@ import { usePrevious } from '../../hooks/usePrevious';
 import { getKeyboardFocusableElements } from '../../utils/navigation';
 import { useFocusAndFocusWithinState } from '../../hooks/useFocusState';
 
-export const useItemHasFocus = ({ itemIndex }) => {
-  const listContext = useListContext();
-
-  const currentFocus = listContext?.currentFocus;
-
-  const [itemHasFocus, setItemHasFocus] = useState(false);
-
-  useEffect(() => {
-    if (currentFocus === itemIndex) {
-      setItemHasFocus(true);
-    } else {
-      setItemHasFocus(false);
-    }
-  }, [currentFocus, itemIndex]);
-
-  return itemHasFocus;
-};
-
 type RefOrCallbackRef = RefObject<HTMLLIElement> | ((instance: HTMLLIElement) => void);
 
 //TODO: Implement multi-line
@@ -71,6 +53,9 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
   } = props;
 
   let content: ReactNode, start: ReactNode, middle: ReactNode, end: ReactNode;
+
+  // eslint-disable-next-line no-console
+  console.log('RENDER', itemIndex);
 
   const listContext = useListContext();
 
@@ -187,7 +172,10 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
    * Focus management
    */
   const setCurrentFocus = listContext?.setCurrentFocus;
-  const updateFocusBlocked = listContext?.updateFocusBlocked;
+  // const updateFocusBlocked = listContext?.updateFocusBlocked;
+
+  const [updateFocusBlocked, setUpdateFocusBlockedInternal] = useState(false);
+
   const setUpdateFocusBlocked = listContext?.setUpdateFocusBlocked;
   const shouldFocusOnPress = listContext?.shouldFocusOnPress || false;
   const shouldItemFocusBeInset =
@@ -195,7 +183,6 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
   const listFocusedWithin = listContext?.isFocusedWithin;
   const addFocusCallback = listContext?.addFocusCallback;
 
-  // const itemHasFocus = useItemHasFocus({ itemIndex });
   const [itemHasFocus, setItemHasFocus] = useState(undefined);
 
   const listItemTabIndex = getListItemBaseTabIndex({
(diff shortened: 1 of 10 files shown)


src/hooks/useOrientationBasedKeyboardNavigation.ts
```diff
@@ -46,11 +51,46 @@ const useOrientationBasedKeyboardNavigation = (
...
+    updateFocusBlockedRef.current = value;
+    setUpdateFocusBlockedInternal(value);
+  }, []);
+
+  const setCurrentFocus = useCallback((index: ListItemBaseIndex) => {
+    currentFocusRef.current = index;
+    setCurrentFocusInternal(index);
+  }, []);
+
+  const addFocusCallback = useCallback((index, callback) => {
```
martroos, line 76: suggestion: There should probably be a `removeFocusCallback` equivalent provided so that items can "unregister" themselves when unmounted, otherwise we end up with a potential memory leak (especially when non-numeric indexes are used - numeric ones will just override each other, but non-numeric are likely to be uuid).

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -167,37 +168,38 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
...
-  const updateFocusBlocked = listContext?.updateFocusBlocked;
   const setUpdateFocusBlocked = listContext?.setUpdateFocusBlocked;
   const shouldFocusOnPress = listContext?.shouldFocusOnPress || false;
   const shouldItemFocusBeInset =
     listContext?.shouldItemFocusBeInset || DEFAULTS.SHOULD_ITEM_FOCUS_BE_INSET;
   const listFocusedWithin = listContext?.isFocusedWithin;
+  const addFocusCallback = listContext?.addFocusCallback;
 
-  const listItemTabIndex = getListItemBaseTabIndex({ interactive, listContext, focus });
+  const [itemHasFocus, setItemHasFocus] = useState(undefined);
```
martroos, line 179: question: Why not initialize to `false`? `getItemBaseTabIndex` expects a boolean for `focus` anyway, and I don't think there are any checks that explicitly check for `undefined`.

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -225,17 +227,29 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
...
         )
       );
-    // Also include "focus" in the dependencies to update the tab indexes when the focus changes
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [focus, ref, isFocusedWithin, focusChild, listItemTabIndex]);
+  }, [ref, isFocusedWithin, focusChild, listItemTabIndex]);
+
+  // We must not autofocus when rendering new elements
+  // If a new element is rendered that has the same index as current focus (i.e. the focused element is replaced)
+  // then it would otherwise try and render, because focus is not blocked and the element is the current focus
```
martroos, line 234: nitpick: Should this be "then it would otherwise try and focus..."?

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -245,35 +259,34 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
...
       }
+    },
+    [focusChild, ref]
+  );
+
+  // This registers the list item base with the orientation based keyboard navigation hook
+  // The hook will then tell the list item when it has focus and when the focus is blocked
+  useLayoutEffect(() => {
+    if (addFocusCallback) {
+      addFocusCallback(itemIndex, onFocusCallback);
```
martroos, line 270: suggestion: As mentioned in the hook file, it would be good to unregister when the list item is unmounted.

martroos (changes requested):
Requesting changes for the callback unregister thing - if there is a good reason not to have it then this is probably fine as is, but would be good to consider it.

src/components/ListItemBase/ListItemBase.tsx
Coread: Changed to initialize to false

src/hooks/useOrientationBasedKeyboardNavigation.ts
Coread: I'm going to use the same callback and pass in undefined

src/components/ListItemBase/ListItemBase.tsx
Coread: fixed

src/components/ListItemBase/ListItemBase.tsx
Coread: callbacks are now unregistered

src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -267,7 +267,13 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
   // The hook will then tell the list item when it has focus and when the focus is blocked
   useLayoutEffect(() => {
     if (addFocusCallback) {
+      const currentIndex = itemIndex;
```
martroos, line 270: question: Is this needed? The value of `itemIndex` cannot change because it's a different value each render (not mutable like a useRef would be), so it shouldn't need re-assigning.

Coread: changed

src/hooks/useOrientationBasedKeyboardNavigation.ts
Coread: changed to delete

src/hooks/useOrientationBasedKeyboardNavigation.ts
```diff
@@ -74,7 +74,11 @@ const useOrientationBasedKeyboardNavigation = (
   }, []);
 
   const addFocusCallback = useCallback((index, callback) => {
```
martroos, line 76: nitpick: It's a bit odd calling this `addFocusCallback` when it also removes them, I can't think of a much better name (maybe `setFocusCallback` but even that's not great) - I personally would have gone for separate callbacks, one for add and one for remove (well I would have gone for register and unregister items which provided an object of callbacks since that is more flexible in the long run). Nonetheless, I'm not too concerned with changing this, but apparently I'm incapable of not commenting!

martroos (approved):

PR #756, opened 2025-02-10
feat(spatial-navigation): spatial navigation provider

maxinteger:
# Description

- add spatial navigation provider
- update complex component to work with the new provider

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-596397


src/components/SpatialNavigationProvider/SpatialNavigationProvider.stories.tsx
```diff
@@ -0,0 +1,310 @@
...
+            button {i + 1}
+          </ButtonPill>
+        ))}
+      </main>
+    </SpatialNavigationProvider>
+  );
+}).bind({});
+ButtonGrid.argTypes = { ...argTypes };
+
+const Form = Template(({ backKey }) => {
```
cscheffauer, line 69: question: not sure if this is the expected behavior:
when going through the form, the radio group does automatically change its value AND the focus then jumps - is this expected? isn't this a usability issue? like if i use the spatial navigation and come to the radio group, i would like to be able to change the radio group selection without loosing focus?
steps to reproduce:
1. go to Form story
2. Using down key to toggle focus through the form and see how the Example radio group changes its selection

src/components/SpatialNavigationProvider/SpatialNavigationProvider.stories.tsx
```diff
@@ -0,0 +1,310 @@
...
+    tNode('4', true, [
+      tNode('4.1', false, [
+        tNode('4.1.1'),
+        tNode('4.1.2')]),
+      tNode('4.2')]),
+  ]);
+
+const exampleTreeMap = convertNestedTree2MappedTree(exampleTree);
+
+const ExampleTreeNode = ({ node }: { node: TreeNodeRecord }) => {
```
cscheffauer, line 180: question: when trying the following:
1. focus to the "right" button in the Tree Navigation Story
2. Press Left arrow, focus will go to the Tree in the middle
3. Pressing left arrow again will not focus the "Left" button, but the "Top" button
Is this expected? i thought it would jump to the left button instead of top.

src/components/SpatialNavigationProvider/SpatialNavigation.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  /**
+   * Set the active element and call .focus() on it
+   * @param element
+   */
+  setActiveElementAndFocus(element: HTMLElement): void {
+    this.setActiveElement(element);
+    queueMicrotask(() => element.focus());
```
cscheffauer, line 143: q: is there a specific reason why we need queueMicrotask here? if yes, can u please add a comment?

src/components/SpatialNavigationProvider/SpatialNavigationProvider.stories.args.ts
```diff
@@ -0,0 +1,16 @@
+import { DEFAULTS } from './SpatialNavigationProvider.constants';
+import { NavKeyMapping } from './SpatialNavigationProvider.types';
+
+export const AWSD_KEY_MAPPING = {
+  ...DEFAULTS.SPATIAL_NAVIGATION_KEY_MAPPING,
+  back: 'Backspace',
+} as NavKeyMapping;
+
+export default {
+  backKey: {
```
cscheffauer, line 10: q: i tried setting the backkey, but it did not have any effect for me in storybook - what is it supposed to do?

src/components/SpatialNavigationProvider/SpatialNavigationProvider.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Setup other components
+   */
+  useEffect(() => {
+    setupHideOnPlugin({ hideKeys: [navigationKeyMapping.back], stopEventPropagation: true });
+  });
+
+  useEffect(() => {
+    const validKeys = [
```
cscheffauer, line 47: nitpick: this could be moved to a constants file, doesn't need to. not too worried if addressed or not

cscheffauer (changes requested):

src/components/SpatialNavigationProvider/SpatialNavigationProvider.stories.tsx
maxinteger: It was a known issue, I just forgot to fix it. It will be fixed in the next commit

src/components/SpatialNavigationProvider/SpatialNavigationProvider.stories.tsx
maxinteger: Yes this is an unfortunate case, If you press the shift+left before the focus jumps to the top button. It will show you the distances and the closest button is indeed the top one. Probably we can narrow the detection aria so the algorithm will not pick up elements from perpendicular directions.

If it is OK I will address this issue under a new story.

src/components/SpatialNavigationProvider/SpatialNavigationProvider.tsx
maxinteger: Not sure I can, `validKeys` created from `navigationKeyMapping` which is a prop of the SpatialNavigationProvider and can change

src/components/SpatialNavigationProvider/SpatialNavigationProvider.stories.args.ts
maxinteger: back key for backward navigation which has 2 cases
1. navigation backward in terms of pages (same as the back button in the browser nav)
2. close the last opened overlay

So when user presses back key multiple times we close the open overlay(s) first then do page navigation back.
The idea is the same as on the mobile phones back navigation.

In the story you can try only the remove overlay, for example open the select menu and then press the back key.

src/components/SpatialNavigationProvider/SpatialNavigation.ts
maxinteger: done

cscheffauer (approved):

PR #757, opened 2025-02-10
feat(alert-banner): do not render button section if no buttons passes

nataliadelmar:
At the moment, AlertBannerNext has an optional property buttons that accepts a ReactElement[]. If not defined, AlertBanner still draws its container, leaving a weird blank space at the end of the banner.

If we have no buttons, we want the space not to be there

image


nataliadelmar:
we don't need this

PR #758, opened 2025-02-12
feat: create TooltipMenuTriggerCombo component

dominiccarrington:
# Description
This PR fixes 2 issues related to MenuTrigger:
1. onOpenChange doesn't fire and a11y properties don't update when the user activates the trigger component using the keyboard.
2. MenuTrigger with Tooltip as triggerComponent doesn't work correctly.

react-aria functionality doesn't fire when the trigger component is activated by the keyboard. This is causes the onOpenChange callback to not fire and a11y props (such as aria-expanded) to not update on the trigger component.

This PR makes the react-aria useMenuTriggerState control if the menu is open or not and makes the Popover trigger manually when the useMenuTriggerState.isOpen is updated. (like the Select component)
This causes a small breaking change if you are using Tooltips as the triggerComponent since tippy doesn't add the click event handler to the button anymore.

Additionally MenuTrigger doesn't provide all the props to the trigger component when using a Tooltip that it does normally when a button is provided directly. (for example aria-controls) This is because the properties don't get merged onto the Button but instead the Popover created by the Tooltip.

This PR creates a TooltipMenuTriggerCombo component, similar to TooltipPopoverCombo, which handles capturing the props given to the MenuTrigger triggerComponent and passes them to the button component instead.

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-615094
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-616381

### DIFF
--- a/src/components/MenuTrigger/MenuTrigger.stories.tsx
+++ b/src/components/MenuTrigger/MenuTrigger.stories.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { MultiTemplate, Template } from '../../storybook/helper.stories.templates';
 import { DocumentationPage } from '../../storybook/helper.stories.docs';
 import StyleDocs from '../../storybook/docs.stories.style.mdx';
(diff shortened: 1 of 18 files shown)


dominiccarrington:
Closed in favour of the versions coming from momentum-design

PR #762, opened 2025-03-03
fix: use result of onShow callback in TooltipPopoverCombo

maxinteger:
# Description

Passing the result of the custom (popover and tooltip) onShow result to tippy as expected.

# Links

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-623004


dominiccarrington (commented):
onHide can also be cancelled: https://atomiks.github.io/tippyjs/v5/lifecycle-hooks/#onhide

question: instead of `onShow` / `onHide`, should we use different events that won't be cancelled like `onMount` / `onHidden`? These do fire at a slightly different point in the lifecycle.

maxinteger:
@dominiccarrington 

I avoided `onHide` because it is not relevant for my use case, but you are right, we should have it at least for consistency

Not sure why we picked onShow/onHide instead of onMount/onUnmount, but for me cancelation is feature which I need to solve a problem in the app.

dominiccarrington (approved):

PR #768, opened 2025-03-17
fix: adding new themes to themeprovider

cscheffauer:
# Description

- Bumping tokens repo version to include new themes
- Adding new themes to constants to pass it through to momentum-design
- Tested in Storybook, works as expected (background gradient changing)


chrisadubois (approved):
fantastic, looks perfect just as expected.

robstax (approved):

PR #779, opened 2025-04-09
fix: migrate addreactionbutton reactionbadge

cscheffauer:
# Description

- Migrating AddReactionButton and ReactionBadge to use the new momentum-design component
  - Changed API to fire onClick instead of onPress (breaking change for consuming applications!) 
  - Added example story to AddReactionButton to check it still working with TooltipPopoverCombo
- Includes a styling change to match latest specs (background color -> black)
- Enabling unit testing of web components which do have attachInternals by using `element-internals-polyfill`
- Added unit testing utility to renderWebComponent with testing-library
- Adding brand-visuals package to resolutions to use latest momentum-design component package
- Remove global focus behavior from all mdc-button components (focus is handled there)
- Excluding `--mdc-` css variable overrides from cssTokenCheck pre-commit hook


src/components/ReactionBadge/ReactionBadge.unit.test.tsx
```diff
@@ -1,15 +1,19 @@
...
 import * as jsonImport from '../../hooks/useDynamicJSONImport';
 import smile from '@momentum-design/animations/dist/lottie/reactions/smile.json';
-
-import ButtonSimple from '../ButtonSimple';
+import { screen } from '@testing-library/react';
 
 describe('<ReactionBadge name="haha" />', () => {
   const reactionProp = <Reaction name="haha" />;
+
+  const setup = async (component: any) => {
```
dominiccarrington, line 14: nitpick:
```suggestion
  const setup = async (component: React.ReactElement) => {
```

test/utils.ts
```diff
@@ -106,3 +107,19 @@ export const sypOnEventListener = (target: EventTarget, spyOnEvents?: string[])
...
+  const { container } = render(component);
+
+  // we have to wait for the web component to be rendered
+  await waitFor(() => {
+    if (selector) {
+      expect(container.querySelector(selector)).toBeTruthy();
+    }
+  });
+
+  return container;
```
maxinteger, line 124: I understand, this is a replacement for `mountAndWait` which returns with only the container. 

But `render` returns with many other useful properties. `renderWithWebComponent` should return with all the exposed props.

maxinteger (changes requested):

maxinteger:
Not sure it is a real problem but I noticed the old and new add reaction button trigger different events:

image

`onPressStart` and `onPressEnd` probably coming from react-aria

src/components/AddReactionButton/AddReactionButton.stories.args.ts
```diff
@@ -9,9 +9,28 @@ const addReactionButtonArgTypes = {};
 
 export { addReactionButtonArgTypes };
 
+delete commonAriaButton.onPress;
```
maxinteger, line 12: Do nat use delete 
1. it is inefficient and ruin V8 internal object representation
2. `commonAriaButton` is exported by reference, so if you delete any property it will be missing from other components as well for example the AlertBadge also using `commonAriaButton` and it does not have `onPress` props anymore in your branch while it is there in prod https://momentum-design.github.io/momentum-react-v2/?path=/story/momentum-ui-alertbadge--example

maxinteger (changes requested):

maxinteger (approved):

PR #780, opened 2025-04-09
feat(ListItemBase): allow tooltip on component

jor-row:
# Description

We need to be able to have an Item inside a Menu with a tooltip. This will allow that to happen with the help of the MdcTooltip component.

# Links
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-647080

https://github.com/user-attachments/assets/d9bad553-8118-4e4c-b9df-1ce9e80b3b79


src/components/ListItemBase/ListItemBase.tsx
```diff
@@ -49,9 +51,12 @@ const ListItemBase = (props: Props, providedRef: RefOrCallbackRef) => {
     onBlur,
     onBlurWithin,
     onFocusWithin,
+    tooltipProps,
     ...rest
   } = props;
 
+  const [itemId] = useState(rest.id || uuidv4);
```
cscheffauer, line 58: issue: `uuidv4` is a function, needs to be called here.

cscheffauer (changes requested):
one issue, rest looking good

src/components/ListItemBase/ListItemBase.tsx
jor-row: Good catch, fixed

cscheffauer (approved):

PR #786, opened 2025-04-10
fix: migrate buttondialpad

cscheffauer:
# Description

- Migrating the ButtonDialpad to use the new momentum-design button
- Removing size and children attribute to simplify component (not necessary)
- Updating unit tests

### DIFF
--- a/src/components/ButtonDialpad/ButtonDialpad.constants.ts
+++ b/src/components/ButtonDialpad/ButtonDialpad.constants.ts
@@ -14,6 +14,7 @@ const STYLE = {
   primaryText: `${CLASS_PREFIX}-primary-text`,
   secondaryText: `${CLASS_PREFIX}-secondary-text`,
   wrapper: `${CLASS_PREFIX}-wrapper`,
+  innerWrapper: `${CLASS_PREFIX}-inner-wrapper`,
 };
 
 export { CLASS_PREFIX, DEFAULTS, SIZES, STYLE };
--- a/src/components/ButtonDialpad/ButtonDialpad.tsx
+++ b/src/components/ButtonDialpad/ButtonDialpad.tsx
@@ -1,28 +1,31 @@
 import React, { forwardRef, RefObject } from 'react';
-import ButtonSimple from '../ButtonSimple';
+import { Button as MdcButton } from '@momentum-design/components/dist/react';
 
-import { DEFAULTS, STYLE } from './ButtonDialpad.constants';
-import { Props } from './ButtonDialpad.types';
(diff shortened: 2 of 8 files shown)


src/components/ButtonDialpad/ButtonDialpad.constants.ts
```diff
@@ -14,6 +14,7 @@ const STYLE = {
   primaryText: `${CLASS_PREFIX}-primary-text`,
   secondaryText: `${CLASS_PREFIX}-secondary-text`,
   wrapper: `${CLASS_PREFIX}-wrapper`,
+  innerWrapper: `${CLASS_PREFIX}-inner-wrapper`,
```
jor-row, line 17: The default SIZE and SIZE constants can also be removed as they are no longer needed

src/components/ButtonDialpad/ButtonDialpad.tsx
```diff
@@ -1,28 +1,31 @@
...
       ref={providedRef}
-      data-size={size || DEFAULTS.SIZE}
-      data-disabled={disabled || DEFAULTS.DISABLED}
-      isDisabled={disabled}
-      {...otherProps}
+      // size 40 is maximum size for MdcButton, setting it here and overriding
+      // size in scss file to 64px
+      size={40}
+      variant="tertiary"
+      {...(otherProps as any)}
```
jor-row, line 22: can we type this otherProps? as ```Omit<Props, 'className'>```

src/components/ButtonDialpad/ButtonDialpad.tsx
```diff
@@ -1,28 +1,31 @@
...
+import type { Button } from '@momentum-design/components';
+import type { Props } from './ButtonDialpad.types';
 
-  const children = props.children || props.primaryText;
+const ButtonDialpad = forwardRef((props: Props, providedRef: RefObject<Button>) => {
+  const { className, ...otherProps } = props;
 
   return (
-    <ButtonSimple
+    <MdcButton
```
jor-row, line 15: the disabled button doesn't really look disabled but i guess it was that way before so its fine as is

src/components/ButtonDialpad/ButtonDialpad.constants.ts
jor-row: the default for DISABLED is only used in a test so can be removed as well

jor-row (changes requested):
some very small comments like unused constants that can be removed and a type issue

chrisadubois (approved):

jor-row (approved):

PR #798, opened 2025-05-12
fix: chip migration

cscheffauer:
# Description

- Migrate to use new Chip from momentum-design
- Updated unit tests


src/components/Chip/Chip.stories.docs.mdx
```diff
@@ -1,9 +1,3 @@
 The `<Chip />` component.
 
-The component can display text, icons, and/or an avatar supplied by the consumer
-
-The defaul size of the component is 28, the icon size as designed is 16, but can be set by consumer.
-
-Both icons and avatar will be positioned within chip by default.
-
-Padding of .25em overrides supplied style prop.
+Rendering LabelChip from momentum-design
```
maxinteger, line 3: The story of this component looks very simplistic, which is OK, but we should at least provide a link to the more detailed story: https://momentum-design.github.io/momentum-design/en/components/?path=/docs/components-chip-alert--docs

Even if the current server ignore the `path` parameter

src/components/Chip/Chip.stories.args.ts
```diff
@@ -82,70 +58,72 @@ const chipArgTypes = {
...
-        summary: 'TeamColor',
+        summary: '(e: FocusEvent) => void',
       },
       defaultValue: {
-        summary: CHIP_CONSTANTS.DEFAULTS.COLOR,
+        summary: 'undefined',
       },
     },
   },
-  multiline: {
```
maxinteger, line 114: I saw in momentum design we have alternatives for `error` and `search` but what about `multiline` is it completely deprecated now?

maxinteger (commented):
Looks good but I have 2 questions

src/components/Chip/Chip.stories.args.ts
cscheffauer: Chips should not do multiline, correct

maxinteger (approved):

PR #801, opened 2025-05-19
fix: buttonpill migration

cscheffauer:
# Description

- Migrating the ButtonPill component to use the new momentum-design component
- Update unit tests respectively


src/components/NotificationSystem/NotificationSystem.unit.test.tsx
```diff
@@ -404,10 +412,10 @@ describe('<NotificationSystem />', () => {
     });
 
     it('should close the `medium attention` notification after clicking on the close button', async () => {
-      expect.assertions(5);
+      // expect.assertions(5);
```
maxinteger, line 415: please remove it if you do not needed

src/components/ButtonPill/ButtonPill.types.ts
```diff
@@ -13,42 +20,39 @@ export interface Props extends ButtonSimpleProps {
...
-   */
-  inverted?: boolean;
-
-  /**
-   * Size index of this ButtonPill.
+   * Whether or not to stop the event from bubbling up.
    */
-  size?: number;
-}
+  stopPropagation?: boolean;
```
maxinteger, line 57: Probably you should mention in the docs as well:
- this stops click event propagation
- temporal solution, do not use it if it is not really necessary

src/components/List/List.unit.test.tsx
```diff
@@ -533,16 +535,25 @@ describe('<List />', () => {
           </ListItemBase>
         </List>
       );
-      const listItems = getAllByRole('listitem');
 
       await user.tab();
 
-      expect(listItems[0]).toHaveFocus();
+      await waitFor(() => {
+        expect(screen.getAllByRole('listitem')[0]).toHaveFocus();
```
maxinteger, line 542: I prefer to use query functions from the render function's return instead of from `screen` 
- less typing
- more explicit
- clear(er) because you query the rendered DOM instead of some global object. Probably technical point of view the 2 are identical, but still

maxinteger (changes requested):
There are some storybook deferences:
1. ButtonPill - Ghost Not Outline Not inverted: image
2. ButtonPill - Not Ghost Outline Not inverted: image
3. ButtonPill - Not Ghost Outline inverted: image
4. ButtonPill - Ghost Outline Not inverted  image
5. ButtonPill - Not Ghost Outline inverted: image

src/components/List/List.unit.test.tsx
cscheffauer: in this scenario you have to destructure the return of the render function though, so i would disagree with less typing, especially if type ahead is enabled. Since we don't have a convention for it and it works the exact same way, i will leave it as is for now.

PR #802, opened 2025-05-19
feat(reaction): migrate reaction component to mdc-animation

maxinteger:
# Description

Migrated the Reaction component to the new mdc-animation web component.

Also, expose MdcAnimation for non-reaction use cases.


package.json
```diff
@@ -105,8 +105,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@momentum-design/animations": "^0.0.4",
-    "@momentum-design/components": "0.66.3",
+    "@momentum-design/animations": "0.3.1",
```
cscheffauer, line 108: suggestion: i would suggest to add animations also to the "resolutions", to make sure there is only 1 package version in use

src/components/Reaction/Reaction.tsx
```diff
@@ -1,67 +1,69 @@
...
-    return () => {
-      if (animation) {
-        if (onComplete) {
-          animation.removeEventListener('complete', onComplete);
-        }
-        animation.destroy();
-      }
-    };
-  }, [svgContainer, animationData, autoPlay, loop, name, onComplete]);
+  const [loadingState, setLoadingState] = useState<AnimationLoadingState>(() => 'loading');
```
cscheffauer, line 25: there is a mixture of strings and functions used for the loadingState, is this expected?
like default is `() => 'loading'` but then there is also `loading` as a string. we should clean this up

src/components/Reaction/Reaction.unit.test.tsx
```diff
@@ -146,63 +135,33 @@ describe('<Reaction/>', () => {
...
-      const wrapper = await mountAndWait(
-        <Reaction name="haha" size={16} onComplete={onComplete} />
-      );
-
-      expect(addEventListener).toHaveBeenCalledWith('complete', onComplete);
-      expect(removeEventListener).not.toHaveBeenCalledWith('complete', onComplete);
-
-      wrapper.unmount();
+    it('should render loading spinner first', () => {
+      const { container } = render(<Reaction name="haha" />);
```
cscheffauer, line 139: suggestion: there is a utility function called `renderWithWebComponent` which does the querySelector waiting automatically. This will shorten the lines here.
Same for tests below

cscheffauer (changes requested):

src/components/Reaction/Reaction.unit.test.tsx
maxinteger: Thanks for the tip. I replaced `render` with `renderWithWebComponent`, but I still need to wait for the loading spinner to disappear because the actual animation is loaded async.

PR #803, opened 2025-05-19
fix: remove tag component

cscheffauer:
# Description

- Removing Tag Component in favor of Chip component


src/components/MeetingContainer/MeetingContainer.unit.test.tsx.snap
```diff
@@ -974,126 +974,40 @@ exports[`<MeetingContainer /> snapshot should match snapshot with disable suppli
...
-                          </button>
-                        </FocusRing>
-                      </FocusRing>
-                    </ButtonSimple>
-                  </Tag>
-                  <Tag
+                        Messages
+                      </mdc-chip>
+                    </Chip>
+                  </Chip>
```
maxinteger, line 993: Do you know why does it render a nested `Chip`?
```html
                  <Chip key="0" >
                    <Chip ...>
                      <mdc-chip ...>
                      </mdc-chip>
                    </Chip>
                  </Chip>
```

maxinteger (changes requested):
The Chip's text is missing in storybook:

image

src/components/MeetingContainer/MeetingContainer.unit.test.tsx.snap
cscheffauer: cause of the React component which gets auto-generated by momentum-design

maxinteger (changes requested):
I found one more issue in the storybook, there is no space between the chips:
image

maxinteger (approved):

PR #823, opened 2025-07-02
fix: remove buttoncontrol and overlayalert

cscheffauer:
# Description

- Removal of ButtonControl from index, since it will be depredated (keeping in Code since its still used by Toast.
- Removal of OverlayAlert code now completely (was only removed from index before).
- Updated stories.


src/index.js
```diff
@@ -49,7 +49,6 @@ export {
   PresenceType,
   ButtonCircle,
   ButtonCircleToggle,
-  ButtonControl,
```
maxinteger, line 52: Why do you remove `ButtonControl`? it is used 6 places in Cantina.
How is it related to removing the OverlayAlert?

maxinteger (changes requested):

PR #825, opened 2025-07-08
fix(menu): re-export mdc menu components

maxinteger:
# Description

Add new MDC menu related component by re-export them from the momentum-design/components package

# Links

[Momentum design / Menu popover](https://momentum-design.github.io/momentum-design/en/components/)


src/components/Icon/Icon.types.ts
```diff
@@ -1,6 +1,7 @@
 import { CSSProperties } from 'react';
 import type IconKeys from '@momentum-design/icons/dist/types/types';
 import { Props as TooltipProps } from '../Tooltip/Tooltip.types';
+import { MdcIconProps } from '../../types';
```
cscheffauer, line 4: nitpick: this can be a type only import, using `import type {...`

